PR #7, opened 2018-04-24
new example demonstrating tagged boundaries from pygmsh

gdmcbain:
Here's another extension of `examples/ex1.py`, still solving the Laplace equation but now with mixed boundary conditions, two parts isopotential (charged and earthed) and the rest insulated. The isopotential parts are tagged during the construction of the geometry in pygmsh, as introduced in `ex13.py`.

The example is ∇² _u_ = 0 in _Ω_ = {(_x, y_) : 1 < _x_ ² + _y_ ² < 4, 0 < _θ_< π/2}, where tan _θ_ = _y/x_, with _u_ = 0 on _y_ = 0 and _u_ = 1 on _x_ = 0.   Although these boundaries would be simple enough to identify using the coordinates and `skfem.assembly.find_dofs` as in `ex3.py`, the present technique generalizes to more complicated shapes.

The exact solution is _u_ = 2 _θ_ / π.  The field strength is |∇ _u_|² = 4/π² (_x_ ² + _y_ ²) so the conductance (for unit potential difference and conductivity) is ‖∇ _u_‖² = 2 ln 2 / π.

### DIFF
--- a/examples/conductance.py
+++ b/examples/conductance.py
@@ -1,0 +1,57 @@
+from skfem import *
+from skfem.models.poisson import laplace, mass, unit_load
+
+import numpy as np
+
+from pygmsh import generate_mesh
+from pygmsh.built_in import Geometry
+
+geom = Geometry()
+points = []
+lines = []
+radii = [1., 2.]
+lcar = .1
+points.append(geom.add_point([0.] * 3, lcar))  # centre
+for x in radii:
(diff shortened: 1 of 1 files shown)


kinnala:
Thank you for both of the new contributions. If it's fine by you, I included your pull request comments as explanations to the respective files.

I really need to improve the explanation of my own examples in the future.

Issue #19, opened 2018-07-25
skfem.models not installed


gdmcbain:
Retrying some examples in a fresh venv failed.  It appears that skfem.models isn't installed with a plain `pip install`; I think I must previously have been passing `--editable`, which brings in the whole repo.

To reproduce:
```shell
python3.6 -m venv skfem
. skfem/bin/activate
pip install git+https://github.com/kinnala/scikit-fem#egg=scikit-fem
python -c 'import skfem'
python -c 'import skfem.models'
```

The second-last command does nothing.  The last raises `ModuleNotFoundError`.

I think the problem is that there's no `skfem/models/__init__.py`.  An empty file here fixes the problem, as in 8dfe60c0ce848eeaa81cd6c8845cdf695db38364:

```shell
pip install -U git+https://github.com/gdmcbain/scikit-fem@models-module#egg=scikit-fem
python -c 'import skfem.models'
```

PR #20, opened 2018-07-25
make skfem.models a subpackage

gdmcbain:
An empty `__init__.py` in `skfem/models` makes Python recognize it as a subpackage.  For #19.


kinnala:
Thanks for this. It would be convenient to have an automated way to test the installation workflow. Also, the model subpackage should be expanded. I will make a new pip release later today or tomorrow and check that this is indeed fixed.

PR #24, opened 2018-08-07
new example demonstrating spatially varying coefficients

gdmcbain:
It wasn't immediately clear to me how to write forms with variable coefficients so I wrote an example, solving the Legendre equation.

I got the clue that `x = w[0][0]` from the example ex05 with the forms defined on a single facet.

### DIFF
--- a/examples/legendre.py
+++ b/examples/legendre.py
@@ -39,7 +39,7 @@ def stiffness(u, du, v, dv, w):
 L = asm(stiffness, basis)
 M = asm(mass, basis)
 
-k, u = eigsh(L, k=6, M=M, which='SM')
+k, u = eigsh(L, M=M, which='SM')
 u = u * np.sqrt([2 / (2 * n + 1) for n in range(len(k))]) * np.sign(u[-1, :])
 
 fig, ax = subplots()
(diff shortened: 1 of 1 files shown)


kinnala:
The thing is, that originally I had a bunch of additional variables in the form definition, such as (x,y,z,h,w,dw) where w is some interpolation of an earlier solution and dw its derivatives. Obviously this is not opimal due to the huge number of unnecessary symbols. Thus, I combined everything to w which is now a list of default variables, defined in InteriorBasis.default_parameters. This introduces another problem: remembering the order of the variables. Moreover, user can always overwrite w by supplying it to skfem.assembly.asm as an optional parameter.

As a side note: If you come up with a better approach, that would be fantastic. I'll check out this pull request later today, though.

gdmcbain:
I quite like [`attrs`](http://www.attrs.org) for this kind of thing. They're like better namedtuples. I believe they inspired the [`dataclasses`](https://docs.python.org/3/library/dataclasses.html) introduced in Python 3.7 but I haven't looked at them yet.

The basic idea is an immutable data structure that's easy to initialize and can be used like `w.x`, `w.h`, &c.

Issue #29, opened 2018-08-09
Modify form variables


kinnala:
In some pull request comment thread, @gdmcbain had a good idea of using 'dataclass-type approach' for the additional form variables.

https://github.com/kinnala/scikit-fem/pull/24#issuecomment-410960587

I think something like this should be implemented. I'm not happy about adding attrs-dependency though if it's not used in more than one place.

It should be evaluated whether attrs is a reasonable dependency (= is used for more than one thing) or whether a custom approach is better.

gdmcbain:
Although [‘reasons to prefer `attrs` over Data Classes’](http://www.attrs.org/en/stable/why.html#data-classes) have been given by the authors of the former, I've just taken a look at the [`dataclasses`](https://docs.python.org/3/library/dataclasses.html#module-dataclasses) added to the Standard Library in 3.7 and they seem to have all the features that I use.  However, as scikit-fem is currently at 3.6, as is the popular [Anaconda Distribution](https://www.anaconda.com/download), it might be too early to require that.  One approach is simply to wait till 3.7 is more widely adopted; shouldn't be too long, I wouldn't think.

The closest thing available in the Standard Library before 3.7 was [typing.NamedTuple](https://docs.python.org/3/library/typing.html?highlight=typing%20namedtuple#typing.NamedTuple).  Indeed, I was using these before I heard about `attrs`.  They would probably suffice here too I think.  I might attempt that in a branch.

Although I haven't looked into the scikit-fem internals here, it also occurs to me that the implementation might not really matter insofar as consistency of the interface is concerned so long as the `w` argument has attributes `w.x`, `w.h`, `w.n`, &c., since the user just accesses these in writing linear and bilinear forms and `w` itself is by default and usually defined behind the scenes in the `default_parameters` method of `InteriorBasis` or `FacetBasis`.  Thus, because of [duck-typing](https://docs.python.org/3/glossary.html?highlight=duck#term-duck-typing), it wouldn't matter if today `default_parameters` returned a `typing.NamedTuple` and tomorrow a `dataclasses.dataclass` and the user wouldn't notice (unless of course they were still using 3.6).  Further, the `assembly` module could use `typing.NamedTuple` in `default_parameters` and a 3.7 user overriding it by explicitly passing `w` to `assembly.asm` could pass in a `dataclasses.dataclass` and all would be well, so long as both had required attributes `x`, `h`, `n`, or some combination thereof.

gdmcbain:
[...]

PR #30, opened 2018-08-10
WIP: 29-modify-form-variables

gdmcbain:
This is a (very) quick experiment at reimplementing the `default_parameters` for the optional `w` argument of `asm` as a `NamedTuple`, as discussed in #29.  Seems to work O. K. on the tests.

### DIFF
--- a/examples/ex04.py
+++ b/examples/ex04.py
@@ -63,7 +63,7 @@ for i in range(2):
     for j in range(2):
         @bilinear_form
         def bilin_penalty(u, du, v, dv, w):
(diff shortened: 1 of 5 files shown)


kinnala:
This looks fantastic. I'll check it out during the weekend.

I'll probably make small modifications as this is a good spot for creating a better interface for interpolating old solutions inside forms (this is required e.g. in nonlinear problems and iterative coupling). Now one just replaces the whole 'w' by providing an array to asm (e.g. ex10.py).

Issue #34, opened 2018-08-16
nonzero natural conditions in 1-D


gdmcbain:
I can't figure out how to impose nonzero natural conditions on a one-dimensional problem.  

Essential conditions are easy, same as in higher dimensions.

Consider u'' = 0 with u(0) = 0 and u'(1) = 1 so that the exact solution is u = x.

I can solve this as trivially two-dimensional problem (∇v, ∇u) = [v, n . ∇u]|(x=1) = [v, 1]|(x=1):

```python
m = MeshQuad.init_tensor(*(np.linspace(0., 1.),)*2)
e = ElementQuad1()
ib = InteriorBasis(m, e)
fb = FacetBasis(m, e)


@linear_form
def boundary_flux(v, dv, w):
    return v * (w.x[0] == 1.)


n = m.p.shape[-1]
L = asm(laplace, ib)
b = asm(boundary_flux, fb)
u, D = ib.find_dofs(lambda x, y: x == 0.)
I = ib.dofnum.complement_dofs(D)  # noqa E741
u = np.zeros_like(b)
u[I] = solve(*condense(L, b, I=I)) # noqa E741
```

but trying the same thing with the minimal modifications to drop the dimension from two to one
```python
m = MeshLine(np.linspace(0., 1.))
e = ElementLineP1()
```
fails at `FacetBasis` with
> NotImplementedError: The given mesh type is not supported!

I think that this is merely the proximate cause, raised by `get_quadrature`, but on inspection `MeshLine` appears incomplete, lacking attributes `facets`, `t2f`, `f2t`.

gdmcbain:
Filling in the quadrature rule for `refdom is "point"` in 904da9b on branch neumann1d with
```python
    elif refdom is "point": # 0
        return np.zeros((1, 1)), np.ones(1)
```
the next hurdle hit is
> AttributeError: 'MeshLine' object has no attribute 'f2t'

kinnala:
I sort of saw these challenges when I reimplemented the 1D support but forgot to do anything about it. Let us try to find a fix so that the implementation is as similar as in higher dimensions.

gdmcbain:
Indeed this works and gets me going again on what I was really working on—coupled domains, but more on that later, all going well (or otherwise).  

Thank you!

Meanwhile a couple of thoughts:
[...]

PR #35, opened 2018-08-17
revised dimension for quadrature rule for point for #34

gdmcbain:
As discussed in #34, "point" is zero-dimensional so should have zero-dimensional abscissæ in its quadrature rule.   This change is more logically consistent but doesn't seem to behave any differently in the example, still giving the correct answer.


kinnala:
I agree with this. I'll make a quick check how the shape of the output is used in the codebase before merging. It's probably fine but since the test coverage is currently only about 80% it doesn't hurt.

We should also add some new tests for the boundary forms, also higher dim. The current tests more or less check only that a constant function is integrated correctly.

gdmcbain:
Yes, besides unit tests on monomials, a practical test & demonstration of boundary forms might involve the calculation of the conductance in ex13.  This is currently done using the Joule heating but can also be done using the more elementary definition in terms of the current through the isopotential boundaries.  This is like conductance = current / voltage versus conductance = power / current squared.  This wasn't added originally as I hadn't worked out how to evaluate boundary forms.

PR #45, opened 2018-09-12
set_aspect for 2-D plots, fixes #44

### LINKED ISSUE #44: equal aspect ratio for Mesh2D.plot
Tiny æsthetic quibble: maybe by default two-dimensional plots should have equal aspect ratio so that they don't distort the domain.  Of course sometimes one might want to, e.g. in hydrology, etc., where the depth is typically orders of magnitude smaller than the width, but, by default, circles and squares shouldn't be ellipses & oblongs.

gdmcbain:
This calls `.set_aspect(aspect)` on the `ax` if it is automatically generated (so not modifying an `ax` that is specified), with `aspect: float = 1.` being the ratio of vertical to horizontal length-scales, or, as matplotlib puts it 
> a circle will be stretched such that the height is num times the width


kinnala:
This is a good idea. Thanks.

PR #47, opened 2018-09-19
factor Mesh.load to read files or meshio.Meshes

### LINKED ISSUE #46: factor Mesh.load for files or constructed meshio.Meshes
I started looking at the `skfem.Mesh.boundaries` attribute used in `test_mesh.Loading`.  Very nice!  This will make the top-level scripts much more readable.  

I wanted to improve `examples/ex14.py` in this way but its `skfem.MeshTri` lacked this attribute; an `skfem.Mesh` only gets `boundaries` if constructed by `classmethod Mesh.load` and that takes the name of a file whereas there's no accessible file containing the mesh in the example.  It uses `pygmsh.generate_mesh` which writes the mesh in a `tempfile.NamedTemporaryFile`.

So I could eschew `generate_mesh` in favour of writing `geom.get_code()` to a GEO file and then running Gmsh myself in a `subprocess` to generate the MSH file, but that's verbose and duplicates a lot of what's in `generate_mesh`.

Instead, what about factorizing `Mesh.load` so that everything after the definition of
```python
        meshdata = meshio.read(filename)
```
is delegated to another `classmethod`, say `from_meshio_mesh` and then either `load` calls that or the user if they have a `meshio.Mesh` (say by passing the values returned from `pygmsh.generate_mesh`) can call it directly.

I'll attempt an implementation.

gdmcbain:
This fixes #46 and demonstrates that by using the proposed new `classmethod Mesh.from_meshio` in `examples/ex13.py`.

### DIFF
--- a/examples/ex13.py
+++ b/examples/ex13.py
@@ -23,6 +23,7 @@ from skfem.models.poisson import laplace, mass, unit_load
 
 import numpy as np
 
+import meshio
 from pygmsh import generate_mesh
 from pygmsh.built_in import Geometry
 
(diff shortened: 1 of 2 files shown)


gdmcbain:
I wasn't able to specify the type of `meshdata` in `classmethod Mesh.from_meshio` since it's `meshio.Mesh` but `meshio` is only imported inside the classmethod.

kinnala:
Fantastic. I'm merging this.

PR #48, opened 2018-09-19
need to import numpy as np for np.zeros

gdmcbain:


gdmcbain:
Without this, `examples/ex08.py` raises 
```
NameError: name 'np' is not defined
```

Issue #31, opened 2018-08-13
Ideas for documentation


kinnala:
After finishing with hex-support, I'd really wish to improve the tutorials and user documentation. I'm open for any ideas, but here are some:

- Write more thorough introductions to the examples and parse them onto a web page for easy access.
- Choose a couple of interesting examples and promote them on README.md as 'highlights'.

Any obvious ideas that I'm missing?

kinnala:
I'm working on Read the Docs implementation where descriptions written in reStructuredText are parsed from example docstring.

gdmcbain:
Yes, I envisage a few different categories of examples.
* Simple introductory ones, 
  * solving standard partial differential equations
  * on meshes simple enough to generate in code in a few lines
  * demonstrating basic kinds of boundary condition (Dirichlet, Neumann, Robin, integral constraints)
  * basic kinds of analysis (steady, unsteady, modal)
  * Together these in sequence form an introduction covering most of the intrinsic functionality
  * Maybe one of these in the README but the rest don't need to be in the package, could be in readthedocs, scikit-fem.github.io, &c.
* More detailed didactic ones
  * Demonstrating technical details of scikit-fem or finite elements (preprocessing, discretization, solution, postprocessing)
  * Forming subsequent chapters of user's guide
  * Doesn't need to be in package
* A gallery
  * More involved specialized problems
  * Not intended to be read in sequence
  * Possibly involving less common equations that require some scientific background or mathematical derivation
  * Can involve secondary packages which are not to be regarded as dependencies of scikit-fem
  * Demonstrate that scikit-fem plays nicely with a large number of other packages
  * Can be big (both large individual examples and many in number), because the gallery should not be part of the package
  * Can involve complicated geometry or mesh
  * Some might have colourful pictures or animations
  * Should be reasonably easily reproducible though, e.g. contain full source code
* Regression test cases
  * Need to be part of the package, though not necessarily installed by default
  * Should only involve the essential dependencies
  * Don't have to be the same as introductory examples, but can be related
  * Probably not the same as the gallery as these tests should run reasonably quickly and don't have to be showy or interesting
  * Might use method of manufactured solutions, which isn't great for introductory examples or gallery
  * This is the suite of tests that has to be passed in continuous integration
* Convergence and performance test cases
  * These tend to have to be bigger than the tests of the previous category to be useful
  * I find convergence tests a very good way of uncovering bugs
[...]

PR #58, opened 2018-09-27
WIP: first Stokes-flow example

gdmcbain:
Here's a start on one of the hydrodynamic examples proposed in #31.

It builds on ex02 (plate), ex12 (pygmsh disk), and ex13 (pygmsh boundaries).  What's new is an apppropriately simplified biharmonic operator and the use of `tricontour` to extract the stream-lines.


kinnala:
Thanks, it looks interesting. I'll check it out in the weekend.

Issue #62, opened 2018-10-02
interpolate Morley


gdmcbain:
The problem is to interpolate a function defined on Morley elements at a specified (*x, y*) point which isn't necessarily a node; e.g. the deflection at the centre of the Kirchhoff plate in ex02, or the value of the stream-function at the centre of the circular-cylindrical cavity in ex18.

Attempting `MeshTri.interpolator` with the `MeshTri` returned by `refinterp` fails #59; the mesh is invalid, having duplicate points.

I think interpolation involves two steps:
1. Find the element (if any) that contains the point.
1. Evaluate the function based on the local basis functions of that element and the elemental degrees of freedom.

and I suspect that in general the first is the harder.  

Thus far, this operation is only implemented in scikit-fem as `MeshTri.interpolator` (P1) and `MeshTri.const_interpolator` (P0) and makes use of :
* [`matplotlib.tri.Triangulation`](https://matplotlib.org/api/tri_api.html?highlight=matplotlib%20tri%20triangulation#matplotlib.tri.Triangulation),
* [`matplotlib.tri.LinearTriInterpolator`](https://matplotlib.org/api/tri_api.html?highlight=matplotlib%20tri%20triangulation#matplotlib.tri.LinearTriInterpolator)
* [`matplotlib.tri.Triangulation.get_trifinder`](https://matplotlib.org/api/tri_api.html?highlight=matplotlib%20tri%20triangulation#matplotlib.tri.Triangulation.get_trifinder)

Perhaps the [`TriFinder`](https://matplotlib.org/api/tri_api.html?highlight=matplotlib%20tri%20triangulation#matplotlib.tri.TriFinder) might be factored out of those so that it could also be used for other elements defined on linear simplicial meshes; e.g., besides Morley, also affine P2.  Looking ahead, with `MeshQuad._splitquads`, perhaps quadrilateral meshes too.

kinnala:
Next, one should use Mapping.invF and Element.gbasis, quite similarly as here: 

https://github.com/kinnala/scikit-fem/blob/167ba29e586c1a2efa7246de6518bc109adf4e80/skfem/assembly/global_basis/facet_basis.py#L99

Instead of leaving the different evaluated basis functions separate (as is done in the previous url), we should sum their values similarly as in GlobalBasis.interpolate.

gdmcbain:
Thanks for the tips; I haven't fully understood the machinery here but continuing to scratch around:

```python
[...]

PR #66, opened 2018-10-04
interpolate at centre rather than using maximum P1 nodal value

gdmcbain:
This is an example of interpolating Morley #62.


gdmcbain:
I think this is ready for inclusion now.  It's much tidier following the resolution of #62 .

Issue #72, opened 2018-10-17
MeshQuad._splitquads sometimes misorients second triangle


gdmcbain:
Consider a mesh with a single quadrilateral:
```
0---1
|   |
2---3
```
i.e.
```python
m = MeshQuad(np.array([[0, 1, 0, 1], [1, 1, 0, 0]]), np.array([[2, 3, 1, 0]]).T)
```
This is just the reference Q1 with differently numbered points (but ordered correctly in `m.t`).

Then
```python
m._splitquads().t
```
returns
```python
array([[0, 0],
       [2, 1],
       [3, 3]])
```
which has the second triangle clockwise.

This doesn't happen for the default quad:
```
3---2
|   |
0---1
```
`MeshQuad()._splitquads().t` returns
```python
 array([[0, 1],
       [1, 2],
       [3, 3]])
```
both anticlockwise.

gdmcbain:
It looks like this happens in the constructor for the new `MeshTri`, in
https://github.com/kinnala/scikit-fem/blob/740021f05497a05be6a677866328d6679c66e635/skfem/mesh/mesh2d/mesh_tri.py#L130

where in all calls from `_splitquads`, `sort_t==True`.

gdmcbain:
The sorting of the indices of the vertices of each triangle looks suspicious:

https://github.com/kinnala/scikit-fem/blob/740021f05497a05be6a677866328d6679c66e635/skfem/mesh/mesh2d/mesh_tri.py#L228

gdmcbain:
I'm not sure what the intention of the sorting is here, but perhaps what's in order is to cycle each column so that it begins with its least entry? i.e. entry `self.t.argmin(axis=0)`.

gdmcbain:
[numpy.roll](https://docs.scipy.org/doc/numpy/reference/generated/numpy.roll.html#numpy.roll) looks like it might do that.

gdmcbain:
As in:
```python
    self.t = np.roll(self.t, self.t.argmin(axis=0), 0)
```

kinnala:
Fixing this by removing the sort would require quite a lot of work, see https://github.com/kinnala/scikit-fem/pull/73#issuecomment-430503834

Do you find it problematic that the input orientation is not preserved?

gdmcbain:
[...]

PR #73, opened 2018-10-17
roll don't sort, to preserve orientations in constructing triangular meshes

gdmcbain:
This is for #72, but note that I don't fully comprehend what the sorting (or rolling) is exactly for.

### DIFF
--- a/skfem/mesh/mesh2d/mesh_tri.py
+++ b/skfem/mesh/mesh2d/mesh_tri.py
@@ -223,9 +223,9 @@ class MeshTri(Mesh2D):
         return cls(p, t)
 
     def _build_mappings(self, sort_t=True):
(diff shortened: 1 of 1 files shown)


gdmcbain:
Hmm.  Besides the order of magnitude error in ex02 revealed by Travis, there's a worse one in ex18:
> psi0 = 0.0007402344108927771 (cf. exact = 1/64 = 0.015625)

I conclude that this PR is wrong.

gdmcbain:
Thank you for the explanation. I'm trying to think this through.

I'm not seeing how this applies to the second example in #72, the splitting of the default quadrilateral, `MeshQuad()._splitquads()`. That seems to have the diagonal traversed in opposite senses in the two triangles: [1, 3] and [3, 1].

PR #74, opened 2018-10-17
import plt in mesh_line

gdmcbain:
Otherwise ex15 breaks
    
https://github.com/kinnala/scikit-fem/blob/740021f05497a05be6a677866328d6679c66e635/examples/ex15.py#L31


kinnala:
We really should add the rest of the examples to the test suite somehow.

PR #77, opened 2018-10-29
re-solve side-heated circular cavity with Taylor-Hood

gdmcbain:
This follows overnight discussion of Taylor-Hood on Gitter.

Indeed, P2-P1 is much easier than Q2-P1 since then both variables share the same mesh.

So far only the pressure is plotted; being scalar and P1 this is easy with current facilities.  The exact answer is that the isobars should be rectangular hyperbolae, i.e. level-sets of _xy_.  This looks about right.

For further details of exact solution, see 
* G. D. McBain 2016 Creeping convection in a horizontally heated ellipsoid. In _Proceedings of the Twentieth Australasian Fluid Mechanics Conference,_ Paper 548. ([PDF](http://people.eng.unimelb.edu.au/imarusic/proceedings/20/548%20Paper.pdf))

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,12 @@
+name: scikit-fem-dev
+dependencies:
+  - python=3.7
+  - numpy
+  - scipy
+  - matplotlib
+  - ipython
+  - sphinx
+  - pip:
+    - meshio
+    - pygmsh
+    - sphinx_rtd_theme
--- a/examples/taylor_hood.py
+++ b/examples/taylor_hood.py
@@ -1,3 +1,28 @@
+"""This solves for the same creeping flow as ex18 in primitive variables,
+
+i.e. velocity and pressure instead of stream-function.  These are governed by
+the Stokes momentum
+
+.. math::
+    0 = -\rho^{-1}\nabla p + \mathbf f \nu\Delta\mathbf u
+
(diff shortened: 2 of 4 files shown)


gdmcbain:
Still thinking about how to postprocess the P2 vector velocity.

kinnala:
Nice job.

For visualising the flow field a simple approach would be
```
ux = velocity[basis['u'].nodal_dofs[0, :]]
uy = velocity[basis['u'].nodal_dofs[1, :]]
plt.quiver(mesh.p[0, :], mesh.p[1, :], ux, uy, np.sqrt(ux^2 + uy^2))
```
I didn't test it, but the idea is to simply discard the edge dofs and use only nodal data in the visual.

gdmcbain:
Thanks, yes, the .nodal_dofs attribute is just the ticket.  There's no problem in discarding the nonnodal arrows; vector fields usually need to be thinned out anyway for quiver-plots, as in ex18.

kinnala:
This seems to be complete? I'll check it out asap.

gdmcbain:
O. K. the point-pressures look O. K.; the exact solution is _xy_ / 2.

kinnala:
I'm renaming and adding this to docs.

Issue #89, opened 2018-11-05
piecewise-definition of coefficients on subdomains


gdmcbain:
Consider a Laplace equation with variable coefficient, &nabla;&sdot; (_k_ &nabla; _u_) = 0, with _k_ piecewise-constant on ‘regions’ or ‘subdomains’; e.g. steady conduction of heat between two blocks of different thermal conductivity (with no thermal contact resistance).

I think that such problems could also be done with mortar methods (which I see from ex04 are implemented, at least for two-dimensional problems, via `InterfaceMesh1D`); however, monolithic approaches can be convenient too.  The issue here is to assign different values to _k_ depending on which submesh an element belongs to, e.g. as defined in Gmsh as `Physical Surface` (in two dimensions).  Should _k_ be defined with `ElementTriP0`?  And then populated using the `cell_data` returned by `pygmsh.generate_mesh` or the `.cell_data` attribute of the `meshio.Mesh` returned  by `meshio.read`&mdash;I don't think `skfem.mesh.from_meshio` currently does this, it only looks for tags on facets, e.g. for mixed boundary conditions.

kinnala:
The additional parameter `w` given to `asm`, i.e.
```python
asm(bilinf, basis, w=w)
```
has shape number-of-elements x number-of-quadrature-points.

Each row corresponds to a single element. If the material field is constant within each element, you can simply create 1D array of length `mesh.t.shape[1]` (number-of-elements) and replicate this by the number of quadrature points. The result can be passed to `asm`.

kinnala:
Ok, now that I meditated this for a while: I think we should create a method to `GlobalBasis` (say `GlobalBasis.indicator`) which takes in Submesh and outputs 2-d array which can be passed to `asm` via the keyword argument `w`.

This 2-d array represents the indicator function, i.e. it is 1 if the respective quadrature point belongs to the Submesh and 0 otherwise.

These indicator function arrays can be multiplied with specified values and summed together at will.

What do you think?

gdmcbain:
That would be very useful for switching on a term in a subdomain, yes.   Sticking with the example of the heat equation for concreteness, it's very common for tbe source term to be confined to a region.  The Joule-heated wire in an insulator is a textbook case.

For the piecewise-constant conductivity though, this style would involve one `asm` for each region and there could be a few of those, so the P0 approach, simplified as above ('simply create') would be more convenient.

[...]

PR #90, opened 2018-11-05
WIP: script demonstrating piecewise-constant definition of coefficient 

gdmcbain:
This is one way to do #89.


gdmcbain:
It's kind of one-dimensional but done with MeshTri as I wasn't sure that the facilities of MeshLine were the same and I was ultimately interesting myself in two-dimensional problems.

docs/examples/piecewise.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@bilinear_form
+def conduction(u, du, v, dv, w):
+    return w.w * sum(du * dv)
+
+
+k = np.ones(basis.N) * lengths[1]
+k[mesh.elements_satisfying(lambda x, y: x < 0)] = lengths[0]
+L = asm(conduction, basis, w=basis0.interpolate(k))
```
kinnala:
```suggestion
X, Y = basis.global_coordinates()
k = 3 * (X < 0) + 2 * (X > 0) 
L = asm(conduction, basis, w=k)
```
This should also work.

kinnala: Note: I should've read your issue more carefully. This approach is not as general as you probably require if you eventually want to use Submeshes to define k.

gdmcbain:
Maybe instead of https://github.com/kinnala/scikit-fem/blob/e52e0ee1fd4fe0c1cd81365bc9700c04eb58cfc0/docs/examples/insulated.py#L38
it would be nicer to specify something like
```python
thermal_conductivity = {'wire': 101., 'insulation': 11.}
```
and read in the `field_data` from `meshio.read` which here contains
```python
{'convection': array([3, 1]), 'wire': array([1, 2]), 'insulation': array([2, 2])}
```
and can be parsed with
```python
{k: i for k, (i, d) in mesh.field_data.items() if d == mesh.dim()}
```
to give the mapping
```python
{'wire': 1, 'insulation': 2}
```

kinnala:
I'll check this out and merge it tomorrow while trying to make the part where you modify mesh.py nicer.

PR #95, opened 2018-11-12
WIP: input subdomain-dependent coefficients as Dict[str, float]

gdmcbain:
This modifies docs/examples/insulated.py #90 to improve the user-level syntax for piecewise-defined coefficients #89.  

No modification of the library is required, just the example.

Hitherto, the script made use of knowledge of the integer identifiers given to the subdomains by Gmsh but these identifiers can't be specified in pygmsh and I'm not sure that it's safe to rely on them being numbered consecutively from one; a `dict` is more readable anyway.

This also demonstrates the use of the names of subdomains in constructing an indicator function (actually elemental indicator ndarray).

### DIFF
--- a/docs/examples/insulated.py
+++ b/docs/examples/insulated.py
@@ -34,7 +34,7 @@ from skfem.utils import solve
 radii = [2., 3.]
 joule_heating = 5.
 heat_transfer_coefficient = 7.
-thermal_conductivity = np.array([101.,  11.])
+thermal_conductivity = {'wire': 101.,  'insulation': 11.}
 
 
 def make_mesh(a: float,         # radius of wire
@@ -56,7 +56,9 @@ def make_mesh(a: float,         # radius of wire
     return mesh
 
 mesh = make_mesh(*radii)
-regions = mesh.external.cell_data['triangle']['gmsh:physical'] - 1
+regions = mesh.external.cell_data['triangle']['gmsh:physical']
+region_ids = {i: name for name, (i, d) in
+              mesh.external.field_data.items() if d == mesh.dim()}
 
 @bilinear_form
 def conduction(u, du, v, dv, w):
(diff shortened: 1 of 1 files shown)


kinnala:
You should now be able to do
```python
conductivity = np.zeros((basis.nelems, len(basis.W)))
for subdomain, indices in mesh.subdomains.items():
    np.add.at(conductivity, indices, thermal_conductivity[subdomain])
```
and then pass `conductivity` to `asm` via the keyword `w`.

I did not test this yet.

kinnala:
The first line of my previous snippet is kinda technical, so maybe the following method in GlobalBasis would be helpful:
```python
    def empty_w(self) -> ndarray:
        """Return an empty array with correct dimensions
        for :func:`~skfem.assembly.asm`."""
        return np.zeros((self.nelems, len(self.W)))
```

kinnala:
I did not perform these modifications btw, I think the current implementation is fine.

gdmcbain:
Thanks!  I had been looking for something like `add.at` but didn't realized that it was hidden in [`numpy.ufunc.at`](https://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.ufunc.at.html).  

This could also be used to replace the familiar
```python
x[I] = solve(*condense(A, b, I=I))
```
with 
```python
np.add.at(x, I, solve(*condense(A, b, I=I)))
``` 
as in https://github.com/gdmcbain/scikit-fem/blob/46da27f8f98c54fac0233cd97982a09e508bd132/docs/examples/ex12.py#L49

(Possibly not actually an improvement, but good to know.)

gdmcbain:
And thanks!  I had not noticed the addition of `skfem.Mesh.subdomains` via the `._parse_submeshes` method.  I'll look into that and very likely make use of it here.

Issue #86, opened 2018-11-03
Cholesky factorization for fast time-stepping of symmetric problems


gdmcbain:
For problems like the heat equation with say Crank&ndash;Nicolson discretization in time and constant time-step, the resulting matrix pencil is the same at each step so it is appealing to factorize it once and for all; e.g. using the Cholesky factorization (since the matrices are symmetric positive-definite).

I see that `cholesky` was imported from `sksparse.cholmod` in `sp.fem` 

https://github.com/kinnala/sp.fem/blob/1398225075c9ab90dbc78616a451f7dd9139ff4c/spfem/utils.py#L71 

(though the factors were not returned or stored) and propose to use that again in scikit-fem.

gdmcbain:
Perhaps this should not be part of the scikit-fem library itself (to avoid another dependency, and a tricky one at that) but just demonstrated in an example; I should be able to do that early next week.

gdmcbain:
It is convenient here that [`skfem.utils.condense`](https://github.com/kinnala/scikit-fem/blob/f4722043b8106371e2f49366dcc0f284454c2e85/skfem/utils.py#L17) preserves symmetry.

kinnala:
I have not included any built-in support since I thought CHOLMOD was GPL licensed. Now that I briefly rechecked, it seems that it's, in fact, LGPL licenced so that dynamic linking should be fine. I have to verify this.

kinnala:
It seems to me that entire SuiteSparse is LGPL and it makes me wonder why anyone would buy a commercial license. Oh well, nevertheless, example is totally fine.

gdmcbain:
Thanks for checking. I had forgotten about the legal issue, though that is the _raison d'être_ of scikit-sparse.

gdmcbain:
I do also have a pure Python scipy.sparse implementation which I had to write because it was too hard to rely on CHOLMOD for deployment to Microsoft Windows. From memory that was workable; although slower, it didn't matter because the one factorization was amortized over thousands of time-steps. I'll see what the situation is with Windows these days.

gdmcbain:
> This might work better in Windows: https://anaconda.org/conda-forge/scikit-sparse (#98)

Sadly no.  In a fresh Python 3.7 _Anaconda for Windows_ environment terminal in _Microsoft Windows 10_ I tried
```shell
conda install -c conda-forge scikit-sparse
```
but that failed
```
Solving environment: failed

[...]

PR #98, opened 2018-11-12
WIP: example demonstrating use of sksparse.cholmod.cholesky #86

gdmcbain:
This is a fairly minimal two-dimensional example of the heat equation, again taking a problem with a closed-form solution from Carslaw & Jaeger's _Conduction of Heat in Solids_.

Apart from the difficulty of installing CHOLMOD on Microsoft Windows mentioned [previously](https://github.com/kinnala/scikit-fem/issues/86#issuecomment-435642098), the example brings a couple of minor items to light:
* `sksparse.choldmod.cholesky` prefers `csc_matrix` so the `csr_matrix` assembled has to be transposed
* If there are Dirichlet conditions, the right-hand side needs to be condensed at each time-step whereas the matrix only needs to be condensed once, before the Cholesky factorization.  Since `skfem.utils.condense` requires a matrix, I've passed an empty sparse matrix of the appropriate size.  It might be tidier to factor the `condense` function so that the right-hand side can be condensed directly.

### DIFF
--- a/docs/examples/cholesky.py
+++ b/docs/examples/cholesky.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+
+"""Decay of the fundamental Dirichlet mode in a rectangle.
+
+*This example requires :module:`sksparse`.*
+
+The solutions of the heat equation
+
+.. math::
+    \frac{\partial T}{\partial t} = \kappa\Delta T
+
+in various tensor-product domains consist of the sum of modes which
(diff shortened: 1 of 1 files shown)


kinnala:
This might work better in windows: https://anaconda.org/conda-forge/scikit-sparse

kinnala:
Nice example, thanks for this.

I'll try to add these new examples to the documentation at some point.

Issue #112, opened 2019-01-11
coordinate-dependent Dirichlet data for other than P1


gdmcbain:
The specification of inhomogeneous Dirichlet data for the Laplace equation is demonstrated in ex14.py for `ElementTriP1`:
https://github.com/kinnala/scikit-fem/blob/9c0cf416745941abd72f70865351ef0e095f83b2/docs/examples/ex14.py#L37

however, `Mesh.p` only contains the nodal coordinates, so this doesn't work if the element is changed to say `ElementTriP2`.

How might one go about specifying inhomogeneous Dirichlet data for elements with degrees of freedom that aren't all located at nodes of the mesh?  Does this relate to #84?  The P2 degrees of freedom on the boundary are available through `basis.get_dofs().all()`.

I suppose one could fall back on other techniques like:
* penalization, i.e. approximating the Dirichlet condition with a Robin condition (Ern & Guermond [2002](https://www.springer.com/gp/book/9783540426158), &sect;9.5.5)
* Lagrange multiplier (Stenberg [1995](http://dx.doi.org/10.1016/0377-0427(95)00057-7))

kinnala:
You are correct about #84. Currently you have to use your expertise on the location of the DOF's, i.e. edge midpoints. You could find all boundary edges, find their midpoints, and use this info to assign DOF values. This is cumbersome, hence #84. If you need this feature, I could try to find time this weekend but cannot promise yet. 

Both alternative methods you mentioned should also work and be supported. 

There are at least two additional ways: Nitsche's method (see interior penalty example in docs, this I'd prefer over penalty) and using mixed formulation where Dirichlet condition is natural (i.e. mixed Poisson with H(div) elements).

gdmcbain:
No, I don't need this feature at the moment, just thought it might be a good addition to the suite. A sequence of examples demonstrating all of the above techniques would be worthwhile.

I noticed this while looking into generalizing the membrane example from P1 to P1 or P2, i.e. composing complement_dofs with get_dofs instead of relying on interior_nodes. That works fine for the homogeneous Dirichlet condition.

gdmcbain:
I changed the title as constant inhomogeneous Dirichlet data is easy too; e.g.
https://github.com/kinnala/scikit-fem/blob/9c0cf416745941abd72f70865351ef0e095f83b2/docs/examples/ex13.py#L60

gdmcbain:
[...]

Issue #84, opened 2018-10-31
Finding dof locations 


kinnala:
One should be able to easily find global DOF locations using facilities such as InteriorBasis.get_dofs.

- Make local DOF locations to each element
- Map these to global coordinates
- Make nice interface for fetching them

gdmcbain:
In #113, I noticed another case where this hurts.  Besides coordinate-dependent Dirichlet data #112, in one of the Laplace examples, the exact solution was constructed using `Mesh.p` which only gave nodal values.
 https://github.com/kinnala/scikit-fem/blob/5f1687d32f67e014d21ff477b82eea30332c796a/docs/examples/ex13.py#L63
This is gotten around in #113 by L²-projection, which might be a better idea anyway.  That's easy as we do have access to the coordinates of the quadrature points when writing a linear form.

kinnala:
I guess one could also implement Dirichlet conditions via an explicit L2 projection. This might introduce consistency error but I'm not sure.

gdmcbain:
I hadn't thought of that Dirichlet via L2.  Should work?  Why do you think that it might introduce consistency error?  I suppose I would have usually thought that L²-projection was "truer" (in some functional-analytic sense which I probably can't precise) that mere nodal interpolation.  Is it that to avoid increasing the consistency error a different kind of projection is called for, suited to the correct trace space?

Would this end up giving identical answers (assuming all linear algebraic systems solved exactly) to the Lagrange-multiplier method, except that it's done in two stages (boundary then interior)?  My understanding is that that effectively finds the flux such that the boundary values weakly satisfy the Dirichlet condition, i.e. w.r.t. the L² inner product, and I think that in a finite element space that set of boundary values is unique.

One to add to the sequence of examples proposed in #112 anyway, perhaps.

gdmcbain:
It looks like the coordinates of the DOFs in ElementTriP2 are given by those of the nodes of the mesh followed by the midpoints of the facets:
```python
np.hstack([mesh.p, mesh.p[:, mesh.facets].mean(axis=1)])
```

This matches, for example, what's obtained by _L_&sup2; projection using the `w.x` available inside a `linear_form`.

gdmcbain:
It looks like the coordinates of the DOFs in ElementTriP2 are given by those of the nodes of the mesh followed by the midpoints of the facets:
```python
np.hstack([mesh.p, mesh.p[:, mesh.facets].mean(axis=1)])
```

[...]

PR #113, opened 2019-01-14
upgrade ex12, 13 from P1 to P2

gdmcbain:
While thinking about #112 and #84, I generalized a couple of the Poisson examples so that they could use either ElementTriP1 or ElementTriP2, without other modification.  The issues here all  stem from #84: that one doesn't know the coordinate-locations corresponding to the nonnodal degrees of freedom in the quadratic elements.

### DIFF
--- a/docs/examples/ex12.py
+++ b/docs/examples/ex12.py
@@ -43,7 +43,8 @@ basis = InteriorBasis(m, e, map, 2)
 A = asm(laplace, basis)
 b = asm(unit_load, basis)
 
-I = m.interior_nodes()
+D = basis.get_dofs().all()
+I = basis.complement_dofs(D)
 
 x = 0*b
 x[I] = solve(*condense(A, b, I=I))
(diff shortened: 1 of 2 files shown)


kinnala:
You might want to remove the fixed integration order when initializing InteriorBasis now that you have higher order polynomials. In case of mass matrix in ex13, you are integrating P2 * P2 = P4 which would require intorder=4 for exactness. 

I'll check this out more carefully later.

kinnala:
I'd love to have these in the test suite (tests/test_examples.py), but we'd need to either have the additional dependencies installed to travis or save the mesh just for testing purposes and bypass the mesh generation in the test suite only.

gdmcbain:
Thanks. Sorry, I didn't think about quadrature at all, just eyeballed the answers and plots.

gdmcbain:
Although a major part of the original motivation for ex12 was to demonstrate how to use pygmsh for scikit-fem, that is since covered in other examples; it would be easy enough to switch to dmsh, which is a much lighter dependency.

gdmcbain:
I figured that since I wasn't trying to make any specific point about quadrature in these two examples, it might make more sense to just fall back on the defaults.  (Ditto for `InteriorBasis.mapping`.)

PR #117, opened 2019-01-24
use numpy's linalg.norm and diff in Mesh2d.param

gdmcbain:
It might be good to expand the docstring too, possibly with a reference.


gdmcbain:
I started looking through the treatises on the finite element method for parameters characterizing meshes.  Braess's _Finite Elements_ (2nd edn, 2001, &sect;II.5) refers to the inradius and the diameter.

I note that for a simplex the diameter is equal to the length of the longest side but not for more general elements like quadrilaterals or hexahedra.

PR #120, opened 2019-01-31
use __file__ to work out where to load mesh from

gdmcbain:
To load a mesh from a file for an example, specifying the path can be tricky.  The following

https://github.com/kinnala/scikit-fem/blob/7dbe3cbc6f1e11188d72337a1e4577a4f8a1cbaa/docs/examples/ex21.py#L6

only works if the example is run from the top-level directory as
```shell
python docs/examples/ex21.py
```

but of course failed when I ran it in `docs/examples`.

There are facilities in the standard library for getting around this.  The one adopted here might even work on non-POSIX systems.

### DIFF
--- a/docs/examples/ex21.py
+++ b/docs/examples/ex21.py
@@ -3,7 +3,9 @@ from skfem.models.elasticity import linear_elasticity,\
                                     lame_parameters
 import numpy as np
 
-m = MeshTet.load("docs/examples/beams.msh")
+from pathlib import Path
+
+m = MeshTet.load(str(Path(__file__).with_name("beams.msh")))
 e1 = ElementTetP1()
 e = ElementVectorH1(e1)
 


gdmcbain:
Slightly annoying to have to recast the `pathlib.Path` as `str`, but that's a limitation of `meshio.read` which I might look into upstream rather than in `skfem.Mesh.load`.

kinnala:
This looks like a reasonable addition.

gdmcbain:
> Slightly annoying to have to recast the `pathlib.Path` as `str`, but that's a limitation of `meshio.read` which I might look into upstream rather than in `skfem.Mesh.load`.

nschloe/meshio#538 is a request to ‘Accept os.PathLike as well as str in read function’.

clbarnes:
> [nschloe/meshio#538](https://github.com/nschloe/meshio/issues/538) is a request to ‘Accept os.PathLike as well as str in read function’.

And https://github.com/nschloe/meshio/pull/545 is a PR implementing it :smile:

clbarnes:
And https://github.com/nschloe/meshio/pull/546 is a now-merged PR by the owner doing the same thing :tada:

Issue #123, opened 2019-02-11
save vector solution to VTK, XDMF, or similar


gdmcbain:
I expect this involves meshio.write and will involve just passing the right thing to `skfem.Mesh.save`, but I haven't worked out how to save, e.g. the three-dimensional displacement `u` in ex11 or two-dimensional `velocity` in ex18 for investigation in ParaView or similar when saving to `*.vtk`, `*.xdmf`, &c.

I expect the fulfillment of this will just be an addition to an example to show how it's done rather than modification of the library, unless it is deemed worthwhile to enhance `Mesh.save` to facilitate whatever transformation might be required.

gdmcbain:
In ex18, the right shape seems to be
```python
mesh.save('velocity.vtk', velocity[basis['u'].nodal_dofs.T])
```

gdmcbain:
…except that a vector in VTK needs to have three components, so more like
```python
mesh.save('velocity.vtk', np.vstack([velocity[basis['u'].nodal_dofs],
                                     np.zeros_like(mesh.p[0])]).T)
```

gdmcbain:
In three dimensions, as in ex11, of course the padding isn't needed, just the transposition
```python
m.save(f'displacement.vtk', u[ib.nodal_dofs].T)
```

gdmcbain:
XDMF doesn't work like VTK does, but I expect that that's an issue for meshio and not here.

gdmcbain:
There might be something to be done upstream but a workaround is demonstrated in #124.

gdmcbain:
This is now fixed upstream in  nschloe/meshio#326 so that should be exploited here.

PR #124, opened 2019-02-11
demonstrate saving a 2-D vector solution to VTK

gdmcbain:
This fulfils #123.


gdmcbain:
I have flagged this issue upstream in nschloe/meshio#325 and proposed a fix in nschloe/meshio#326; if accepted that will render this redundant but not break it.

PR #131, opened 2019-02-28
use __file__ so that tests can be run from anywhere

### LINKED ISSUE #130: test_mesh fails if run other than in top level
While working towards coping with MSH4.1 #129, on nschloe/meshio#330 to get nschloe/meshio#331 accepted, I was in the meshio source tree and tried to run the scikit-fem test suite to check that I wasn't breaking anything here; i.e.
```shell
pytest -n auto ../scikit-fem/tests
```
But then `test_mesh.Loading.run_test` tries

https://github.com/kinnala/scikit-fem/blob/f17d4366b6ee22c57532842339bf8bced120e2d9/tests/test_mesh.py#L49

which of course raises
> FileNotFoundError: [Errno 2] No such file or directory: 'docs/examples/box.msh'

gdmcbain:
Same idea as #120; fixes #130.

### DIFF
--- a/tests/test_mesh.py
+++ b/tests/test_mesh.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 import unittest
 import numpy as np
 from skfem.mesh import *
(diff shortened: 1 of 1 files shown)


kinnala:
Is it possible to parse strings like `'../docs/examples/box.msh'` into these path objects and do this automatically in `Mesh.load`?

gdmcbain:
I think that '.' is interpreted relative to the working directory rather than `__file__`. Both are useful points of reference. It might be confusing or obstructive  to reinterpret '.'

gdmcbain:
A different question is `pathlib` versus `os.path`. I think, if one doesn't like the former, one can do something similar with `os.path.join` or even string-concatenation like `__file__ + '/../docs/examples/box.msh`.

PR #132, opened 2019-03-04
demonstrate L2_projection for scalar coordinate-dependent Dirichlet data

gdmcbain:
Before adopting `skfem.utils.L2_projection` from #112 where it's really needed, in ex24.py for the P2 parabolic inlet velocity profile, I thought I'd try it in the simplest example with scalar coordinate-dependent Dirichlet data, ex14.py

https://github.com/kinnala/scikit-fem/blob/71edb0bcaf98e8f1c72a07a8c74dbbf08e3db4d8/docs/examples/ex14.py#L23

This should enable promoting the element from 

https://github.com/kinnala/scikit-fem/blob/71edb0bcaf98e8f1c72a07a8c74dbbf08e3db4d8/docs/examples/ex14.py#L6

to `elementTriP2`, which otherwise fails.


gdmcbain:
This is O. K., but where I've put 
```python
u = L2_projection(dirichlet, basis)
```
I would have thought that I could restrict it to the boundary with
```python
u[D] = L2_projection(dirichlet, basis, D)
```
but no.

kinnala:
I would have expected that to work. There might be a bug.

kinnala:
Aha, I think you might have to use FacetBasis for that to work. Otherwise you'll get something unexpected.

gdmcbain:
> Aha, I think you might have to use FacetBasis for that to work. Otherwise you'll get something unexpected.

I've been puzzling over this.  I mean FacetBasis does work as expected, but I was puzzling over why ElementBasis didn't.  Is it because of numerical quadrature?  Because the quadrature rules for triangles don't have abscissæ on the edges?

kinnala:
When using InteriorBasis, you have to interpret the operation as L2
projection where the integral is over the domain. (With FacetBasis the
integral is over the boundary.) If you restrict this projection to boundary
nodes, you still have the associated basis functions that are nonzero also
on the elements that are closest to the boundary. Because these go quickly
to zero inside the domain, I expect the resulting projected function to
greatly overshoot (have higher than expected values) at the boundary.

gdmcbain:
Thank you.

PR #133, opened 2019-03-04
FacetBasis in L2_projection for coordinate-dependent Dirichlet data

gdmcbain:
…as suggested by @kinnala #132

### DIFF
--- a/docs/examples/ex14.py
+++ b/docs/examples/ex14.py
@@ -1,6 +1,8 @@
 from skfem import *
 from skfem.models.poisson import laplace
 
+import numpy as np
+
 m = MeshTri()
 m.refine(4)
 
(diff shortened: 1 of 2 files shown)


gdmcbain:
A couple of things that might be worth adding to ex14 or its documentation:

* The example now works with `e = ` either `ElementTriP1` or `ElementTriP2`, with no other change required.  It might be worth showing that somehow.

* And of course comparing the error.  With `ElementTriP2`, the numerical solution is exact since the exact solution is of degree two and so lies within the finite element function space.  Maybe point this out.  There's a bit of a fuss made about this in the FEniCS tutorial, under the heading [_Choosing a test problem_](https://fenicsproject.org/pub/tutorial/html/._ftut1004.html#ch:poisson0:testproblem):

>  **Tip: Try to verify your code with exact numerical solutions!** A common approach to testing the implementation of a numerical method is to compare the numerical solution with an exact analytical solution of the test problem and conclude that the program works if the error is "small enough". Unfortunately, it is impossible to tell if an error of size $10^{−5}$ on a 20×20 mesh of linear elements is the expected (in)accuracy of the numerical approximation or if the error also contains the effect of a bug in the code. All we usually know about the numerical error is its asymptotic properties, for instance that it is proportional to _h_ ² if _h_ is the size of a cell in the mesh. Then we compare the error on meshes with different _h_-values to see if the asymptotic behavior is correct. This is a very powerful verification technique and is explained in detail in the section [_Computing convergence rates_](https://fenicsproject.org/pub/tutorial/html/._ftut1020.html#ch:poisson0:convrates). However, if we have a test problem for which we know that there should be no approximation errors, we know that the analytical solution of the PDE problem should be reproduced to machine precision by the program. That is why we emphasize this kind of test problems throughout this tutorial. Typically, elements of degree _r_ can reproduce polynomials of degree r exactly, so this is the starting point for constructing a solution without numerical approximation errors.

PR #135, opened 2019-03-13
use __file__ to work out where to savefig 

gdmcbain:
as in #120 

This time I was experimenting in `docs/examples` and tried to run ex05.py for comparison; it didn't work because of
https://github.com/kinnala/scikit-fem/blob/c8ad891eb94532aa51bd634b6bd9a05622f81400/docs/examples/ex05.py#L50

### DIFF
--- a/docs/examples/ex01.py
+++ b/docs/examples/ex01.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from skfem import *
 
 m = MeshTri()
@@ -24,4 +25,4 @@ x[I] = solve(*condense(A, b, I=I))
 
 if __name__ == "__main__":
     m.plot(x, smooth=True, colorbar=True)
-    m.savefig("docs/examples/ex01_solution.png")
+    m.savefig(Path(__file__).stem + '_solution.png')
--- a/docs/examples/ex02.py
+++ b/docs/examples/ex02.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from skfem import *
 import numpy as np
 
@@ -61,4 +62,4 @@ if __name__ == "__main__":
     M, X = ib.refinterp(x, 3)
     ax = m.draw()
     M.plot(X, smooth=True, ax=ax, colorbar=True)
-    M.savefig("docs/examples/ex02_solution.png")
+    M.savefig(Path(__file__).stem + '_solution.png')
--- a/docs/examples/ex05.py
+++ b/docs/examples/ex05.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from skfem import *
 
 m = MeshTri()
@@ -47,4 +48,4 @@ x[I] = solve(*condense(K, f, I=I))
 
 if __name__ == "__main__":
     m.plot(x[:-1], colorbar=True)
-    m.savefig("docs/examples/ex05_solution.png")
+    m.savefig(Path(__file__).stem + '_solution.png')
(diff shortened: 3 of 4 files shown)


docs/examples/ex01.py
```diff
@@ -1,3 +1,4 @@
+from pathlib import Path
```
kinnala: This breaks the line number based `literalinclude` in ex01.rst. Either fix the line numbers there or add keyword argument to `Mesh.savefig` which expands the filename using pathlib. E.g. `m.savefig("{}_solution.png", expand_stem=True)` or `m.savefig("{}/ex01_solution.png", expand_dir=True)`. If you decide to modify `Mesh.savefig` then feel free to experiment with different parameters. I'm happy as long as the changes are backwards compatible.

docs/examples/ex02.py
```diff
@@ -1,3 +1,4 @@
+from pathlib import Path
```
kinnala: Same comment regarding `literalinclude` in ex02.rst as above.

docs/examples/ex01.py
gdmcbain: > This breaks the line number based literalinclude in ex01.rst.

Oops.  Sorry.  I'll fix that in a moment.

gdmcbain: > which expands the filename using pathlib. E.g. m.savefig("{}_solution.png", expand_stem=True) or m.savefig("{}/ex01_solution.png", expand_dir=True). If you decide to modify Mesh.savefig then feel free to experiment with different parameters. I'm happy as long as the changes are backwards compatible.

This is a bit tricky.  It's not [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html#pathlib.Path) that does the magic, it's the builtin `__file__`; I had begun typing the following into `Mesh.savefig`:
```python
        if expand_dir:
            path = path.format(Path(__file__).parent)
 ```
when I realized that there `__file__` would expand to `".../skfem/mesh/mesh.py"` which is not what's wanted.

The main problem here is to make the examples work correctly regardless of where they're called from, loading meshes
https://github.com/kinnala/scikit-fem/blob/c8ad891eb94532aa51bd634b6bd9a05622f81400/docs/examples/ex21.py#L8
 and saving figures 

 https://github.com/kinnala/scikit-fem/blob/c8ad891eb94532aa51bd634b6bd9a05622f81400/docs/examples/ex23.py#L97

from and to the directory containing the script, `…/scikit-fem/docs/examples`.  The only two ways I can think of doing this are via `sys.argv[0]` (which will contain the path of the script as a str) as in 
https://github.com/kinnala/scikit-fem/blob/c8ad891eb94532aa51bd634b6bd9a05622f81400/docs/examples/ex17.py#L87
and `__file__` as above.

File: docs/examples/ex02.py (hunk 2 above)
Comment (gdmcbain) [reply docs/examples/ex02.py]: What about replacing `:lines:` for [`literalinclude`](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) with `:start-at:` and `:end-at:`?   For example:
```python
.. literalinclude:: ex02.py
    :start-at: skfem
    :end-at: refine
```

Otherwise I think every `:lines:` field downpage of an edit has to be amended.  These are also more readable in the .rst source files.

kinnala: That is a good idea. I didn't know about this option.

gdmcbain:
A few of the examples still save a figure to the working directory.  This usually works but not if the figures are required for the documentation. Those examples currently just have .rst stubs without figures. They can be brought into conformity with the present issue when the stubs are filled out.

PR #136, opened 2019-03-13
set parabolic inlet profile using L2_projection

### LINKED ISSUE #112: coordinate-dependent Dirichlet data for other than P1

gdmcbain:
Fixes #112 for a case with a vector basis.

### DIFF
--- a/docs/examples/ex24.py
+++ b/docs/examples/ex24.py
@@ -76,7 +76,7 @@ inlet_dofs = np.concatenate([inlet_dofs_.nodal[f'u^{1}'],
 
 def parabolic(x, y):
     """return the plane Poiseuille parabolic inlet profile"""
-    return np.stack((4 * y * (1. - y), np.zeros_like(y)))
+    return ((4 * y * (1. - y), np.zeros_like(y)))
 
 
 
--- a/docs/examples/ex24.py
+++ b/docs/examples/ex24.py
@@ -76,7 +76,7 @@ inlet_dofs = np.concatenate([inlet_dofs_.nodal[f'u^{1}'],
 
 def parabolic(x, y):
     """return the plane Poiseuille parabolic inlet profile"""
-    return np.moveaxis(np.dstack([4 * y * (1. - y), np.zeros_like(y), ]),
+    return np.moveaxis(np.dstack([4 * y * (1. - y), np.zeros_like(y)]),
                        [0, 1, 2], [1, 2, 0])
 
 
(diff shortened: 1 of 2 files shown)


gdmcbain:
I wasn't at all sure what shape the `fun` passed to `L2_projection` should return in the case of a vector basis.  For scalar basis (ex14), it's the same shape as the `x` and `y` passed in.

I got ex24 working here, but only after extensive trial and error with 

```python
def parabolic(x, y):
    """return the plane Poiseuille parabolic inlet profile"""
    return np.moveaxis(np.dstack([4 * y * (1. - y), np.zeros_like(y), ]),
                       [0, 1, 2], [1, 2, 0])
```

I don't think the `moveaxis` can be the right way to do this.

gdmcbain:
On reading the library source, I found
https://github.com/kinnala/scikit-fem/blob/c8ad891eb94532aa51bd634b6bd9a05622f81400/skfem/assembly/asm.py#L108-L109

so figured a better approach might be

```python
def parabolic(x, y):
    """return the plane Poiseuille parabolic inlet profile"""
    return np.transpose(np.dstack([4 * y * (1. - y), np.zeros_like(y)]),
                        (2, 0, 1))
```

It gives the same answer.  Still doesn't seem quite right though.

gdmcbain:
Ah, how about:
```python
def parabolic(x, y):
    """return the plane Poiseuille parabolic inlet profile"""
    return np.stack((4 * y * (1. - y), np.zeros_like(y)))
```

gdmcbain:
Of course `stack` is redundant:
```python
def parabolic(x, y):
    """return the plane Poiseuille parabolic inlet profile"""
    return ((4 * y * (1. - y), np.zeros_like(y)))
```

PR #140, opened 2019-03-15
c.f. -> see

gdmcbain:
I don't think _c.f._ is right; see https://en.wiktionary.org/wiki/c.f. .


kinnala:
You're correct. I have no idea where this came from. I've been relying too much to the copy editing of journals in the past.

PR #145, opened 2019-03-21
steady Navier–Stokes by natural parameter continuation

gdmcbain:
This is extends the creeping flow over the backward-facing step from ex24 to finite Reynolds number using `pacopy.natural` parameter continuation in the Reynolds number.

This example is twice the length of the previous longest (ex04 and ex24) and still requires substantial rewriting and tweaking, but it's starting to take form and starting to work so I thought I'd share it early.

It also motivates the proposal to add &lsquo;milestones&rsquo; to nschloe/pacopy#5 (so that the solution can be obtained at specified values of Reynolds number).

Natural parameter continuation is preferred to arc-length continuation (`pacopy.euler_newton`) as used in ex23 and discussed in #119 because there's no attempt here to track bifurcation, so the solution is monotonic in the parameter.  Physically, the bifurcation involves three-dimensional effects; see, e.g.

* Barkley, D., Gomes, M. G. M. & Henderson, R. D. (2002). Three-dimensional instability in flow over a backward-facing step. _Journal of Fluid Mechanics,_ **473**, 167&ndash;190. [doi: 10.1017/s002211200200232x](https://dx.doi.org/10.1017/s002211200200232x)


kinnala:
An interesting way to structure the example. I guess it's related to pacopy and its interface? I'll read it more carefully later.

gdmcbain:
> An interesting way to structure the example. I guess it's related to pacopy and its interface?

Yes, for [pacopy](https://github.com/nschloe/pacopy)

> The only thing the user must provide is a class with some simple methods, e.g., a function evaluation `f(u, lmbda)`, a Jacobian a solver `jac_solver(u, lmbda, rhs)` etc.

I figured that since a class was needed anyway, I might as well pack the other data and methods in it too.  It can be judged at the end whether this is clearer or more cumbersome.

(While working on this, it also occurred to me that a similar structure might be useful for abstracting time-stepping, where the residual function `f` would also depend on the rate of change of `u` and the jacobians with respect to `u` and its rate of change would be the stiffness and mass matrices, respectively, but that's for later.)

gdmcbain:
Once this is done, quite a lot of code could and should be shared with the creeping version, ex24.

gdmcbain:
Ah, I had the indices around the wrong way on the vector-gradient!  Is the order of indices documented somewhere?  (I'm not sure that it shows up in the elastic examples because the tensors there are symmetric.) Anyway, after 1aaa7f0, I'm getting the right answers now, like Barkley, Gomes, & Henderson (2002).  Something like $u_j v_{i,j}$ would be written `u[j] * dv[i][j]`which makes sense, but I had been thinking more like $u_j \partial_j v_i$ which led me astray.

gdmcbain:
It doesn't take too long to run now (about a minute and a half, depending on fineness of mesh and parameters for Newton iteration and continuation, etc., the hardware), but here are some results.

ns-240 72-psi
ns-490 54-psi
ns-765 04-psi

The messy Reynolds numbers are an artefact of `pacopy.natural`, hopefully to be fixed by nschloe/pacopy#6.

kinnala:
Impressive work. I believe this starts to be ready for merge?

gdmcbain:
Thank you.  (I confess that I am quite excited about this one.)  

Before merging, I would like to sort out the steps in the Reynolds number.  They really should be at round values, as, e.g., by Gartling ([1990](https://doi.org/10.1002%2ffld.1650110704), §3):

> The backward step flow was formulated and solved as a steady flow problem.  In order to obtain a solution for the specified Re of 800, a series of steady state solutions were obtained at intermediate Re-values of 200, 400 and 600.  Zeroth-order continuation was used to advance from one Re solution to another.  The indicated sequence of steps in Re does not represent an optimized path to the final required state but was selected for convenience and with a view towards comparing solutions at other standard conditions.

The adoption of pacopy.natural might have been premature; its use here will probably involve lots of tweaks and that might be unreasonably disruptive upstream.

Locally,  there's also:
* refactoring with the creeping version in ex24 
   * (I haven't decided what the best way to do that is, or about what was perhaps charitably described as an ‘interesting way to structure the example’; the big `class` makes this example seem more different to those preceding than it really is.)
* organizing the sequence of viscous incompressible flow examples in the documentation (ex20, ex18, ex 24, this), as mentioned in #150 

but they could be deferred till after merging.

gdmcbain:
O. K., nschloe/pacopy#6 has been accepted so the natural parameter continuation hits a list of specified Reynolds numbers (150, 450, 750) and plots the stream-lines there.  I think this is ready to merge.

docs/examples/ex27.py
```diff
@@ -0,0 +1,237 @@
...
+from pygmsh.built_in import Geometry
+
+from pacopy import natural
+
+
+@linear_form
+def acceleration(v, dv, w):
+    """Compute the vector (v, u . grad u) for given velocity u."""
+    u, du = w.w, w.dw
+    # TODO: Handle the indices more cleverly
```
kinnala: Resolve TODO before merge.

gdmcbain: Yes.

I'm struggling a bit with this.  I'd like to write this so that it was independent of the number of spatial dimensions instead of only working for two, as here.  And then probably move the function to `skfem.models.general` or a new hydrodynamical submodule of `skfem.models`.

I've made a slight improvement with
```python
    # return (v[0] * (u[0] * du[0][0] + u[1] * du[0][1])
    #         + v[1] * (u[0] * du[1][0] + u[1] * du[1][1]))
    return sum(u * du[0]) * v[0] + sum(u * du[1]) * v[1]
 ```
but that's still two-dimensional.  Perhaps I should be using [`numpy.einsum`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.einsum.html) here.

Comment (gdmcbain) [reply docs/examples/ex27.py]: This is a bit [above](https://github.com/kinnala/scikit-fem/pull/145#issuecomment-475484311).  This time, I should work out properly all the indices on u, dv, w.w, and w.dw, how they relate to components of velocity, components of gradient, and quadrature point, and what shape is required in the output, and document all that.

Comment (kinnala) [reply docs/examples/ex27.py]: I'm just guessing here:
```python
return sum(np.einsum('ij...,j...', du, u), v)
```

kinnala: The reason why you need the ellipsis here is that the last dimension corresponds to the number of quadrature points and the second-to-last dimension is the number of elements. The shape of `u` is `Ndim x Nelems x Nqp` and the shape of `du` is `Ndim x Ndim x Nelems x Nqp`.

kinnala: Let me add that the output shape should be `Nelems x Nqp`.

gdmcbain: Thank you.  That looks good.  I'll study einsum more closely so that I actually know what I'm doing.  I see that it has two different modes, ‘explicit’ and ‘implicit’, depending on whether the indices of the result are specified; ‘explicit’, as in your [suggestion](https://github.com/kinnala/scikit-fem/pull/145#discussion_r279053772) looks clearer.  

I wonder whether it would also be able to subsume the `sum` with `v`, and then whether that would be more or less clear.  It might be more explicit, instead of having to remember or work out which index the `sum` was summing over.  I'll experiment.

gdmcbain: Oh, actually it's even easier than that since the elements and abscissæ are the trailing two indices for all three entities; thus, a trailing ellipsis for all three and no need to explicitly state the indices of the output:
```python
np.einsum('j...,ij...,i...', u, du, v)
```

gdmcbain: That seems reasonably close to the Cartesian tensorial notation ignoring the discretization: $u_j u_{i,j} v_i$.

gdmcbain: A single `einsum` can't be used for the bilinear form though, so it's either 
```python
    return sum((np.einsum('j...,ij...->i...', w.w, du)
                     + np.einsum('j...,ij...->i...', u, w.dw)) * v)
```
or
```python
np.einsum('j...,ij...,i...', w.w, du, v) + np.einsum('j...,ij...,i...', u, w.dw, v)
```
Not much between them, so I've gone with your `sum`–`einsum` suggestion for both linear and bilinear.  Thanks again.  

I think this is good to go now.  

I'll leave these forms in this example for now, moving them to `skfem.models` later if they're needed again in future examples.

gdmcbain:
In the course of developing this example, I learnt `numpy.einsum` and was impressed.  It occurred to me that it might be possible to use it to write a higher-level interface to the `bilinear_form` and `linear_form` decorators in `skfem.assembly.asm`, something like FEniCS's [UFL](https://fenics-ufl.readthedocs.io) or [`nutils.expression`](http://www.nutils.org/en/latest/nutils/expression/).

Issue #150, opened 2019-03-27
Lagrange multipliers etc for Dirichlet conditions


gdmcbain:
One of the techniques for imposing Dirichlet conditions in the finite element method is Lagrange multipliers [TODO: add references].  This has been mentioned in a few previous issues [TODO: ditto].

This issue is to add a fairly simple example (plane Laplace? like ex14?) demonstrating how to implement the technique in scikit-fem.

gdmcbain:
The discussion in #112 included the advice that

> There are at least two additional ways: Nitsche's method (see interior penalty example in docs, this I'd prefer over penalty) and using mixed formulation where Dirichlet condition is natural (i.e. mixed Poisson with H(div) elements).

with which I agree; however, I thought that it might be at least a worthwhile exercise in constructing the appropriate `FacetBasis` to see what penalization looks like in scikit-fem.  Unsurprisingly, it's easy and works well for the simple example of ex14. ee81411  Is it worth including for completeness of the discussion of boundary conditions?  I'm thinking of rewriting ex14.rst to explain the various techniques for essential conditions: three so far (`condense`, Lagrange multiplier ab6ea3aacc9e5b06fcead67f05e99bdff07fb414, and now penalization).   

TODO: Nitsche (which I haven't learned yet) and Raviart–Thomas (is that the intended mixed formulation?  I have implemented that previously in another code but found the performance abominable, presumably because I didn't have a good preconditioner, though I presume they are out there.)

For this, I was thinking that the examples in https://kinnala.github.io/scikit-fem-docs/ are a bit flat/linear/unstructured.  Would a tree make more sense than a list?  Also perhaps for the three existing examples on the Stokes equations and the pending one on Navier–Stokes #145.

kinnala:
> Is it worth including for completeness of the discussion of boundary conditions? I'm thinking of rewriting ex14.rst to explain the various techniques for essential conditions: three so far (condense, Lagrange multiplier ab6ea3a, and now penalization).

Extending ex14 is a good idea. Regarding penalty method, I'd like the example to be clear on its cons:
- Not consistent (exact solution does not satisfy the penalized weak form)
- Converges optimally only for low-order basis (if I remember correctly only p=1)
- Condition number increases asymptotically faster than for standard FEM discretizations. This makes it much harder to use iterative methods for the resulting linear system as their convergence rate depends on the condition number. This can be fixed by choosing penalty parameter cleverly but fixing it will further increase the discretization error.

[...]

PR #151, opened 2019-03-27
WIP: various techniques for a Dirichlet condition

gdmcbain:
This is for #150.

It reproduces ex14, viz. Δ _u_ = 0 in the unit square with _u_ = _x_ ² &minus; _y_ ² on the boundary.  Since _x_ ² &minus; _y_ ² is harmonic, that's the exact solution throughout.

### DIFF
--- a/docs/examples/ex14.py
+++ b/docs/examples/ex14.py
@@ -1,5 +1,5 @@
 from skfem import *
-from skfem.models.poisson import laplace
+from skfem.models.poisson import laplace, mass
 
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


gdmcbain:
This is working pretty well, though I'm hardly the resident expert on this topic!  So definitely leaving this marked WIP until reviewed.  I note that although convergence appears quadratic (in offline tests), this is not the exactness that was obtained by ex14 (P2 being able to represent the exact solution of degree two).  I haven't thought through whether that's expected from the reformulation.

gdmcbain:
On hold pending :
* discussion of relative merits
* demonstration of same
* addition of better alternatives

 See #150 for details.

gdmcbain:
Reopening with addition of Nitsche's method.

gdmcbain:
Putting this back on hold while I experiment more with the various approaches and read more about them.  I won't delete the lagrange-multiplier branch though as it still contains implementations of Lagrange multipliers, penalization, and Nitche's method for reference.  (I should rename it though.)

PR #152, opened 2019-03-29
new example: plane Graetz

gdmcbain:
After the steady Navier–Stokes example #145 is sorted out, I was planning on extending it to include forced convective heat transfer, as in Blackwell & Pepper's 1992 [_Benchmark Problems for Heat Transfer Codes_](https://www.worldcat.org/title/benchmark-problems-for-heat-transfer-codes-presented-at-the-winter-annual-meeting-of-the-american-society-of-mechanical-engineers-anaheim-california-november-8-13-1992/oclc/28344562)); however, before that, it probably makes sense to try advection–diffusion with a known velocity field, e.g. the Graetz problem, which is done here.

### DIFF
--- a/docs/examples/plane_graetz.py
+++ b/docs/examples/plane_graetz.py
@@ -1,0 +1,44 @@
+from skfem import *
+from skfem.models.poisson import laplace
+
+from math import ceil
+
+import numpy as np
+
+mesh_inlet_n = 2**5
+height = 1.
+length = 10.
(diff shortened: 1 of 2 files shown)


gdmcbain:
The figure, being the temperature, cold inlet at left, hot floor, symmetry top, outlet right.

plane_graetz

kinnala:
I'll add it to the docs.

PR #156, opened 2019-04-02
correct implementation of damped Newton iteration

gdmcbain:
While still lamenting the disappointing failure of `scipy.optimize.root` to handle sparse Jacobians in #119, I took another look at the hand-implementation of damped Newton iteration in ex10. I'm not sure it's correct?

https://github.com/kinnala/scikit-fem/blob/752dc7f47043f818c2c64c1d697c36881c7a3bd7/docs/examples/ex10.py#L35

I think it's more supposed to look like 
```python
x[I] -= 0.7 * solve(*condense(J, F, I=I))
```

as in equation 1.8b' of 

*  Seydel, R. (1988). _From Equilibrium to Chaos: Practical Bifurcation and Stability Analysis._ New York: Elsevier. ISBN: 0444012508

Compare the old and new figures, below; the new one looks more like what a soap-film would do.

ex10-old

_**Figure:—**_ Old

ex10-new

_**Figure:—**_ New

### DIFF
--- a/docs/examples/ex10.py
+++ b/docs/examples/ex10.py
@@ -26,17 +26,17 @@ I = m.interior_nodes()
 D = m.boundary_nodes()
 x[D] = np.sin(np.pi*m.p[0, D]) 
 
-relaxation = 0.7
 for itr in range(100):
+    import copy
     w = basis.interpolate(x)
     J = asm(jacobian, basis, w=w)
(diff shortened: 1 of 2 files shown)


kinnala:
Thanks for this. I remember wondering that why the solution looks so strange but didn't realize that the implementation was incorrect. I've so used to writing fixed-point iterations. :-)

PR #160, opened 2019-04-03
DOC: demonstrate how to solve a PDE in a subdomain

### LINKED ISSUE #159: Solve PDE just in a subdomain
Thinking about different ways to represent multiphysics problems like conjugate heat transfer, one of the basic questions is whether to use one or many meshes.   If one, it might be divided into subdomains (in the mesh, e.g. in Gmsh), as done previously for piecewise-definition of coefficients #89.  It could be that a single mesh is easier to work with than several, e.g. for identifying the interfaces and handling transmission conditions.

The issue then is how to solve a partial differential equation just in one subdomain.  Note that `InteriorBasis` always extends over the whole mesh, not having a restricting optional argument like `facets` of `FacetBasis`.  Thus a finite element function is always defined over the whole mesh; I don't think it's possible with the current machinery to define one just on a subdomain.

I think one way to restrict a problem to a subdomain might be via `Mesh.subdomains`, `InteriorBasis.element_dofs`, and `condense`.  I'll try to put an example together and see how it goes.

gdmcbain:
Fixes #159.

Based on ex17, but fixing the temperature in the 'insulator' and only solving in the 'wire'.

This turned out to be quite easy, given the existing machinery for subdomains and condensation.


gdmcbain:
I'll update this if `InteriorBasis` gets its optional `elements` argument #161.

PR #162, opened 2019-04-04
WIP: forced convection with velocity computed from Navier–Stokes

gdmcbain:
This example combines:
* forced convection 
* steady Navier–Stokes

going beyond the previous example of forced convection, plane Graetz #152, in that the velocity field is fully two dimensional and calculated by solving the steady Navier–Stokes equation on the same mesh, roughly as in #145.  The geometry is slightly different, having the inlet flush with the step.  It might be better to change the geometry in #145 from that of Barkley et al to match this. 

This problem is taken from:

* Dyne, B. R., & Heinrich, J. C. (1992). Flow over a backward-facing step: a benchmark problem for laminar flow with heat transfer. In B. F. Blackwell & D. W. Pepper (eds).  _Benchmark problems for heat transfer codes_ : presented at the Winter Annual Meeting of the American Society of Mechanical Engineers, Anaheim, California, November 8-13, 1992.

Like #145, it relies on nschloe/pacopy#6 being merged for natural parameter continuation to get to Re = 800; currently I'm using my branch.

With the currently chosen numerical parameters (mesh refinement, continuation step-size, tolerance, &c.), it takes about five minutes to run (as well as the special branch of pacopy), so I'll attach the resulting PNG here.

Basically warm air is blown in through the top half of the left wall and subject to uniform cooling on the floor and ceiling.

blackwell_pepper_solution

### DIFF
--- a/docs/examples/blackwell_pepper.py
+++ b/docs/examples/blackwell_pepper.py
@@ -234,7 +234,7 @@ def callback(k, reynolds, uvp):
         velocity[reynolds] = bfs.split(uvp)[0]
         
 
-natural(bfs, bfs.make_vector(), 2e2, callback,
+natural(bfs, bfs.make_vector(), 1e2, callback,
         lambda_stepsize0=200.,
         lambda_stepsize_max=200.,
         max_newton_steps=2**3,
--- a/docs/examples/blackwell_pepper.py
+++ b/docs/examples/blackwell_pepper.py
@@ -1,0 +1,304 @@
+"""
+
+This redoes the backward-facing step, but as in Gartling (1990), or rather Dyne & Heinrich (1992)
+
+*  Dyne, B. R., & Heinrich, J. C. (1992). Flow over a backward-facing step: a benchmark problem for laminar flow with heat transfer. In B. F. Blackwell & D. W. Pepper (ed.), *Benchmark Problems for Heat Transfer Codes,* Vol. 222 (pp. 73–76). ASME.
+
+* Gartling, D. K. (1990). A test problem for outflow boundary conditions—flow over a backward‐facing step. *International Journal for Numerical Methods in Fluids,* 11, 953–967. doi: 10.1002/fld.1650110704
+
+"""
+
+from skfem import *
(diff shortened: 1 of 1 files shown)


kinnala:
Really impressive work again. I'll ready through the code and see if I can figure out any optimizations.

kinnala:
Nothing obvious yet. Maybe precalculating Reverse Cuthill-McKee reordering for the rows and columns of self.S would help a bit but can't say for sure. Another idea is to save LU-decomposition of self.S and use GMRES with the initial guess from solve(L, solve(U, rhs)) - I assume rhs changes each time. That would require incomplete LU preconditioning based on self.S + block_diag(...).

These are just some ideas but if you are fine with the 5 minute execution time, no need to do any actions.

gdmcbain:
Thank you for these suggestions.  The five minutes is almost all in the Navier&ndash;Stokes.  The reason this example takes longer than #145 is that the Reynolds number is effectively doubled.  It's nominally 800 in both cases but the length-scale was the height of the step for Barkley, Gomes, & Henderson (2002) but of the downstream channel for Gartling (1990).

The forced convection which is basically by comparison a cheap postprocessing operation.

Actually although this example is a benchmark from the literature, it might be better here to demonstrate forced convection on the Stokes example of the [backward-facing step](https://kinnala.github.io/scikit-fem-docs/examples/ex24.html) and note the implementation for Navier&ndash;Stokes would be identical.  This would make it reasonable to just import the `ex24.velocity`, ex24 running in under three seconds.

Five minutes is O. K. but it should be added that the method here and in the Navier&ndash;Stokes example #145 does not scale well to larger problems.  I've launched issues for:
* reordering #168 and
* iterative linear solvers #169.

gdmcbain:
I'm thinking of omitting this example.  It doesn't really add too much, just straightforwardly combining `ex25` ‘Forced convection’ and the Newton steady Navier–Stokes #145, but it takes a while to run.  Maybe just leave the branch and close this pull-request?  I might add a natural convection example later, like a nonzero Grashof number extension of `ex18` ‘Stokes equations’.

PR #166, opened 2019-04-07
Subdomain assembly

### LINKED ISSUE #161: optional subset of elements in InteriorBasis
It is already possible to solve a PDE in a subdomain #159 but:

> I'm surprised that InteriorBasis.__init__ does not have an optional argument for specifying the set of elements.

This would be consistent with the optional facets argument of FacetBasis.

kinnala:
Fixes #161. Leaving it here for any comments, @gdmcbain.

### DIFF
--- a/docs/examples/ex26.py
+++ b/docs/examples/ex26.py
@@ -6,12 +6,13 @@ import numpy as np
 from ex17 import mesh, basis, radii, joule_heating, thermal_conductivity
 
 
-L = asm(laplace, basis)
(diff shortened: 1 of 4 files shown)


gdmcbain:
I've proposed the corresponding changes to ex26.py in #167.  It seems fine.

I haven't timed `asm` to see whether it's faster or measured the size of the CSR stiffness matrix in memory to see whether it's smaller.

I read through the changes to `InteriorBasis` and they seemed reasonable.

Issue #13, opened 2018-06-27
multicolumn right-hand sides


gdmcbain:
This is a nonissue in a sense; multicolumn right-hand sides just work with no modification required to the usual idiom of e.g. ex14 36d7ace10dcff53ecdb9e500f09dad4ebbba4fc1
```python
u[dofs] = solve(*condense(A, 0.*b, u, dofs))
```
where dofs is still a one-dimensional array of indices but now `u` and `b` are two-dimensional with `u.shape[0] == b.shape[0] == A.shape[0]` and `u.shape[1] == b.shape[1] >= 1`.

However, as that might not be obvious and as it is so useful in applications, this is being noted here as worth documenting and worth demonstrating in an example.  (I hope to provide one shortly; I have it working in complicated cases but would like to strip it down to essentials before publishing.)

kinnala:
I'm curious that what kind of applications you have in mind?

gdmcbain:
Yes, sorry, a month has passed without my having come up with a nice potted example here.

[...]

PR #171, opened 2019-05-03
demonstrate a conjugate gradient Uzawa algorithm for the Stokes equation

### LINKED ISSUE #170: DOC: Uzawa algorithm for Stokes equation
As a step towards the iterative solution of the Navier–Stokes equation #169, an Uzawa algorithm could be implemented for the Stokes equation, redoing say the creeping convection example ex18.  This should be fairly straightforward and enable solution on much finer meshes and in three dimensions.  I don't think it should require any extension or modification of the `skfem` package.

The algorithm, which applies to any saddle-point problem, is described by:

* Braess, D. (2001). _Finite Elements_. Cambridge University Press, §IV.5.1
* Glowinski, R. & Pironneau, O. (1979). On numerical methods for the Stokes problem. In R. Glowinski, E. Y. Rodin & O. C. Zienkiewicz (eds.), _Energy Methods in Finite Element Analysis_ (pp. 243–264), Wiley

gdmcbain:
Fixes #170.


gdmcbain:
Some preliminary timings in comparison with the direct solver of ex18, using _h_ as the second argument to `dmsh.generate` and
```python
tic = time()
```
before the first `asm` and
```python
print('elapsed time:', time() - tic)
```
after the `split` of the velocity and pressure.

| _h_ | ex18 / s | CG Uzawa / s |
| ----- | -------- | ----------------| 
| .1 | 0 | 0 |
| .1/2 | 1 | 1 |
| .1/4 | 12 | 1 |
| .1/8 | 128 | 7 |

Promising but unfortunately `dmsh` struggles with _h_ = 1./2⁴.  I might try the same problem on a square.  Not for the example suite, just to show how the time scales.

gdmcbain:
Some preliminary timings in comparison with the direct solver of ex18, using _h_ as the second argument to `dmsh.generate` and
```python
tic = time()
```
before the first `asm` and
```python
print('elapsed time:', time() - tic)
```
after the `split` of the velocity and pressure.

| _h_ | ex18 / s | CG Uzawa / s |
| ----- | -------- | ----------------| 
| .1 | 0 | 0 |
| .1/2 | 1 | 1 |
| .1/4 | 12 | 1 |
| .1/8 | 128 | 7 |

Promising but unfortunately `dmsh` struggles with _h_ = 1./2⁴.  I might try the same problem on a square.  Not for the example suite, just to show how the time scales.

gdmcbain:
```python
mesh = MeshTri.init_tensor(*(np.linspace(0., 1., 2**k),)*2)
```

| _k_ | ex18 / s | CG Uzawa / s |
|-|-|-|
| 5 | 1 | 0 |
| 6 | 7 | 1 |
| 7 | 71 | 6 |
| 8 | — | 26 |
| 9 | — | 169 |

The direct solver exhausted the memory for _k_ ≥ 8, the CG Uzawa for _k_ ≥ 10.

kinnala:
This looks ready. Should we merge with stub documentation or are you willing to improve that part?

gdmcbain:
While I'm very pleased with the progress thus far, I'd like to experiment with this a bit more.

* Maybe replace the direct solver for the viscous vector-Laplacian subproblem with something iterative.
   * Perhaps as in ex09.
   * Maybe split the Cartesian components of this subproblem—they're uncoupled, the matrix is block-diagonal.  Indeed the two blocks are the same, just the scalar laplacian with the same pattern of Dirichlet degrees of freedom.  They can be solved as multicolumn right-hand sides #13, or in parallel.  In another code, this roughly halved the overall run-time for three-dimensional Stokes problems.
   * Maybe introduce algebraic multigrid preconditioning.  Preliminary experiments with pyamgcl are promising.  On the other hand, perhaps AMG is worth a new example all its own, e.g. extending ex09.
* Also I'll scrap this branch and create a cleaner one with less digressions.
* Maybe demonstrate the algorithm with a three-dimensional example; the exact solution for the circular cavity in exx 18, 20 is just a special case of that for an ellipsoid ([hal-02013383](https://hal.archives-ouvertes.fr/hal-02013383)).
* I think the conjugate gradient iteration for the pressure should be preconditioned, maybe with the pressure's mass matrix.

I'll definitely write some documentation, once an approach has been settled on, explaining what has been implemented and why, though I'm conscious of not being an expert on iterative methods or preconditioners.

gdmcbain:
> Maybe introduce algebraic multigrid preconditioning.

I resumed work on this after successful use of AMGCL in #273, starting with
```python
Aint = condense(A, D=D, expand=False)
solver = solver_iter_pcg(M=amgcl(Aint))
I = basis['u'].complement_dofs(D)


def flow(pressure: np.ndarray) -> np.ndarray:
    """compute the velocity corresponding to a guessed pressure"""
    velocity[I] = solve(Aint,
                        condense(csr_matrix(A.shape),
                                 f + B.T @ pressure, I=I)[1],
                        solver=solver)
    return velocity
```
but hit an unexpected snag:
```
scipy._lib._threadsafety.ReentrancyError: cg is not re-entrant
```
I do want to use nested conjugate gradients here: 
* inside for the momentum equation (as the accelerator for AMG), 
* outside for Uzawa.  

Perhaps I'll need another implementation of `cg`.

gdmcbain:
`pyamgcl.solver` seems to work O. K. in place of `sksparse.cholesky` in the momentum Poisson equation, avoiding the reeentrancy.

kinnala:
I suppose this is ready for a review?

gdmcbain:
Yes.

I might tweak this later, following the development of an example for the unsteady Navier&ndash;Stokes equations #240, so that the suite of examples are more coherent, but I don't think it will be much different from this.

The present version of this doesn't use AMG since CG+ILU works just as well at this size and that doesn't depend on any external packages beyond SciPy.
 
I took care to choose the size here so that the example runs quickly (about five seconds on a humble machine here), but is big enough to start to demonstrate the need for a preconditioned iterative method.

gdmcbain:
The examples, or their presentation in the documentation, should be restructured soon to put the viscous incompressible fluid mechanics in a separate section.  They're overweighing the rest and also form a pedagogical sequence within themselves.  I had a quick look at Restructured Text and Sphinx but didn't immediately see how to do it.

But I don't think this PR needs to wait for that.

docs/examples/ex30.rst
```diff
@@ -0,0 +1,49 @@
+.. _uzawa:
+
+Krylov–Uzawa method for the Stokes equation
+-------------------------------------------
+
+This example solves :ref:`stokesex` iteratively in a square domain.
+
+Direct solvers for the Stokes equation do not scale; iterative solvers are required for all but the smallest two-dimensional problems.  A square domain is used here instead of the disk since it easier to generate the finer mesh required to demonstrate the time taken for a slightly larger problem.
+
+A classical iterative procedure for the Stokes equation is the Uzawa congugate gradient method in which an auxiliary operator is formed which maps a guessed pressure field :math:`p` to the divergence
```
kinnala: congugate -> conjugate

docs/examples/ex30.rst
```diff
@@ -0,0 +1,49 @@
...
+   K (p) = \Theta (p) - \Theta (0)
+   
+for which the linear operator equation
+
+.. math::
+   K (p) = -\Theta (0)
+   
+can be solved by a Krylov method, classically the method of conjugate gradients.  The solution is the pressure :math:`p` that gives the velocity :math:`\boldsymbol u` that has zero divergence.
+
+At each iteration, the above vector Poisson equation is solved for a velocity field.  As the problem gets larger, this too should be solved iteratively.  The method of conjugate gradients could be used for this too, but a catch is that :func:`scipy.sparse.linalg.cg` is not re-entrant can so cannot be used at both levels.  This is simply avoided here by using :func:`scipy.sparse.linalg.minres` for the pressure.
```
kinnala: can so cannot -> and so cannot

docs/examples/ex18_square.py
```diff
@@ -0,0 +1,84 @@
...
+D = basis['u'].get_dofs().all()
+uvp = np.zeros(K.shape[0])
+uvp = solve(*condense(K, f, D=D))
+
+velocity, pressure = np.split(uvp, [A.shape[0]])
+print('elapsed time:', time() - tic)
+
+
+@linear_form
+def rot(v, dv, w):
```
kinnala: This is already imported and defined in skfem.models.general.

docs/examples/ex18_square.py
```diff
@@ -0,0 +1,84 @@
+from skfem import *
```
kinnala: Should this file replace ex18.py or be separate?

kinnala: Add to the test suite?

gdmcbain: Sorry, ex18_square was just a preliminary sketch which I thought I had deleted.

docs/examples/ex30.rst
gdmcbain: Ta.

docs/examples/ex30.rst
gdmcbain: Ta.

docs/examples/ex18_square.py
gdmcbain: Obsolete script, subsequently removed.

Issue #175, opened 2019-05-08
MeshLine.load submeshes & boundaries


gdmcbain:
`MeshLine.load` warns

> UserWarning: Unable to load submeshes

for any input, and fails to read any subdomains and boundaries present.

The immediate cause is that `'line'` is absent from the keys in

https://github.com/kinnala/scikit-fem/blob/dadd4bac13ca66786c6a545fbb3f139c8b34836a/skfem/mesh/mesh.py#L286-L291

While the one-dimensional examples to date just use `MeshLine.__init__`, and this suffices for most purposes, it is occasionally useful to generate even one-dimensional meshes externally in Gmsh; e.g. for problems like the ‘composite wall composed of _n_ slabs of thicknesses _l1, …, ln_ and conductivities _K1, …, Kn_’ of Carslaw & Jaeger (1959, _Conduction of Heat in Solids_, §3.2 ‘Steady temperature’, Oxford University Press).

Minimal example: [line.geo.gz](https://github.com/kinnala/scikit-fem/files/3155517/line.geo.gz).

```
Point(1) = {0, 0, 0};
Point(2) = {1, 0, 0};
Line(1) = {1, 2};
```

Generate mesh with
```shell
gmsh -1 line.geo
```
and attempt to load with
```python
MeshLine.load('line.msh')
```

gdmcbain:
Just adding `'line': 'vertex'` seems to be enough to get `MeshLine.load` to not warn and to successfully read the subdomain defined by the addition line in `line.geo`
```
Physical Curve("domain") = {1};
```

gdmcbain:
`MeshLine.load` is not loading the boundaries correctly; e.g. for

```
Point(1) = {0, 0, 0};
Point(2) = {1, 0, 0};
Line(1) = {1, 2};
Physical Point("left") = {1};
Physical Point("right") = {2};
Physical Curve("domain") = {1};
```

which produces [line.msh.gz](https://github.com/kinnala/scikit-fem/files/3160026/line.msh.gz), containing
```
$PhysicalNames
3
0 1 "left"
0 2 "right"
1 3 "domain"
$EndPhysicalNames
```

gdmcbain:
I'm wondering whether the problem follows from `MeshLine._build_mappings` sorting `.p` whereas that method of other descendents, e.g. `MeshTri`, `MeshTet`, doesn't.

gdmcbain:
Yes, this is the problem.  In the above example, at

[...]

PR #176, opened 2019-05-08
enable MeshLine.load

gdmcbain:
For #175, which in a minimal sense is just because `'line'` is missing from the `dict` in `bnd_types` in 'Mesh._parse_submeshes`, but I'll check whether anything else is required.

### DIFF
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -284,6 +284,7 @@ class Mesh():
 
         # element to boundary element type mapping
         bnd_type = {
+            'line': 'vertex',
             'triangle' : 'line',
             'quad' : 'line',
             'tetra' : 'triangle',
--- a/skfem/mesh/mesh_line.py
+++ b/skfem/mesh/mesh_line.py
@@ -100,19 +100,18 @@ class MeshLine(Mesh):
         p = self.p
 
         # new vertices and elements
-        midpoints = p[:, self.t].mean(axis=1)
-        newp = np.hstack((p, midpoints))
-        # update fields
-        self.p = newp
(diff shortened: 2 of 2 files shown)


gdmcbain:
This is enough to suppress the warning and load Physical Curves as `MeshLine.subdomains`, if present.

Fixes #175.

gdmcbain:
Shortcoming: see #178; basically `MeshLine.p.shape[0]` and therefore `MeshLine.dim()` is whatever was in the file, typically 3.

gdmcbain:
Fixes #178; the problem was that `MeshLine.strip_extra_coordinates` wasn't overridden and defaulted to that from `Mesh` which does nothing.

gdmcbain:
Shortcoming: doesn't read all boundaries #175.

gdmcbain:
This has been rewritten so that `MeshLine._build_mappings` no longer sorts `MeshLine.p`.  This required fairly extensive changes, e.g. to the
*  `__init__` method, which previously silently ignored `t`
*  `_uniform_refine`, which relied upon a subsequent call to `_build_mappings` to resort `p` and regenerate `t` from scratch

It now all works though, successfully running ex15, ex16, ex23, the various one-dimensional `tests`, and an in-house example involving multiple Gmsh `Physical Points` which become entries of `MeshLine.boundaries`.

kinnala:
Very impressive work. Thanks a lot for these contributions.

kinnala:
It's slightly inconvenient that building facets is now done in two places (refine and init) instead of one (_build_mappings), but we can try to workaround it later. I'm going to merge this.

Issue #174, opened 2019-05-08
DOC: plot stream-lines at milestone Reynolds numbers 


gdmcbain:
(This was part of #172, accidentally automatically closed with #173.)

> The plots should be at the nice round Reynolds numbers 150, 450, and 750, courtesy of the milestones in `pacopy.natural`, but they seem to be the old ones.

It refers to https://kinnala.github.io/scikit-fem-docs/examples/ex27.html.

Should `docs/Makefile` be modified to regenerate those figures required in `docs/examples/*.rst`?

kinnala:
I think running the examples during each Sphinx compile is not efficient.
Maybe we can add another option to Makefile which does just that. It'd be
nice to have such feature only for the missing images but maybe its hard to
get reverse mapping from png to py.

gdmcbain:
Yes, no, one wouldn't want to have to run the examples every time, only if the script had changed.

What about a rule for each required image, this rule being added explicitly to the Makefile?  Then all the images listed in a Make variable and the Sphinx job to depend on that image Make variable.   I'm not exactly sure how Sphinx works, but I've used Make-rules like that for papers.  I'll dig one out and sketch the idea.

gdmcbain:
Here's a snippet of an old Makefile:
```Make
FIGURES=geometry.eps bryan.eps ellipse.eps
%.eps: %.py
	python ${PYFLAGS} $<
%.eps: %.asy
	asy $<
%.pdf: %.dvi ${FIGURES}
	dvipdf $<
```

Here the last rule would be replaced by something appropriate to Sphinx and instead of the pattern-based rules for Asymptote and Python I expect it would be easier and clear just to have an explicit rule for each set of images, like
```python
ex27-150.0-psi.png ex27-450.0-psi.png ex27-750.0-psi.png: ex27.py
    python $<
```

gdmcbain:
…except that the examples are in a subdirectory, so create a new Makefile in `docs/examples` with the explicit rules for the images and then call that recursively from `docs/Makefile`.

Then maybe we don't need a variable like `${FIGURES}` above and can just have a `.PHONY` rule `all` in `docs/example/Makefile` that depends on all the figures.

gdmcbain:
Ugh, no, this isn't right:

```Make
ex27-150.0-psi.png ex27-450.0-psi.png ex27-750.0-psi.png: ex27.py
    python $<
```

That runs ex27 three times…

gdmcbain:
[...]

PR #177, opened 2019-05-08
Make figures for Sphinx

gdmcbain:
This would fi× #174 but I suspect a manual step is also required somewhere.


kinnala:
No problem, this is much better than the current workflow of manually running each and every example.

PR #184, opened 2019-05-14
set test_suite to tests to allow python setup.py test

### LINKED ISSUE #183: python setup.py test fails
Forgetting the command line to run the test suite, I tried
```python
python setup.py test
```
and it raised
> TypeError: asm() missing 2 required positional arguments: 'kernel' and 'ubasis'

The command line is `python -m unittest discover`, as given in `.travis.yml` and `Makefile`, but I wonder what's required to get the other to work.

gdmcbain:
Fixes #183.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,4 +22,5 @@ setup(
     ],
     packages=find_packages(exclude=['tests']),
     install_requires=['numpy', 'scipy', 'matplotlib', 'meshio'],
+    test_suite='tests',
 )


gdmcbain:
This was the [answer](https://stackoverflow.com/a/21726329/6238185) to [‘How to run unittest discover from “python setup.py test”?’](https://stackoverflow.com/questions/17001010/how-to-run-unittest-discover-from-python-setup-py-test).

PR #186, opened 2019-05-16
solver_iter_krylov

### LINKED ISSUE #185: generalize utils.solver_iter_pcg to other Krylov methods
This was touched on as part 67f3aba of the iterative linear solution for the Newton decrement of the steady Navier–Stokes equation #169, but Krylov subspace methods other than conjugate gradient are also required for simpler problems, e.g. anything not symmetric definite, such as the advection–diffusion equation of ex25.  ILU-preconditioned GMRES or Bi-CGSTAB should help for that (Elman, Silvester, & Wathen 2014, _Finite Elements and Fast Iterative Solvers_, §7.1 ‘Solution of discrete convection–diffusion problems/Krylov subspace methods’).

The call signatures of [`scipy.sparse.linalg.gmres`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.gmres.html) and [`bicgstab`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.bicgstab.html) are very similar to that of `cg`, so they should be able to share most of the code in `skfem.utils.solver_iter_pcg`, as in 67f3aba; moreover, I believe `build_pc_ilu` works without change.

Other Krylov solvers such as [`minres`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.minres.html) might come in handy too for other problems (ibid, §§2.4, 4.1).

gdmcbain:
This fixes #185, minimally generalizing `skfem.utils.solver_iter_cg` enabling use of, e.g. `gmres` and `bicgstab`.


gdmcbain:
Some preliminary results for motivation, based on ex25, increasing the refinement-level exponent from 5 to 6, 7, & 8 in

https://github.com/kinnala/scikit-fem/blob/625ac42ed2185982752c49def5259c6f4e6fa4a2/docs/examples/ex25.py#L8

and replacing the solver with 

```python
solver = None if krylov is None else solver_iter_krylov(
            krylov, pc=pc, verbose=True)
t[self.interior] = solve(self.A1, self.b1, solver=solver)
```
where `pc` is the result of applying `build_pc_ilu` to the condensed matrix.

refinement| `basis.N`|      direct|        ILU|   Bi-CGSTAB|     GMRES
----------|--|------------|-----------|-----------|-----------
5|     40257|    0.621676|   0.726331|   0.052581|   0.087286
6|    162433|    5.751564|   4.802587|   0.316622|   0.559407
7|    652545|   53.741143|  29.543612|   3.174956|   4.043258
8|   2615809|         —| 194.140337|  39.615441|  38.589795

The direct solver killed the machine at refinement level 8.

I'm sure that there's much room for improvement here but even this first obvious step into iterative methods using just default settings already allows the solution of much bigger problems than the direct solver.  They're also faster for all but the smallest problem.

Not much to decide between GMRES and Bi-CGSTAB thus far.

PR #189, opened 2019-05-23
simplify skfem.utils.solver_iter_krylov by passing on kwargs

### LINKED ISSUE #188: tolerance for Krylov solvers
Experimenting with the new `skfem.utils.solver_iter_krylov` using `scipy.sparse.linalg.gmres` with `skfem.utils.build_pc_ilu` for a three-dimensional version of ex25 ('Forced convection'), it's going quite well (up to a `basis.N` of 911&thinsp;752 so far, solving in a few minutes).

But I found that the optional keyword argument `tol` to `solver_iter_krylov` didn't seem to be doing anything.  It turns out that this `tol` is passed in to the solver (here `gmres`) as `atol` and the `gmres` (relative) `tol` left as the default and this was dominating.

https://github.com/scipy/scipy/blob/0f5d8aab5d9990c433168ad3424b3f5c9440978e/scipy/sparse/linalg/isolve/lgmres.py#L168

Perhaps it should be possible to pass in `tol` as well as `atol`?

Maybe `solver_iter_krylov` should just pass on `*args, **kwargs` to the solver?  That would be simpler, if a little less explicit.  

It might also facilitate using solvers from outside `scipy.sparse.linalg` that take different arguments (presumably though always taking the linear operator and right-hand side as first two positional arguments, and maybe the initial guess as the third?).

gdmcbain:
Fixes #188 by enabling the user to pass on whatever they like to the underlying solver.

### DIFF
--- a/docs/examples/ex09.py
+++ b/docs/examples/ex09.py
@@ -21,10 +21,10 @@ if __name__ == "__main__":
 else:
     verbose = False
 # run conjugate gradient with the default preconditioner
-x[I] = solve(*condense(A, b, I=I), solver=solver_iter_pcg(verbose=verbose))
+Aint, bint = condense(A, b, I=I)
+x[I] = solve(Aint, bint, solver=solver_iter_pcg(verbose=verbose))
 
 # run conjugate gradient with the incomplete LU preconditioner
-Aint, bint = condense(A, b, I=I)
 x[I] = solve(Aint, bint,
              solver=solver_iter_pcg(M=build_pc_ilu(Aint), verbose=verbose))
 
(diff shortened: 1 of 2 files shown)


gdmcbain:
The calls to `solver_iter_pcg` in ex09 are no longer exactly the same in intent since the default `tol` and `maxiter` have been removed, falling back on the defaults of `scipy.sparse.linalg.cg`, but this does not change the behaviour, taking 22 and 4 iterations, respectively, with diagonal and ILU preconditioners, regardless of the implicit change in tolerance.

Although it would be easy enough to preserve the intent by appended `atol=1e-8, maxiter=100` to the call to `solver_iter_pcg`, I didn't bother.

Issue #118, opened 2019-01-30
SymPy for preprocessing examples


gdmcbain:
While dependence on SymPy was dropped in #37 for computing basis functions, it might be useful for preprocessing certain examples.  It would not then be a dependence for the scikit-fem package, just for the examples.

One is suggested by the FEniCS tutorial example [ft05_nonlinear_poisson.py](https://fenicsproject.org/pub/tutorial/python/vol1/ft05_poisson_nonlinear.py). The example is basically like `docs/examples/ex10.py`, except that it adds an artificial right-hand side to balance the equation for a specified solution; i.e. it uses the method of manufactured solutions.  That right-hand side is computed from the stipulated exact solution using SymPy.

A translation of the FEniCS example into scikit-fem is at https://github.com/gdmcbain/fenics-tuto-in-skfem/tree/master/05_poisson_nonlinear.  It goes beyond the original in using:
* [sympy.vector](https://docs.sympy.org/latest/modules/vector) for gradient and divergence
* [sympy.utilities.lambdify](https://docs.sympy.org/latest/modules/utilities/lambdify.html) to evaluate symbolic vector fields on points specified by a `numpy.ndarray` like `Mesh.p`

There are a few examples (ex12, ex13, ex14, ex16, ex17, ex18, ex19, ex20) which involve exact solutions; I wonder whether it would be worth modifying one or more of them to use SymPy to compute the exact solution to show how it's done here.

gdmcbain:
According to Cellier & Ruyer-Quil (2019), [Scikit-FDiff](https://gitlab.com/celliern/scikit-fdiff) uses SymPy to allow ‘easy and automated finite difference discretization’; might be worth a look for inspiration or comparison.

* Cellier, N. & Ruyer-Quil, C. (2019). scikit-finite-diff, a new tool for PDE solving. _Journal of Open Source Software,_ 4, 1356. [doi:10.21105/joss.01356](https://doi.org/10.21105/joss.01356)

PR #192, opened 2019-05-30
Add possibility to return elemental form of the system matrix

kinnala:
This feature was requested in Gitter for solving topology optimization problems. I could also see this to be useful if someone wanted to solve the resulting systems iteratively on GPU's. Any further ideas or objections @gdmcbain ?

### DIFF
--- a/skfem/assembly/asm.py
+++ b/skfem/assembly/asm.py
@@ -14,7 +14,8 @@ def asm(kernel,
         ubasis: GlobalBasis,
         vbasis: Optional[GlobalBasis] = None,
         w: Optional[Any] = (None, None, None),
-        nthreads: Optional[int] = 1) -> csr_matrix:
+        nthreads: int = 1,
+        return_elemental: bool = False) -> csr_matrix:
     """Assemble finite element matrices and vectors.
 
     Parameters
@@ -34,6 +35,8 @@ def asm(kernel,
         Number of threads to use in assembly. Due to Python global interpreter
         lock (GIL), this is only useful if kernel is numba function compiled
         with nogil = True, see Examples.
+    return_elemental
+        If True, return the elemental system matrices.
 
     Examples
     --------
@@ -105,10 +108,13 @@ def asm(kernel,
         for t in threads:
             t.join()
 
-        K = coo_matrix((np.transpose(data, (1, 0, 2)).flatten('C'), (rows, cols)),
-                        shape=(vbasis.N, ubasis.N))
-        K.eliminate_zeros()
-        return K.tocsr()
+        if return_elemental:
+            return np.transpose(data, (1, 0, 2)).flatten('C'), rows, cols
+        else:
+            K = coo_matrix((np.transpose(data, (1, 0, 2)).flatten('C'), (rows, cols)),
+                           shape=(vbasis.N, ubasis.N))
+            K.eliminate_zeros()
+            return K.tocsr()
 
     elif nargs == 5:
         data = np.zeros((vbasis.Nbfun, nt))


gdmcbain:
The `assembly.asm` function might be getting a bit (cyclomatically) complicated; perhaps it could be refactored, e.g. as `Mesh.load` was in #47, here having most of the work in a new function to compute the elemental matrices, leaving the `asm` function as a wrapper.  I'll have a go at that for comparison.

gdmcbain:
https://github.com/kinnala/scikit-fem/blob/master/skfem/assembly/asm.py#L13-L17

The hinted return type `csr_matrix` is only true for the bilinear (`nargs == 6`) case, not linear form (`nargs == 5`) or catch-all, which I assume is for functionals (`nargs == 2`); for the latter two cases `asm` returns `numpy.ndarray`.  Might be better to not hint the return type?  Or use `Union[csr_matrix, ndarray]`?

Another (more disruptive and therefore quickly discarded) idea I had a while ago was to replace the `@bilinear_form`, `@linear_form` and `@functional` decorators with classes with appropriate `asm` methods.  (I think I thought of that on first seeing `inspect.signature`, which seemed eldritch.)

gdmcbain:
Was the `@functional` `asm` also already like `return_elemental`?  In that ultimately `@bilinear_form`, `@linear_form`, and `@functional` should be assembled as a linear operator, vector, and scalar, respectively, but for `@functional`, `@asm` returns a vector (of length `basis.nelems` rather than `basis.N`).

gdmcbain:
Sorry, I'm not sure whether the above musings are helpful; anyway they're not objections, just (half-formed) ideas.

gdmcbain:
> Another (more disruptive and therefore quickly discarded) idea I had a while ago was to have the `@bilinear_form`, `@linear_form` and `@functional` decorators […] return objects rather than functions (‘closures are a poor person's objects’) with appropriate `asm` methods. 

Puzzled by the recent exchanges on gitter about the elemental forms, I was thinking about this again, so I did a quick sketch to demonstrate that a decorator can indeed return a callable object instead of a function:
```python
class Klass:

    def __init__(self, f):
        self.f = f
        
    def __call__(self, *args, **kwargs):
        return self.f(*args, **kwargs)

def klassy(f: Callable):
    return Klass(f)

@klassy
def double(x: int) -> int:
    return 2 * x
```

This approach (but with the `__call__` method not just calling `self.f` but subsuming the current `bilinear_form`, `linear_form`, and `functional` decorators) would enable features (like generating elemental subassemblies) to be added incrementally to forms as instance methods.

Other benefits might include: 
* Whereas the current `asm` function has several distinct branches because it has to do different things for different types of `kernel`, these would be separated into the `asm` methods of `class BilinearForm`, `class LinearForm`, and `class Functional` (which might be derived from some abstract superclass Form, if there were any shared behaviour).
* The selection between those branches via `nargs = len(signature(kernel).parameters)` can be avoided; each object just calls its class's own `asm` method automatically.
* The function that is passed to the decorator is preserved as an attribute so other things can be done with it, e.g. for debugging; it's currently impossible to call the function since its name is usurped by the decorator.  With the above, it would be available as `double.f`.
* It might be possible to separate the threading code, which again might ease debugging.
* Vague future ideas that this might facilitate: automatic differentiation (inspired by http://conal.net/papers/compiling-to-categories and last year's discussion of ASTs), symbolic manipulation #118 , …

I think this would be backwards-compatible, as in the toy example above, so it needn't necessarily impede 1.0.0 #194.

If it's thought that this might be worth looking into to, I think I could put it together in a new branch.

PR #197, opened 2019-06-19
reimplement boundary_edges using isin

### LINKED ISSUE #196: anomalous broadcast equality for large ndarray and Mesh3d.boundary_edges
I'm running some larger three-dimensional (tens of thousands to millions of nodes) steady convection–diffusion problems (like [ex25](https://kinnala.github.io/scikit-fem-docs/examples/ex25.html) but on bigger three-dimensional meshes from Gmsh) and noticed in profiling what seemed like an anomalously long time being taken by `basis.get_dofs(mesh.boundaries)` on the _smaller_ problems (10⁴ but not 10⁶).  I'm still digging into that, but on the way I turned up something scary:
```shell
$ python
Python 3.7.3 (default, Mar 27 2019, 22:11:17) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from numpy import empty
>>> empty(2**22, dtype=int) == empty((2**10, 1), dtype=int)
array([[ True,  True,  True, ...,  True,  True,  True],
       [ True,  True,  True, ...,  True,  True,  True],
       [ True,  True,  True, ...,  True,  True,  True],
       ...,
       [ True,  True,  True, ...,  True,  True,  True],
       [ True,  True,  True, ...,  True,  True,  True],
       [ True,  True,  True, ...,  True,  True,  True]])
>>> empty(2**22, dtype=int) == empty((2**11, 1), dtype=int)
__main__:1: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
False
```

That is, instead of returning a two-dimensional boolean array, if the inputs are too big, the comparison is just `False`.  I think in a script the warning is ignored, by default.

The critical size seems to vary between machines.

This kind of comparison occurs inside `Mesh3d.boundary_edges`:

https://github.com/kinnala/scikit-fem/blob/1c37530da4412507a6fb5137d539376bae89d3c7/skfem/mesh/mesh3d/mesh3d.py#L78

For a big problem, the comparison returns the `False` and then 
```python
np.sum(False, axis=0)
```
is just 0, so `boundary_edges` returns `np.nonzero(0 * 0)[0] == (array([], dtype='int64'),)[0] = array([], dtype='int64')`, so nowhere is an exception raised.

I expect `boundary_edges` should be rewritten to avoid this, but the weird thing is that the solution looks fine even when `boundary_edges()` is empty as a result…  Still digging into that, but perhaps the function's not needed?  Sorry, I don't have an example to share just yet.

gdmcbain:
Fixes #196.

### DIFF
--- a/skfem/mesh/mesh3d/mesh3d.py
+++ b/skfem/mesh/mesh3d/mesh3d.py
@@ -74,9 +74,8 @@ class Mesh3D(Mesh):
 
     def boundary_edges(self) -> ndarray:
         """Return an array of boundary edge indices."""
-        bnodes = self.boundary_nodes()[:, None]
-        return np.nonzero(np.sum(self.edges[0, :] == bnodes, axis=0) *
-                          np.sum(self.edges[1, :] == bnodes, axis=0))[0]
+        return np.nonzero(np.isin(self.edges,
+                                  self.boundary_nodes()).all(axis=0))[0]
 
     def interior_edges(self) -> ndarray:
         """Return an array of interior edge indices."""


gdmcbain:
Testing on the slow example from #196 (a `MeshTet` with 24&thinsp;929 vertices), the step involving `mesh.get_dofs` (and so `boundary_edges`) took over five minutes on `master` but less than a second with this reimplementation.  The solutions were the same.

PR #199, opened 2019-06-20
class Form

### LINKED ISSUE #198: classes for forms
In the discussion of #192, there was an ([idea](https://github.com/kinnala/scikit-fem/pull/192#issuecomment-498123187)):
> to have the `@bilinear_form`, `@linear_form` and `@functional` decorators…return objects rather than functions (‘closures are a poor person's objects’) with appropriate `asm` methods

Although one of the motivations for this is that it should open up a lot of possibilities for working with forms:
* from the original sequence of elemental matrices #192, 
* through other ideas mooted there like the exploitation of highly parallel processing units, 
* to resurrecting automatic differentiation #27 or enabling symbolic manipulation #118 

the present issue is a call for a minimal refactoring.

gdmcbain:
Fixes #198 .


gdmcbain:
Besides passing `pytest -n auto test`, this runs all the examples without change.  

Profiling a larger unpublished example (two bilinear forms on 2 578 206 `ElementTetP1`) showed that assembly took 37 s on `master` and here, i.e. no discernable effect on speed.  (Incidentally, this 37 s is pretty minor compared to the 399 s ILU and 143 s GMRES.)

Issue #194, opened 2019-06-08
Roadmap to 1.0.0


kinnala:
Some future plans (ping @gdmcbain).

I hope that the release 0.2.0 will be quite close to what 1.0.0 looks like, and then after a while go directly to 1.0.0.

I've started listing some issues that I hope to resolve before/during 0.2.x into the corresponding milestone. But I think the most important ones are those that we know to require backwards incompatible changes. One of them could be the load/submeshes/boundaries/_parse_submeshes refactoring that we have discussed (depending on how it will be done).

I also wonder if there is something that we want to drop from the scope. My academic research was closely related to Nitsche-type mortar methods and I spent quite a bit of time thinking how they should be incorporated into a generic code like this, but I'm still very unhappy with the result and think that making this properly requires too much of my time. So I'm almost ready to drop `InterfaceMesh1d` etc.

Are there any other major changes that we'd like to see before 0.2.0?

gdmcbain:
The loading of submeshes and boundaries, mentioned in #180, but probably deserving its own issue, should be straightforward, I think, and not backwards incompatible. They'd be optional keyword arguments to the constructors. 

I've been dragged away from the work on coupling subdomains #163, which is very closely related to the Nitsche-type mortar methods. It was going well, I didn't foresee any real hurdles. I wasn't going to use `InterfaceMesh1d` as I didn't want to have to specify the interface in terms of coordinates; rather something like `Mesh.boundaries`. There's some more work to be done, but I didn't think it would require any changes to the library, unless eventually perhaps some additions to encapsulate a recommended way of doing things (analogous to the wrappers of `spilu` and `cg` in `utils`).

The one issue which does have me scratching my head is serialization #158, specifically of higher order spaces like P2 which can't be avoided for Navier-Stokes work. One hesitates to invent a new mesh format, but I don't know of an existing one that handles fields like `ElementTriP2` but where the underlying mesh only has corner nodes. 

Also maybe the elemental support #192. Does that require a reworking of the assembly? It might not break the interface though.

kinnala:
I started refactoring `from_meshio` in https://github.com/kinnala/scikit-fem/commits/prepare-release

[...]

PR #201, opened 2019-06-20
reserve Functional.assemble for returning the final float

### LINKED ISSUE #200: should functional return float rather than ndarray?
While implementing `class Form` #199 for #198, I noticed that the behaviour of `functional` differs from `linear_form` and `bilinear_form`; the latter assemble over the finite elements whereas the former doesn't.  This factoring of discretization into two steps (within each element, then assembling over mesh) might be useful for the latter too #192.

Thus far, `functional` is only used in ex22 where its result is passed to `adaptive_theta`, which wants an array of length equal to the number of elements; however, it could also be assembled, e.g. to calculate the area of a mesh, or the volume beneath a deformed two-dimensional membrane or plate; e.g. in the ex12, the `area` can be computed with
```python
sum(asm(functional(lambda w: 1.), basis)
```
or the volume as
```python
sum(asm(functional(lambda w: w.w), basis, w=basis.interpolate(x)))
```

Conversely, various ideas for elemental matrices from bilinear forms were mooted on gitter and in #192.

This is a suggestion for maybe something like a new `instancemethod` `Form.elemental` which returns:
* `functional`: `ndarray` (1-d array, being float per element)
* `linear_form`: `ndarray` (2-d array, being 1-d array per element)
* `bilinear_form`: `ndarray` (3-d array, being 2-d array per element)

plus in the last two cases maybe also the indices required for subsequent assembly; this 'subsequent assembly' in the `functional` case is simply `sum` so no indices are required.

This would naturally follow the resolution of #198.

gdmcbain:
Fixes #200.

### DIFF
--- a/docs/examples/ex22.py
+++ b/docs/examples/ex22.py
@@ -24,7 +24,7 @@ def eval_estimator(m, u):
         x, y = w.x
         return h**2 * load_func(x, y)**2
 
-    eta_K = asm(interior_residual, basis, w=basis.interpolate(u))
+    eta_K = interior_residual.elemental(basis, w=basis.interpolate(u))
     
     # facet jump
     fbasis = [FacetBasis(m, e, side=i) for i in [0, 1]]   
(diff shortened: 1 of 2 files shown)


gdmcbain:
Fixing #200 necessitates breaking the API since now `asm(functional, basis)` should be a `float` whereas it was, and was expected to be in ex22, an `ndarray` of `shape == (basis.nelems)`.  Therefore, here ex22 is updated, replacing `asm(functional, basis)` with `functional.elemental(basis)`, this being the new `instancemethod` equivalent to the old `functional.assemble`.

gdmcbain:
This proposal is minimal in that it doesn't go on to the obvious logical generalization of factoring the `assemble` method for `Forms`'s other subclasses, `LinearForm` and `BilinearForm`; that's less urgent (w.r.t. the roadmap to 1.0.0 #194) since it should not break the interface.  I'll launch a new issue for it though.

PR #211, opened 2019-06-23
Fix normal vectors for MappingIsoparametric

### LINKED ISSUE #85: Add hex facet assembly normal vector support
Continuation of #22. The following method is not implemented.

https://github.com/kinnala/scikit-fem/blob/02f6da0ab83bad658c3b11f503c9087ca30f0c84/skfem/mapping.py#L283

kinnala:
Fixes #85 when ready.


kinnala:
This was easier than I thought. I wonder if I missed something. I'd need some proper test for this before merge.

kinnala:
Turns out the tests weren't comprehensive. I need to spend some more time on this.

kinnala:
This is a tough one. Normal vector implementation is generic for `MappingIsoparametric` and it works for `MeshQuad` so I think the bug must be somewhere in `MeshHex`-specific parts. I added more tests that help in locating the problem. Going to start 1 hour debug session now.

kinnala:
`FacetBasis.normals` point to wrong directions.

kinnala:
`Y` seems to be wrong on line 121 in facet_basis.py for `MeshHex`, `ElementHex1`, `MappingIsoparametric`.

kinnala:
How the other tests can work?

kinnala:
`Y` is also wrong in the other tests (e.g. `IntegrateFuncOverBoundaryPart`) although the integrals work fine for some reason despite this.

kinnala:
`MappingIsoparametric.invF` returns some points outside the reference element for hexahedral elements. This might be the reason for all the problems. No fix yet.

kinnala:
Majority of tests are fixed. One convergence test still fails.

Issue #163, opened 2019-04-05
couple subdomains


gdmcbain:
Coupling domains defined on different meshes by an `InterfaceMesh1D` is demonstrated in ex04, but since a boundary value problem can be solved in a subdomain #159 #160, can problems on subdomains be coupled too?

What are the relative advantages of multiple subdomains and multiple domains?  
* It might be simpler to construct the mesh and bases on the interface between subdomains.     
  * An `InterfaceMesh1D` requires a `rule: Callable[[float, float], bool]` which might be awkward for more complicated shapes.
* Multiple domains needn't have matching meshes.  This is particularly important for problems with large deformation.
* The ease of either solving ‘monolithically’ versus ‘partitioned’, ‘staggered’, ‘segregrated’ (see, e.g. Piperno & Farhat [2001](http://dx.doi.org/10.1016/s0045-7825%2800%2900386-8) for the terminology) is probably about the same for subdomains and domains, given the flexibility of scikit-fem; e.g. ex04 demonstrates the use of `scipy.sparse.bmat` for monolithic solution across mismatched meshes.

This issue is to demonstrate the possibility of coupling subdomains; comparison can follow.

An obvious choice for a first example is one following the [insulated wire](https://kinnala.github.io/scikit-fem-docs/examples/ex17.html) ex17 #89 #90 and the solution for the wire with the coating kept cold ex26 #159 #160, extending the latter with a Laplace equation in the annulus to recover the solution of the former.  This belongs to the class of Poisson problems with discontinuous coefficients treated by Gustafsson, Stenberg, & Videman ([2018](https://arxiv.org/abs/1802.10430)).

The motivation is multiphysics like conjugate heat transfer, in which different subdomains are governed by different partial differential equations; however, this doesn't intrinsically distinguish between multiple subdomains and multiple domains, the difference will come down to practical implementation details, which is what is to be explored here.

gdmcbain:
Monolithic conjugate heat transfer is demonstrated in #218.

gdmcbain:
ex04 is broken without `InterfaceMesh1D` #215.

kinnala:
I'm now convinced that the simplest way to do this is to use a global Lagrange multiplier together with a stabilized method. I'll try to make an example at some point because I want to try out this approach.

PR #218, opened 2019-07-02
example on conjugate heat transfer in a plane channel

gdmcbain:
Eventually this example should be something like ex25 'Forced convection' but with a finitely thick solid wall.

This is published very prematurely (so far it's just conduction with unit properties in both phases) to demonstrate how to generate the mesh, following a discussion in scikit-fem/Lobby.

TODO:
* advection in fluid phase
* different thermal properties in each phase
* more interesting postprocessing

### DIFF
--- a/docs/examples/Makefile
+++ b/docs/examples/Makefile
@@ -7,7 +7,8 @@ all: ex01_solution.png \
      ex23.png \
      ex25.png \
      ex26_solution.png \
-     ex27-150.0-psi.png ex27-450.0-psi.png ex27-750.0-psi.png
+     ex27-150.0-psi.png ex27-450.0-psi.png ex27-750.0-psi.png \
+     ex28.png ex28-inlet-outlet.png
 %.png: %.py
 	python $<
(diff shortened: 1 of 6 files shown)


gdmcbain:
ex28

gdmcbain:
ex28
ex28-inlet-outlet

gdmcbain:
ex28-inlet-outlet
ex28

gdmcbain:
This example is ready now.

It demonstrates a ‘monolithic’ approach, which is easy in scikit-fem thanks to the facility of restricting operators to subdomains.

The other approach to conjugate heat transfer is ‘segregated’, solving each subdomain separately and then coupling along the interface #163; that might be investigated in future examples.  Segregated solution can be appealing because the thermal conductivity of the solid often far exceeds that of the fluid or it can be necessary if the phases are meshed separately.

kinnala:
Thanks. Fantastic work. Adding it to the test suite should be quite straightforward?

PR #228, opened 2019-07-28
Make expand=True the default behavior

kinnala:
Following the discussion from Gitter.


kinnala:
A bunch of tests, examples and documentation to change.

kinnala:
@gdmcbain this should be ready

kinnala:
Oops, obviously not since the documentation is not updated

gdmcbain:
Looks good! I haven't run anything, but every `diff` in `docs/examples` looks like an improvement.

PR #233, opened 2019-10-08
DOC: newton navier stokes

gdmcbain:
Oops.  Forgot that the linear matrix is the first part of the Jacobian of the full nonlinear operator, as in clear in the source code.

https://github.com/kinnala/scikit-fem/blob/5484f6bd4f2327ffd4f882d4b7b3be7b17d19858/docs/examples/ex27.py#L217


kinnala:
Thanks for these new contributions, I'll check them by the end of the week.

PR #235, opened 2019-10-25
ElementLineP2

gdmcbain:
The `ElementLineP2` has an interior dof.

### DIFF
--- a/docs/examples/ex15.py
+++ b/docs/examples/ex15.py
@@ -19,7 +19,6 @@ b = asm(load, basis)
 
 D = basis.get_dofs().all()
 
-x = 0*b
 x = solve(*condense(A, b, D=D))
 
 if __name__ == "__main__":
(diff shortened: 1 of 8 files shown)


gdmcbain:
_TODO:—_ Examples & tests.  

A couple of modifications are required to ex15 (can't use `I = m.interior_nodes()`) and ex16 (can't assume last dof is the terminal value).  &c.

gdmcbain:
I've modified exx 15, 16, 23 so that they all work if `ElementLineP1` is changed to `ElementLineP2`.  

However, I haven't thought of a neat way to be able to select the element programmatically to facilitate testing without changing the style and structure of the examples.

kinnala:
Thanks. I'd like to add some basic tests that use this new element. I'll look into it asap and review at the same time.

kinnala:
I added nodality and convergence tests

PR #236, opened 2019-10-30
WIP: demonstrate SLEPc on ex16

gdmcbain:
I was wondering how hard it would be to hand the matrices assembled here over to PETSc or SLEPc.  It turns out to be fairly easy.  Here's a quick example, repeating the Legendre eigenvalue problem of ex16.

I'll just post this here as a WIP for the moment as it's a bit bare (and I don't know when I'll get back to it) but it does show how to get started and it does work.

### DIFF
--- a/docs/examples/ex_slepc.py
+++ b/docs/examples/ex_slepc.py
@@ -1,0 +1,38 @@
+from sys import argv
+
+import numpy as np
+
+import slepc4py
+from petsc4py import PETSc
+from slepc4py import SLEPc
+
+import ex16
+
+
+def petsc_mat(A):
+    return PETSc.Mat().createAIJ(
+        size=A.shape,
+        csr=(A.indptr, A.indices, A.data))
+
+
+slepc4py.init(argv)
+
+E = SLEPc.EPS().create()
+
+E.setOperators(*map(petsc_mat, [ex16.L, ex16.M]))
+
+E.setType(E.Type.ARNOLDI)
+E.setDimensions(2 * len(ex16.ks), PETSc.DECIDE)
+E.setWhichEigenpairs(E.Which.SMALLEST_MAGNITUDE)
+E.setProblemType(SLEPc.EPS.ProblemType.GHEP)
+E.setFromOptions()
+
+E.solve()
+
+if __name__ == "__main__":
+
+    from os.path import splitext
+    
+    for n, k in enumerate(ex16.ks):
+        print('{:2d}  {:9.6f}  {:9.6f}'.format(
+            n * (n + 1), k, np.real_if_close(E.getEigenvalue(n))))


gdmcbain:
Hmm, interesting, yes.  By ‘how to split’ do you mean not the obvious physical splitting like in ex04 where each body has its own `Mesh` or in ex26 where there are `subdomains` with distinct properties but rather automatic decomposition of an otherwise homogeneous domain?

> When _&Omega;_ defines a complex geometry, it is best to use graph partitioners such as `Metis` [5] or SCOTCH [4] that will lead to more “balanced” decompositions (less communication, similar workload and such).


* Jolivet, P., Dolean, V., Hecht, F., Nataf, F., Prud'Homme, C. & Spillane, N. (2012). High performance domain decomposition methods on massively parallel architectures with FreeFem++. _Journal of Numerical Mathematics_ **20**(3–4):287–302. §4.1.1 ‘Mesh partitioning’
 [doi: 10.1515/jnum-2012-0015](https://doi.org/10.1515/jnum-2012-0015) ([_Semantic Scholar_](https://www.semanticscholar.org/paper/High-performance-domain-decomposition-methods-on-Jolivet-Dolean/0822897b95fdbb4b2bdede3403e31c2715d03ec3))

gdmcbain:
> assembling different parts of the problem in different nodes

Another way to do this is with [`dask.bag`](https://docs.dask.org/en/latest/bag.html).  This is very simple.  I actually used it in my first scikit-fem project (June 2018).  I won't post that here (too many extraneous details) but basically:
* Generate a sequence of objects containing input data describing each subproblem.
* [`dask.bag.from_sequence`](https://docs.dask.org/en/latest/bag-api.html#dask.bag.from_sequence)
* [`dask.bag.map`](https://docs.dask.org/en/latest/bag-api.html#dask.bag.map)
* Postprocess the resulting bag of results like any other `Iterable`.

I'll think about demonstrating this in another example.  Have you got a particular problem in mind?  This is more for cases with an

> obvious physical splitting like in ex04 where each body has its own Mesh or in ex26 where there are subdomains with distinct properties

where no graph partitioning is required.

kinnala:
Closing for now, but added a label so that it can be found later if needed.

docs/examples/ex_slepc.py
```diff
@@ -0,0 +1,38 @@
...
+from petsc4py import PETSc
+from slepc4py import SLEPc
+
+import ex16
+
+
+def petsc_mat(A):
+    return PETSc.Mat().createAIJ(
+        size=A.shape,
+        csr=(A.indptr, A.indices, A.data))
```
kinnala, line 15: Add this as `COOData.to_petsc`?

gdmcbain: Sure.  I'm not expert in PETSc and haven't touched this since, but it did work.  Should it go in #474 or #596?

PR #237, opened 2019-11-01
primitive Orr–Sommerfeld example

gdmcbain:
Here's a new example, a primitive variables (pressure–volume) version of the Orr–Sommerfeld equation of linear hydrodynamic stability.

This is the example that motivated the addition of `ElementLineP2` #235, since that's needed here for the velocity components of the one-dimensional Taylor–Hood elements.

Other features not demonstrated in other examples so far are:
* complex unknowns
* nonhermitian eigenproblem

These are both handled easily by `scipy.sparse.linalg.eigs`.

The example might be a bit specialized, so the documentation might need a bit more background.


gdmcbain:
ex_os

This is the figure produced by the example, reproducing figure 3.2 of Criminale et al and figure 4.19 of Drazin & Reid.  It's like the figure on [Wikipedia](https://en.wikipedia.org/wiki/Orr%E2%80%93Sommerfeld_equation#Solutions) except that that plots _&lambda;_ &equiv; &minus;j _&alpha;c_ so is effectively rotated 90° clockwise, and it's at Re = 5772.22 and _&alpha;_ = 1.02056 instead of 10 000 and 1, but still recognizable.

PR #250, opened 2019-11-15
WIP: 1-norm isoparametric invF termination

gdmcbain:
This is to demonstrate a point in the discussion of the use of an algebraic sum in the termination criterion for the Newton iteration in `MappingIsoparametric.invF` #249.  A 1-norm seems more appropriate but fails five tests:
* 3 in `test_assembly`
* 1 in `test_convergence`
* 1 in `test_mapping`

PR #252, opened 2019-11-15
tolerance for exteriority test in invF

### LINKED ISSUE #244: FacetBasis on MeshQuad
There seems to have been a regression involving `MeshQuad.boundaries` since July.  

A previously working (but too complicated to share) example now raises
```  File "/home/geordie/src/scikit-fem/skfem/assembly/global_basis/facet_basis.py", line 121, in __init__
    Y = self.mapping.invF(x, tind=self.tind)
  File "/home/geordie/src/scikit-fem/skfem/mapping/mapping_isoparametric.py", line 143, in invF
    raise ValueError("Inverse mapped point outside reference element!")
```
after a call like 
```python
FacetBasis(mesh, element, facets=mesh.boundaries['outlet'])
```

I'll reproduce this in a simpler example shortly.

### LINKED ISSUE #249: termination criterion for MappingIsoparametric.invF
While investigating failures in `FacetBasis` for `MeshQuad` #244, which raised an error in `MappingIsoparametric.invF`, suspicion fell on the termination criterion for the Newton iteration:

https://github.com/kinnala/scikit-fem/blob/70aa668560e76925a35f87b1c659a8e8f8ebda23/skfem/mapping/mapping_isoparametric.py#L140

It was asked there:
> Should this use a norm rather than a sum?

Since the elements of the decrement can be signed, the sum can involve cancellation or even be negative, thereby falsely passing the test.

Perhaps something like
```python
if (np.linalg.norm(dX, 1, (0, 2)) < 1e-6).all():
```
captures the intention.

### LINKED ISSUE #251: acceptance criterion for MappingIsoparametric.invF
After the termination criterion for `MappingIsoparametric.invF` #249 comes the acceptance criterion

https://github.com/kinnala/scikit-fem/blob/70aa668560e76925a35f87b1c659a8e8f8ebda23/skfem/mapping/mapping_isoparametric.py#L142-L143

It was pondered in the investigation into the regression of `FacetBasis` for `MeshQuad` #244 that 
> the test against exteriority isn't appropriate when applied to facet quadrature points, which ideally would lie on the boundary of the element and so be marginally exterior.

In the same example snippet as used in #249 
```python
m = MeshQuad()
e = ElementQuad1()
fb = FacetBasis(m, e, facets=[0])
x = fb.mapping.G(fb.X, find=fb.find)
Y = fb.mapping.invF(x, tind=fb.tind)
```
again stepping through the last call with `pdb`, but this time replacing the `np.sum` termination criterion with the `np.linalg.norm` #250 , the terminal value of `X` is
```python
array([[[-0.77459667,  0.        ,  0.77459667]],

       [[-1.        , -1.        , -1.        ]]])
```
which makes `x - self.F(X, tind)` vanish identically but `np.abs(X)` is
```
array([[[0.77459667, 0.        , 0.77459667]],

       [[1.        , 1.        , 1.        ]]])
```
which is clearly marginal for `(np.abs(X) > 1.0).any()`.

I think I can imagine that in a less trivial case than `MeshQuad()`, the `X` produced by the Newton iteration and satisfying the termination criterion might also have components satisfying `np.abs(X) > 1.0`, if only by a little.

gdmcbain:
Fixes #251 

Built atop #250 and so also fixes #249 (which #250 alone does not).

This also fixes #244, the original problem of broken Neumann problems relying on `FacetBasis` in `MeshQuad`.

### DIFF
--- a/skfem/mapping/mapping_isoparametric.py
+++ b/skfem/mapping/mapping_isoparametric.py
@@ -137,7 +137,7 @@ class MappingIsoparametric(Mapping):
             invDF = self.invDF(X, tind)
             dX = np.einsum('ijkl,jkl->ikl', invDF, x - F)
             X = X + dX
-            if np.sum(dX) < 1e-6:
+            if (np.linalg.norm(dX, 1, (0, 2)) < 1e-6).all():
                  break
         if (np.abs(X) > 1.0).any():
             raise ValueError("Inverse mapped point outside reference element!")
--- a/skfem/mapping/mapping_isoparametric.py
+++ b/skfem/mapping/mapping_isoparametric.py
@@ -139,7 +139,7 @@ class MappingIsoparametric(Mapping):
             X = X + dX
             if (np.linalg.norm(dX, 1, (0, 2)) < 1e-6).all():
                  break
-        if (np.abs(X) > 1.0).any():
+        if (np.abs(X) > 1.0 + 1e-12).any():
             raise ValueError("Inverse mapped point outside reference element!")
         return X
 


gdmcbain:
This remedy is admittedly a hack, but I think it's right in identifying the basic problem of #244, that a floating-point test that the facet quadrature points not be outside the associated interior element is dicey.

Is there a better way?  It seems to might that `invF` isn't quite the right function when computing the facet quadrature points.  I'm not yet familiar enough with the machinery to know whether there's way to know where a point is in the reference interior element given where it is in either the reference facet element (which is just `FacetBasis.X`) or in the facet element in global terms (`x` in `FacetBasis.__init__`).

gdmcbain:
I notice that `detDG` is defined for both subclasses of `Mapping` but not used anywhere.  Is it intended to be used in a `invG` method, being for facets what `invF` is for interior elements?

kinnala:
invF is used on purpose. There is no other straightforward way to evaluate
basis function gradients on the boundary as far as I know. I think this
solution is correct. We should add you minimal example as a test so we can
experiment with other approaches.

gdmcbain:
I suppose I was wondering whether there were a more direct relationship between Y and X than the composition of G and invF. If the mapping on the facet is consistent and  one knows which facet of the interior element one is dealing with then  I think Y is either :
* (2 X -1, -1)
* (+1, 2X-1)
* (1-2 X, +1) or
* (-1, 1-2 X),

isn't it? (Maybe with some sign errors.) Does one know at this point which facet of the element a facet is?

Is it that this is too specific to ElementQuad1 and one doesn't want to have to work this out for every other element?

PR #255, opened 2019-11-15
adapt exx17 & 26 from TriP1 to Q1

### LINKED ISSUE #248: create example and test for irregular MeshQuad
The regression in the handling of FacetBasis for MeshQuad #244 might have been avoided by having a tested example.  The bug seems only to bite when the mesh is irregular.  This could be covered by a minor modification of one of the examples using a `MeshTri` from `pygmsh` by using the Gmsh command [`Recombine Surface`](http://gmsh.info/dev/doc/texinfo/gmsh.html#index-Recombine-Surface-_007b-expression_002dlist_002dor_002dall-_007d-_003c-_003d-expression-_003e_003b).  

For example, ex13.

Or should an example be parameterized so that it can be run & tested either with triangular or quadrilateral mesh and elements?

gdmcbain:
Fixes #248 .

### DIFF
--- a/docs/examples/ex17.py
+++ b/docs/examples/ex17.py
@@ -28,7 +28,8 @@ def make_mesh(a: float,         # radius of wire
     insulation = geom.add_circle(origin, b, dx, holes=[wire.line_loop])
     geom.add_physical(insulation.plane_surface, 'insulation')
     geom.add_physical(insulation.line_loop.lines, 'convection')
-
+    geom.add_raw_code('Mesh.RecombineAll=1;')
+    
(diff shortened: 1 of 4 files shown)


gdmcbain:
Requires and includes #246, #252, and #254.

gdmcbain:
It might be worth extending this to plot the quadrilateral mesh rather than relying on `MeshQuad.plot` which draws the `MeshTri` from `_splitquads`.

Maybe use `MeshQuad.draw`?

gdmcbain:
ex17_solution

Issue #241, opened 2019-11-11
encapsulating Taylor-Hood


gdmcbain:
I'm thinking about how to structure a refactoring of some of the common code for Taylor–Hood elements, used for Stokes and Navier–Stokes equations, e.g.
* [ex18 ‘Stokes equations’](https://github.com/kinnala/scikit-fem/blob/master/docs/examples/ex18.py)
* [ex24 ‘Backward-facing step’](https://github.com/kinnala/scikit-fem/blob/master/docs/examples/ex24.py)
* [ex27 ‘Navier–Stokes equations’](https://github.com/kinnala/scikit-fem/blob/master/docs/examples/ex27.py)

as well as work in progress like #171 towards the three outstanding issues on these equations #169 #170 #240.

Each of these begins by repeating the same code to define the elements and interior bases.

https://github.com/kinnala/scikit-fem/blob/f5cc86f067555aac1c3d6449cb8028c8ad8ec283/docs/examples/ex18.py#L13-L16

There are some other common operations too.  I wonder to what extent these commonalities might be encapsulated.  

They might need to be parameterized for `intorder` (as mentioned in #240) and also for the type of mesh; although all the examples thus far (excluding [ex29](https://github.com/kinnala/scikit-fem/blob/master/docs/examples/ex29.py) the one-dimensional Orr–Sommerfeld equation which is a bit special) all use `MeshTri`, I think they should also work for `MeshQuad`, `MeshTet`, and `MeshHex`.  

Given a mesh, would it make sense for a function or class to automatically generate the appropriate pair of elements and interior bases for velocity and pressure?  Something like `if isinstance(mesh, MeshTri): … elif instance(mesh, MeshQuad): …`?

Would this sit in `skfem.models`?

Or could this be getting too specialized and should it be left outside the `scikit-fem` package?

kinnala:
I've actually experimented with related ideas in some early prototypes of the code (back before I was using Matlab). The idea was basically similar to `ElementVectorH1`: to have something like `ElementComposite` which takes as a parameter two elements `e = ElementComposite(ElementVectorH1(ElementTriP2), ElementTriP1)` and then initializing basis objects only for this composite object `e`. Then creating matrices could be done directly for `e` via `asm(bilinf, InteriorBasis(e))` where the definition of `bilinf` includes unpacking the unknowns from the "vectorial unknown" as `ux, uy, p = u`. I don't have any reference code at hand but it is definitely doable.

[...]

Issue #240, opened 2019-11-08
DOC: pressure-correction scheme for unsteady Navier–Stokes


gdmcbain:
Following the Newton steady solution of the Navier–Stokes equation in ex27 and the one-dimensional linear stability analysis in ex29, the next step is to show how to implement an algorithm for the unsteady Navier–Stokes equations.

While this could be done with a fairly straightforward extension of the steady solution (e.g. backward Euler or higher backward differentiation formula), a popular approach is the class of approximate ‘projection methods’ which decouple pressure and velocity.  This is what's used in the FEniCS tutorial.  So far I've only translated tutorial 07 into scikit-fem in gdmcbain/fenics-tuto-in-skfem#4.  It's not very interesting so I propose to develop something else for here but till then that serves as a bit of an example.

gdmcbain:
One thing that I noticed is that whereas the quadrature rules defined in the velocity and pressure interior bases in the previous examples of the Stokes and Navier–Stokes equations, ex18, ex24, ex27, e.g.

https://github.com/kinnala/scikit-fem/blob/9751aee9fc9402995075a3d6f775ff1e3a24ad54/docs/examples/ex18.py#L15-L16

have the `intorder` fixed to a common value which allows mixed bilinear forms (e.g. for pressure-gradient term in momentum equation or the velocity-divergence term in the continuity equation) to be assembled, the order isn't high enough when the velocity-mass bilinear form is required, as it is in most unsteady schemes.  The fix is simply to increase the common `intorder` to 4.  This was already done silently for the same reason in the linear hydrodynamic stability example ex29 #237.

gdmcbain:
A more substantial example is the reproduction in scikit-fem of FEniCS tutorial 08 under gdmcbain/fenics-tuto-in-skfem#5; see [fenics-tuto-in-skfem/08_navier_stokes_cylinder](https://github.com/gdmcbain/fenics-tuto-in-skfem/tree/master/08_navier_stokes_cylinder).

This works, but I don't aim to copy it directly to the docs/examples here; the example takes too long to run (17 minutes in FEniCS, 12 in scikit-fem) and I'm inclined to use a more modern algorithm (gdmcbain/fenics-tuto-in-skfem#13).

PR #257, opened 2019-11-22
DOC: improved animation of heat equation

gdmcbain:
While working on operator-splitting techniques for the unsteady Navier–Stokes equation #240, I found it would be helpful to have a better idiom for animating transient solutions.  This documents them for the simpler example of the heat equation, ex19.

I've also taken the opportunity to make a number of other improvements to the example:
* Optionally save an animated GIF and include this in the docs.
* This optional behaviour is governed by a command-line flag via [`argparse`](https://docs.python.org/3/library/argparse.html) from the Python Standard Library.
  * The default behaviour is as before to display the solution as it's being simulated.
  * No duplication of code between the two modes is needed.
  * This parameterization of an example might also be useful in other cases:
    * type and order of meshes & elements #241 #244 #248
    * fineness of mesh or time-step or tolerance, or degree of nonlinearity, or time to run an unsteady simulation, e.g. to make a test quicker but an example run live at the terminal richer
    * just about any other aspect of a finite element problem
* Animation is sped up using [`matplotlib.animation.FuncAnimation`](https://matplotlib.org/api/_as_gen/matplotlib.animation.FuncAnimation.html#matplotlib.animation.FuncAnimation). (I haven't actually measured this.)
* `MeshQuad` in place of `MeshTri`: more natural for the geometry and also gives another `ElementQuad1` example #248. 
* The mesh is generated with `MeshQuad.init_tensor` instead of using the product of two `MeshLine`s.
* The solution is generated in the script rather than only when `__name__ == '__main__'`, or at least a generator `evolve` is written.  This makes ex19 more like the other examples that can have the solutions imported into `tests.test_examples`. This example still can't be tested as it relies on `sksparse`, but it does demonstrate an idiom for making future unsteady examples testable.

### DIFF
--- a/docs/examples/Makefile
+++ b/docs/examples/Makefile
@@ -3,6 +3,7 @@ all: ex01_solution.png \
      ex05_solution.png \
      ex06_solution.png \
      ex17_solution.png \
+     ex19.gif \
      ex20_stream-lines.png ex20_velocity-vectors.png \
      ex23.png \
      ex25.png \
@@ -14,6 +15,8 @@ all: ex01_solution.png \
 	python $<
 %_solution.png: %.py
 	python $<
+%.gif: %.py
+	python $< --gif
 ex20_velocity-vectors.png: ex20.py
 	python $<
 ex20_stream-lines.png: ex20_velocity-vectors.png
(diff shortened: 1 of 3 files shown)


gdmcbain:
ex19

kinnala:
Looking good https://kinnala.github.io/scikit-fem-docs/examples/ex19.html

PR #259, opened 2019-11-26
correct check on cell_data in Mesh.save

### LINKED ISSUE #258: check checks on Mesh.cell_data in Mesh.save
I just noticed that the check on `cell_data` in `Mesh.save` looks awry.

https://github.com/kinnala/scikit-fem/blob/52ae9dff26af7d661b08bcb55c2faab2616b31e6/skfem/mesh/mesh.py#L290-L293

* Shouldn't check `point_data`.
* I think the type is actually `Dict[str, Dict[str, ndarray]]`.

gdmcbain:
Fixes #258.

### DIFF
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -266,7 +266,8 @@ class Mesh():
     def save(self,
              filename: str,
              point_data: Optional[Dict[str, ndarray]] = None,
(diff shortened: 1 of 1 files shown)


gdmcbain:
Closing this to simplify the type of the `cell_data` at the scikit-fem end, following afterthoughts in #258.

PR #263, opened 2019-11-29
no longer need to cast Path to str in Mesh.load

gdmcbain:
Following nschloe/meshio#546, as announced by @clbarnes in #120, it's no longer necessary to wrap a `pathlib.Path` in `str` (or take its `.name` attribute, &c.).


kinnala:
is this not released or why is the test failing?

gdmcbain:
Oh, is this working now?  It wasn't the last time I looked.  Have they fixed something upstream in meshio?

clbarnes:
It's in meshio master, not sure if it's been released yet.

gdmcbain:
Right, yes, it was `Mesh.save` to XDMF that was the problem.  nschloe/meshio#562 Something with the external XML library, I think.

PR #265, opened 2019-12-01
Fix inverse isoparam mapping

### LINKED ISSUE #256: move MappingIsoparametric.invF acceptance criterion into tests
In the investigation of the acceptance criterion for `MappingIsoparametric.invF` #251, it was [asked](https://github.com/kinnala/scikit-fem/issues/251#issuecomment-554668695) whether it were only to cover the previously leaky termination criterion #249 and whether it could be dispensed with after that was fixed #250 #252.  This resulted in the [suggestion](https://github.com/kinnala/scikit-fem/issues/251#issuecomment-554673525) that

>  it can be turned into a test instead so that it could discover similar bugs in the future.

This issue is to write that test.

I wonder also whether the for-loop of the Newton iteration should raise an exception if it is exhausted rather than finding a tolerable solution, as in

https://github.com/nschloe/pacopy/blob/5861d1f5fed29ca9475cff989a990e8583435fb3/pacopy/newton.py#L32-L35

or 

https://github.com/scipy/scipy/blob/3db12250601a0dd6e18b2677a8220bc2b001e12c/scipy/optimize/zeros.py#L360-L363

Actually such a check might suffice to catch whatever was wrong, since if the Newton iteration does converge, it converges to an `X` for which `F(X)` is close to `x` (the boundary quadrature points) and so can't really be too far outside the interior element.

kinnala:
Fixes #256 .

@gdmcbain can you review? I invited you to be a collaborator if it's fine to you.

### DIFF
--- a/skfem/mapping/mapping_isoparametric.py
+++ b/skfem/mapping/mapping_isoparametric.py
@@ -8,6 +8,10 @@ from skfem.element import Element
 from .mapping import Mapping
 
 
+NEWTON_MAX_ITERS = 50
+NEWTON_TOL = 1e-8
+
+
 class MappingIsoparametric(Mapping):
     """An isoparametric mapping, e.g., for quadrilateral and
     hexahedral elements."""
@@ -132,15 +136,16 @@ class MappingIsoparametric(Mapping):
     def invF(self, x, tind=None):
         """Newton iteration for evaluating inverse isoparametric mapping."""
         X = np.zeros(x.shape)
-        for _ in range(50):
+        for _ in range(NEWTON_MAX_ITERS):
             F = self.F(X, tind)
             invDF = self.invDF(X, tind)
             dX = np.einsum('ijkl,jkl->ikl', invDF, x - F)
             X = X + dX
-            if (np.linalg.norm(dX, 1, (0, 2)) < 1e-6).all():
+            if (np.linalg.norm(dX, 1, (0, 2)) < NEWTON_TOL).all():
                 break
-        if (np.abs(X) > 1.0 + 1e-12).any():
-            raise ValueError("Inverse mapped point outside reference element!")
+        if _ == NEWTON_MAX_ITERS - 1:
+            raise Exception(("Newton iteration didn't converge "
+                             "up to TOL={}".format(NEWTON_TOL)))
         return X
 
     def F(self, X, tind=None):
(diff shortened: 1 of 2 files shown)


skfem/mapping/mapping_isoparametric.py
```diff
@@ -8,6 +8,10 @@
 from .mapping import Mapping
 
 
+NEWTON_MAX_ITERS = 50
```
gdmcbain: Good idea replacing the 'magic numbers' 50 and 1e-6 with variables.  

Should they be defined closer (structurally) to where they're used?  And they're only used in one method, so what about having them as optional keyword arguments (which would never in practice be used); e.g.

```python
def invF(self, x, tind=None, newton_max_iters: int=50, newton_tol=1e-8)
```

skfem/mapping/mapping_isoparametric.py
```diff
@@ -132,15 +136,16 @@ def detDG(self, X: ndarray, find: Optional[ndarray] = None):
...
         X = np.zeros(x.shape)
-        for _ in range(50):
+        for _ in range(NEWTON_MAX_ITERS):
             F = self.F(X, tind)
             invDF = self.invDF(X, tind)
             dX = np.einsum('ijkl,jkl->ikl', invDF, x - F)
             X = X + dX
-            if (np.linalg.norm(dX, 1, (0, 2)) < 1e-6).all():
+            if (np.linalg.norm(dX, 1, (0, 2)) < NEWTON_TOL).all():
                 break
```
gdmcbain: Can we just
```python
return X
```
here?

Not important.

skfem/mapping/mapping_isoparametric.py
```diff
@@ -132,15 +136,16 @@ def detDG(self, X: ndarray, find: Optional[ndarray] = None):
...
-        if (np.abs(X) > 1.0 + 1e-12).any():
-            raise ValueError("Inverse mapped point outside reference element!")
+        if _ == NEWTON_MAX_ITERS - 1:
```
gdmcbain: This `if`-test wouldn't be required if `X` were returned from within the `for`-loop on convergence.

skfem/mapping/mapping_isoparametric.py
kinnala: Made the constants method local.

skfem/mapping/mapping_isoparametric.py
kinnala: Yes. It's more simple this way.

Issue #264, opened 2019-12-01
algebraic multigrid


gdmcbain:
I might make a start this week on algebraic multigrid. 

As mentioned earlier this year, this will be useful for the momentum steps of the Uzawa Stokes solver #171. It's also required for all three steps of the pressure-correction scheme for the unsteady Navier-Stokes equation in gdmcbain/fenics-tuto-in-skfem#5. Of course, once in place it should be useful for much else too.

I guess the three candidates are PyAMG, pyamgcl, and PETSc. Probably one of the first two for most purposes as they're much lighter.

I might start with optional extensions to ex9, comparing AMG to ILU as a preconditioner for conjugate gradients on the Poisson problem. 

I doubt this'll require any changes to the skfem library.

gdmcbain:
Fixed by #273, extending ex09 to show how to use either PyAMG or AMGCL, both being good fits for scikit-fem.  I'll defer PETSc.

gdmcbain:
pyamgcl also worked well with the three steps of the ‘IPCS’ (incremental pressure correction scheme) for the unsteady Navier–Stokes equations in gdmcbain/fenics-tuto-in-skfem#5.

PR #273, opened 2019-12-05
DOC: demonstrate use of PyAMG and AMGCL

gdmcbain:
For #264.

### DIFF
--- a/docs/examples/ex09.py
+++ b/docs/examples/ex09.py
@@ -27,6 +27,18 @@ Aint, bint = condense(A, b, I=I, expand=False)
 preconditioners = [None, build_pc_ilu(Aint)]
 
 try:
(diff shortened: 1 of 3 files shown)


gdmcbain:
Both [PyAMG](https://github.com/pyamg/pyamg) and [AMGCL](https://pypi.org/project/pyamgcl) turn out to be easy to use build effective preconditioners for `skfem.utils.solver_iter_krylov`, analogous to `skfem.utils.build_pc_*`.

This simple extension of ex09 (solving the Poisson problem in a cube) does not attempt to assess the relative merits of the two implementations of AMG, it just shows how they may be used in scikit-fem.

gdmcbain:
Fixes #264.

PR #274, opened 2019-12-06
Add more examples to the test suite

kinnala:
Trying to get these examples to run in Travis.

### DIFF
--- a/skfem/importers/meshio.py
+++ b/skfem/importers/meshio.py
@@ -1,6 +1,7 @@
 """Import any formats supported by meshio."""
 
 from typing import Tuple
+from collections import OrderedDict
 import warnings
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


gdmcbain:
LGTM but I'm not authorized to approve this on a mobile (which is a reasonable policy). A positive desktop review will have to wait till next week. 

Do you suppose that this technique would reasonably extend to pyamg or pyamgcl?

kinnala:
Yes, I think so. In general, I'd like to add a separate "testing requires" section to setup.py but that has to wait until I'm back at the university so this approach (installing dependencies in .travis.yml) can be used in the meanwhile

kinnala:
Now it also fixes #275 with a temporary solution.

skfem/importers/meshio.py
```diff
@@ -1,6 +1,7 @@
 """Import any formats supported by meshio."""
 
 from typing import Tuple
+from collections import OrderedDict
```
gdmcbain: I had to check whether `OrderedDict` was still required or whether a standard [`dict`](https://docs.python.org/dev/library/stdtypes.html#dict) would suffice.  It is till support for 3.6 is abandoned.

> Changed in version 3.7: Dictionary order is guaranteed to be insertion order.

kinnala: Ok, let's have it there for now then. I think 3.6 is supported officially until the beginning of 22?

kinnala:
Is this approved?

kinnala:
I'll postpone merge until I have time to investigate #276 and why it fails on my computer.

PR #276, opened 2019-12-09
Demonstrate broken test

kinnala:
@gdmcbain here is the broken example 17, let's see if it fails in Travis.


kinnala:
Apparently it did not fail. I'll investigate if I have old versions.

kinnala:
Not caused by old dependencies. I need to study this more carefully at a later date.

gdmcbain:
Did it pass now?

gdmcbain:
I know that some techniques of mesh generation use randomness, though I thought that under weight of complaint from users Gmsh had frozen the seeds of its pseudorandom number generators to give consistent results. That must be tricky across platforms, so I don't know to what extent they've achieved that.

Mostly it shouldn't affect us, but the other thing might be that some techniques for generating quadrilateral meshes don't guarantee no triangles ('quad-dominant'). That's acceptable in some applications (finite volume) but unacceptable here. I should check that the examples are only using techniques that generate meshes with a single type of cell.

gdmcbain:
When I run `ex17` here (either on master or demo-broken-ex17 branch), it works but I do see the line
```
Info    : Blossom recombination completed (0.0414269 s): 327 quads, 0 triangles, 0 invalid quads, 0 quads with Q < 0.1, avg Q = 0.72615, min Q = 0.494
```

I suppose ‘0 triangles’ might imply that sometimes there might be triangles other times?

Or could it be that the output MSH 4.1 has an empty block of triangular elements and `meshio.read` turns this into an item `'triangle': []` in the `.cells` attribute.  I don't know whether either of those would actually happen, but if so it could be enough to confuse `skfem.importers.meshio.from_meshio`.

I asked a colleague who has generated a few quadrilateral meshes in Gmsh.  They don't recall ever seeing any triangles, but that was for `gmshToFoam` and OpenFOAM, which I don't think would have minded.

gdmcbain:
```c++
// The default recombination algorithm might leave some triangles in the mesh,
// if recombining all the triangles leads to badly shaped quads.
```
https://gitlab.onelab.info/gmsh/gmsh/blob/master/tutorial/t11.geo#L52

Issue #286, opened 2019-12-24
Get rid of matplotlib depedency?


kinnala:
Should we move the plotting related functionalities from `Mesh` classes into separate files so that `matplotlib` can be an optional dependency? I've really enjoyed the simplicity of
```
m = MeshTri()
m.draw()
```
when experimenting and would like that similar workflow remains but I suppose it would be enough to have
```
m = MeshTri()
skfem.visualize.draw(m)
```
and
```
m = MeshTri()
skfem.visualize.plot(m, m.p[0, :])
```
to do simple plotting.

What do you think @gdmcbain ?

gdmcbain:
Yes, if possible. Matplotlib seems to bring in a lot of other things. I was puzzled to see an entry for pyparsing Japanese in the output of cProfile for #280 recently.

kinnala:
I can work on this, shouldn't be a huge refactor.

kinnala:
I've also considered of building a separate `meshplots` library which would utilize also other tools such as `plotly` and allow simple plotting from all sorts of mesh formats. But moving these into separate modules should suffice for now.

gdmcbain:
I've also been experimenting on three-dimensional visualization with:
* yt: but it only reads Exodus II and doesn't seem to like the flavour from meshio
* pyvista: but it doesn't like my graphics card or Mesa or something.

kinnala:
Remaining work:
- [ ] add the removed functionality to `skfem.visualize`
- [ ] fix examples and tests

gdmcbain:
Running
```shell
parallel python {} ::: docs/examples/ex*.py
```
reveals a few more examples to be updated: exx 04, 13, 18, 21, 24:
```
AttributeError: 'MeshTet' object has no attribute 'draw'
```
or
```
AttributeError: 'MeshTri' object has no attribute 'plot'
```

PR #287, opened 2019-12-24
Make matplotlib an optional dependency

kinnala:
Starting work on #286 .


kinnala:
@gdmcbain can you make a quick review at this point? Comments on the overall structure would be appreciated and if you have any ideas or alternative implementation strategies.

I'll continue with fixing rest of the examples after you think this is the right way to go forwards.

gdmcbain:
Another thought for the further future: Thus far the visualization is based on the `skfem.Mesh` class hierarchy but it also partially implies `ElementLineP1` for `MeshLine`, `ElementTriP1` for `MeshTri`, etc., or sometimes P0, Q0, etc., I think.  Some visualization backends (Gmsh, ParaView?, ...) might support higher order elements.  Does eventually supporting higher order visualization imply anything about the structure of the `skfem.visualization` package?  Should the functions take bases rather than meshes?  (Noting that an `skfem.assembly.GlobalBasis` has a `.mesh` attribute anyway.)

kinnala:
Fantastic suggestions, as usual. Here is a notebook I used during development.
[plotting.pdf](https://github.com/kinnala/scikit-fem/files/4008048/plotting.pdf)

I'm going to implement also the possibility of drawing through InteriorBasis since I removed the plotting of piecewise-constant and DG functions and refinterp should be used instead. I was thinking syntax like this:
```python
plot(interior_basis, solution)
```
and the call is the passed to `interior_basis.refinterp` and, e.g., `plot(m: MeshTri, ...)`.

kinnala:
Next I start fixing the examples. It might take a while.

kinnala:
Starting by running the Makefile under docs/examples and fixing all errors. Later fixing the docs.

kinnala:
Starting to be ready for review. Fixes #286 .

kinnala:
Am I missing some examples?

PR #292, opened 2020-01-01
Add CHANGELOG and rename GlobalBasis to Basis

kinnala:
Would this naming make sense to make it clear that InteriorBasis, FacetBasis, (and MortarBasis) are specialized versions of (Global)Basis?


kinnala:
Added also a changelog so that we can start keeping track of backwards-incompatible changes.

Issue #290, opened 2019-12-29
Add pacopy to the test suite


kinnala:
https://github.com/kinnala/scikit-fem/blob/57d1962a3cc9e70fe30712c1bdc6f7afdefeaaf8/tests/test_examples.py#L174

gdmcbain:
This is a good idea. We'll probably want to shorten ex27 though. Or rather have it run in a short way if imported, as in the test suite, but longer if invoked from the command-line, to generate the nice figures for the documentation; say fewer and lower milestones in Reynolds number for the test, probably just the first nonzero will do.

PR #295, opened 2020-01-01
test pacopy

gdmcbain:
This is for #290.

Just starting with ex23 till I figure out a tidy way to truncate ex27 for testing.


gdmcbain:
Using `if __name__ == '__main__'` to restrict the Reynolds number to 50 when ex27 is called as a test.  This is much quicker, so that the whole suite runs here with `pytest -n auto tests/test_examples` in half a minute; otherwise running up to Re = 750 to produce the figures for the documentation takes about five minutes.

Issue #291, opened 2019-12-29
Move Mesh.save logic to a separate module


kinnala:
I think it would make sense to move this logic to skfem.exporters or similar:

https://github.com/kinnala/scikit-fem/blob/fb542a7eec7db000982b2c4e47465a53de21a86f/skfem/mesh/mesh.py#L262

Also, there isn't always 1-to-1 mapping with the solution+mesh and their representation in a specific format (such as VTK). Maybe we should make this explicit, i.e. export using a combination of a solution vector and a GlobalBasis object, and make the conversion to a representation such as VTK happen automatically if there will be no loss of information, and display a warning if the conversion required interpolation/projection.

gdmcbain:
+1 for skfem.exporters though I think we should continue to use meshio where possible and aim to fix or extend it elsewhere. In some cases it's true that different formats aren't isomorphic but often it's meshio that is incomplete. I think we get a multiplier for doing the work there rather than here, where possible; more people using, testing, and reviewing the code, writing extensions, updates, optimizations, &c.

gdmcbain:
+1 for skfem.exporters though I think we should continue to use meshio where possible and aim to fix or extend it elsewhere. In some cases it's true that different formats aren't isomorphic but often it's meshio that is incomplete. I think we get a multiplier for doing the work there rather than here, where possible; more people using, testing, and reviewing the code, writing extensions, updates, optimizations, &c.

A possible exception to that thinking, which would particularly justify the `skfem.exporters` package, is for formats which have Python APIs which are eschewed by meshio (to avoid dependence). There might be cases where these offer advantages over meshio (faster, more complete, more authoritative, access to specific functionality beyond mere conversion, e.g. VTK, Gmsh).

gdmcbain:
A partial lightening of the burden of dependencies incurred with meshio was achieved with the eschewal of the `lxml` package for writing XDMF in nschloe/meshio#613.

kinnala:
Should be moved to skfem.io after #305 is merged.

Issue #271, opened 2019-12-04
load auxiliary data with meshes


gdmcbain:
This is kind of the opposite of #268, the removal of the old `.external` attribute of a loaded `Mesh`, which served as a catch-all for whatever else had been read from the file and wasn't immediately interpretable as part of the mesh.  That attribute was no longer populated by `Mesh.load` and so was dropped #269.

However, there are cases in which it makes sense to load auxiliary data along with a mesh, e.g. fields that are defined on the points or cells of the mesh; e.g.
* reloaded a saved solution for continuation (transient evolution, numerical continuation, &c.)
* importing a solution from another code
* continuously variable coefficients (not so suited to modelling with subdomains as in exx 17, 26, 28)

or subsets of the points or cells; e.g. 
* subdomains or boundaries #261  

`meshio.Mesh` does have attributes `point_data`, `cell_data`, and `field_data` and has more recently added:
* `node_sets` and `element_sets` for Exodus (only)
* `point_tags` and `cell_tags` for MED (only)

Ignoring these latter for the moment (I would hope that they'd be standardized in meshio eventually), should `Mesh.load` return both a mesh and some kind of container of auxiliary data?  Maybe return `Tuple[Mesh, MeshData]` where `MeshData` is an object like a `meshio.Mesh` without the `.points` or `.cells`?  I guess the `.cell_data` and `.field_data` would have the `'gmsh:physical'` items popped.

I'm not sure that auxiliary data belongs in an object with the mesh (which should just be geometric and topological data), although it does make sense to have data that is associated with a particular mesh, and does make sense to be able to store mesh and auxiliary data together in a file (as many formats allow).

kinnala:
When you have one value per node you also implicitly think of piecewise linear interpolation between the nodes which corresponds to `ElementTriP1` or similar.

I think it makes sense to read this kind of information from the files. However, there is a similar kind of "vagueness" than in the solution vector returned from `solve`: it doesn't really mean anything unless you know which `Element` and which `Mesh` it corresponds to.

I'm trying to think what kind of container would be sufficient but simple enough.

gdmcbain:
> However, there is a similar kind of "vagueness" than in the solution vector returned from solve: it doesn't really mean anything unless you know which Element and which Mesh it corresponds to.

[...]

Issue #261, opened 2019-11-27
Mesh.save subdomains and boundaries


gdmcbain:
Should [`Mesh.save`](https://github.com/kinnala/scikit-fem/blob/52ae9dff26af7d661b08bcb55c2faab2616b31e6/skfem/mesh/mesh.py#L266-L297) save the `.subdomains` and `.boundaries` attributes?  In a way that they might be reread by [`Mesh.load`](https://github.com/kinnala/scikit-fem/blob/master/skfem/mesh/mesh.py#L299-L311).

If I/O is to continue to rely on `meshio.write` and `meshio.read` (as it should), I think that saving boundaries requires saving cells of type `bnd_type` (e.g. `'line'` for `MeshQuad`); currently `skfem.Mesh.save` only saves cells of type `Mesh.meshio_type`.

(I'm currently parsing a `polyMesh` from OpenFOAM and constructing an `skfem.Mesh`.  My parser is very slow so I want to be able to save the mesh while experimenting with the solver.)  

Perhaps instead for the moment I should just construct a `meshio.Mesh` and `meshio.write` that, leaving scikit-fem out of it, but this might be something to think about for other cases, e.g. when the mesh has been refined or adapted inside scikit-fem.

kinnala:
I think this shouldn't be a huge task to implement. I'll keep it in mind for future coding sessions.

gdmcbain:
Not a huge task to implement, no, but maybe there is a preliminary question to think about: ‘tagged subsets of elements’ nschloe/meshio#290.

The current `skfem.Mesh.load` looks for a key `'gmsh:physical'` in the `meshio_type` and `bnd_type` entries of the `.cell_data` attribute of the `meshio.Mesh`.  These are interpreted as one-dimensional integer arrays of length equal to the number of `meshio_type` and `bnd_type` cells specified in the `meshio.Mesh.cells`.

However, the `.subdomains` and `.boundaries` attributes of a `skfem.Mesh` are different; they're `Dict[str, ndarray]`, with the terms of the `ndarray` values being indices of `meshio_type` or `bnd_type` belonging to the subdomain or boundary.

These two kinds of representations:

* {cell} → {subdomain} 
* {subdomain} → {subset of cells}

are discussed in nschloe/meshio#175, nschloe/meshio#290, nschloe/meshio#347, …

[...]

PR #299, opened 2020-01-04
0.4.0 release

kinnala:
Sort of "1.0.0 prerelease". Hopefully one of the last ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.4.0] - 2020-01-03
+
 ### Changed
 - Renamed GlobalBasis to Basis.
 - Moved all Mesh.plot and Mesh.draw methods to skfem.visuals module.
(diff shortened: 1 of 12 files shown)


gdmcbain:
Re 1.0.0: I'm thinking that there are going to be a few changes soon over at meshio nschloe/meshio#619 nschloe/meshio#568 which might affect i/o here #261 #271 #291.  I'm not sure how that affects the roadmap.  Is i/o separate enough that it won't?

nschloe:
You can always depend on "meshio < 4", its API is stable across major versions. You can then make the switch whenever you like.

gdmcbain:
Ta. Yes, no, I'm not so much worried about adapting here to the changes scheduled thus far for meshio 4 (transposition of keys of `cell_data`), that'll be straightforward; it's more that I'm hoping to take advantage in scikit-fem 1 of some of the other advances that might make it in like a more complete implementation of XDMF 3 or the rolling out of `cell_sets` to formats other than MED. Those will permit substantial improvements here.

nschloe:
Ah cool. Let's make sure that we get all of this implemented in meshio then!

Issue #301, opened 2020-01-07
transposition of meshio cell_data keys


gdmcbain:
As mentioned in the context of 0.4.0 release #299, nschloe/meshio#626 has transposed the keys of `meshio.Mesh.cell_data` to name before cell-type.  This is in meshio master but hasn't been released yet.

Using meshio master, this breaks ex21.py which raises
```
/home/geordie/src/scikit-fem/skfem/importers/meshio.py:114: UserWarning: Unable to load tagged boundaries/subdomains.
  warnings.warn("Unable to load tagged boundaries/subdomains.")
Traceback (most recent call last):
  File "docs/examples/ex21.py", line 28, in <module>
    dofs['fixed'].nodal['u^1'],
TypeError: tuple indices must be integers or slices, not str
```

The problem is that the transposition needs to be reflected in

https://github.com/kinnala/scikit-fem/blob/f8909f5d1f95ec2f83ee5141b1f3ce43b5591894/skfem/importers/meshio.py#L77-L78

and elsewhere following.

gdmcbain:
Subsumed by #321.

PR #302, opened 2020-01-07
WIP: adapt to transposition of cell-data keys in meshio

gdmcbain:
This is for #301 and is to cope with upstream changes in nschloe/meshio#626.

It might be better not to merge it into master until after the next meshio release, meshio 4, discussed in #299 and nschloe/meshio#615, possibly ‘about a month from now’.  I'll put it together though so that it's tested and ready to go.


gdmcbain:
This needs to wait till nschloe/meshio#627 is resolved, e.g. by merging nschloe/meshio#628, and then again till the next meshio release.

kinnala:
Maybe we should also add to setup.py the explicit requirement of meshio>4 if that will be the case?

gdmcbain:
Dropping in favour of #338.

Issue #300, opened 2020-01-06
starting guess for solve


gdmcbain:
SciPy's iterative solvers like [`scipy.sparse.linalg.cg`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.cg.html#scipy.sparse.linalg.cg) optionally accept `x0`, a ‘starting guess for the solution’.  This could be passed to `skfem.utils.solver_iter_krylov` or similar, but in some situations, e.g. a transient simulation with unchanging evolution operator, the previous solution might be a good guess and this changes at each time.  One could reconstruct the solver at each time step (saving the preconditioner and other constant options), but maybe it would be tidier to be able to pass `x0` directly to `skfem.utils.solve` after the `LinearOperator A` and right-hand side `b`?

kinnala:
Resolved by #304 .

PR #304, opened 2020-01-11
Pass **kwargs to solvers

kinnala:
@gdmcbain : is this what you propose in #300 ?


skfem/utils.py
```diff
@@ -204,14 +191,15 @@ def solver(A, b):
 
 def solver_iter_pcg(**kwargs) -> LinearSolver:
     """Conjugate gradient solver, specialized from solver_iter_krylov"""
-    return solver_iter_krylov(spl.cg, **kwargs)
+    return solver_iter_krylov(**kwargs)
 
 
 def solve(A: spmatrix,
```
kinnala: See, in particular, this part and the related changes in all of the solvers.

gdmcbain:
No, I didn't foresee not being able to define a solver with a fixed preconditioner.

skfem/utils.py
```diff
@@ -160,8 +153,7 @@ def build_pc_diag(A: spmatrix) -> spmatrix:
 
 
 def solver_iter_krylov(krylov: Optional[LinearSolver] = spl.cg,
-                       verbose: Optional[bool] = False,
-                       **kwargs) -> LinearSolver:
```
gdmcbain: I think we do still want to be able to parameterize Krylov solvers on definition; see #307.

skfem/utils.py
```diff
@@ -187,7 +174,7 @@ def callback(x):
         if verbose:
             print(np.linalg.norm(x))
 
-    def solver(A, b):
+    def solver(A, b, **kwargs):
```
gdmcbain: Yes, we do want to accept here keyword-arguments for passing on to the solver at solve-time but taking care not to smother those passed in here for defining the solver.  I'm thinking of cases like a heat equation for which the solver is used repeatedly with the same operator `A`, preconditioner `M`, &c.,  but changing right-hand side `b` and starting guess `x0`.

gdmcbain: See #307.

kinnala: Makes sense, let's go with this approach.

kinnala:
Sorry for doing the reordering in the same PR. I didn't realize the changelog would become so unreadable. I'll try to avoid this in the future.

Issue #215, opened 2019-06-25
ex04 broken without InterfaceMesh1D


gdmcbain:
Following 078fd379cc550976ab6264feed6a44767de48afc for #195, `python docs/examples/ex04.py
```
NameError: name 'InterfaceMesh1D' is not defined
```

kinnala:
Either I delete the example or recreate the logic of matching facets in InterfaceMesh1D for this specific example only. Thanks, I'll think about it. I also add this to the test suite if it will remain.

kinnala:
I've started again to work with these mortar methods so I'll probably create better support at some point. I'd like to also cover 3D cases but it's going to be much trickier to match elements over the interface.

gdmcbain:
Is the conforming special case (in which the facets are shared across the interface) a stepping stone to the general case?  The monolithic handling of multiple subdomains is well established in the examples but I imagine that the same problems could  be solved better on the same meshes with a segregated technique, connected by Nitsche conditions on the common facets? The generalization of that to three dimensions seems straightforward but I would have no idea even where to begin with a general nonconforming nonplanar interface.

kinnala:
It could be thought as such. But one must be careful to not make any assumptions that are not true in the general case. Obviously matching the facing elements should be much easier in this case. In the general case you create sort of "union mesh" consisting of the elements of both sides. Sometimes this "union mesh" might contain really nontrivial polygonal shapes (not necessarily triangles for tetrahedral domains). These nontrivial polygons must be further splitted into triangles for numerical integration.

gdmcbain:
> nontrivial polygonal shapes (not necessarily triangles for tetrahedral domains). These nontrivial polygons must be further splitted into triangles for numerical integration.

Ah yes, I hadn't thought of that (whereas along an interfacial line in two dimensions the intersections of lineal segments are still always lineal segments), but it's obvious, yes.

What about the special case of a planar interface?  In the general curved case (even in two dimensions) I don't even know what would be meant by a union mesh.  Is some underlying continuous representation of the geometry of the interface required?  So that one knows what surface one is projecting onto?

kinnala:
In curved case you need to either project one side to another or a smooth
parametrized manifold (between the domains) onto which you project both
sides. Depends on the use case which works best. I've seen some cool
electrical engine discretizations using the latter. Contact mechanics
usually relies on the former.

gdmcbain:
O. K., thanks, that makes sense.

[...]

PR #305, opened 2020-01-11
Mortar methods generalizations and fixes

kinnala:
This branch contains my work related to mortar methods.


kinnala:
I'm still trying to figure out what changes are needed in order to support more general cases (such as contact problems where the boundaries do not initially match) in the future.

gdmcbain:
Indeed... I guess just trying increasingly trickier examples might reveal what works and what else might be needed?

Is one still fairly simple example that does have a discretely nonmatching interface the insulted wire, with the wire and insulation meshed independently at different mesh-sizes so that the two sides of the interface are different polygonal approximations of the one geometric circle?

kinnala:
Thanks for the idea of circular mortar boundary. I'll try to support it also.

kinnala:
Now it's on a really ugly temporary state but I think the idea is now there. See the latest commit message for some spoilers. I'll continue on this next week.

kinnala:
Fixes #215 .

kinnala:
Instead of having explicit interface meshes, I've decided to use so-called mortar mappings that use similar interface meshes underneath. Then the syntax becomes quite nice since you only need to initialize two MortarBasis objects like this:
```python
mb = [
    MortarBasis(m, e, mapping = mortar[0], intorder=2),
    MortarBasis(M, E, mapping = mortar[1], intorder=2)
]
```
where `mortar[0]` and `mortar[1]` correspond to the different sides of the interface. Here `(m,e)` and `(M,E)` are the two meshes and their elements.

kinnala:
This is still in very raw state. I'll continue tomorrow. First I create a new example with "real" contact problem and then I tidy up the code.

kinnala:
kontakti

kinnala:
mises_p2_vs_q2

von Mises stress when calculating with TriP2 and Quad2 elements.

kinnala:
Remaining work:
- Separate MortarMapping code into a separate file and delete mesh_mortar.py
- Make creation of MortarMapping more robust to various edge cases
- Add tests

kinnala:
Circular mortar boundary is not yet working/tested but it should be basically simple as adding a new constructor to MortarPair with slightly different definition for the projection (functions proj and param).

kinnala:
I'm still writing ex04.rst but it will take a while because even writing down the equations for contact problems takes some meditation.

kinnala:
Here are the docs https://kinnala.github.io/scikit-fem-docs/examples/ex04.html

kinnala:
> I might try to reproduce the solution for the insulated wire ex17 using this

We would need to first add support for other projections than onto a predefined linear function.

In principle, it should be as easy as using different `proj` and `param` (local functions used in `MortarPair.init_2D`). In the case of insulated wire, you should probably do the projection via `arctan` of the angle of the point in polar coordinates.

In the literature, there are also ways to generalize this (such as closest point search) but I have not looked into it. I just know that it is possible.

My next goal is to create `MortarPair.init_3D` for mortaring 3D problems. It is somewhat more involved to create the integration mesh. My current plan is to use libsupermesh (https://bitbucket.org/libsupermesh/libsupermesh/src) and possibly create Python bindings for it. The idea is same: project points into d-1 dimensional domain, create integration mesh (or "supermesh", in `MortarPair.init_2D` it is simply one call to `np.unique`), generate d point quadrature rule in the integration mesh and map them into the global coordinate system to check that the orientation of the points matches over the global mortaring interface. If orientation does not match, perturb index columns in the auxiliary meshes so that they match.

PR #311, opened 2020-01-18
Add drawing tests. Run tests also for 3.7.

kinnala:


kinnala:
Drawing was failing on 3.6 which I didn't notice because tests didn't cover it. Also, add 3.7 to the test suite. Cannot add 3.8 due to conda (or some of our dependency) not supporting it yet properly.

kinnala:
I'll release these as 0.4.1 as soon as I'll fix the new tests because this might affect potential new users.

kinnala:
Yes, I think that is the case. I'll try it out.

Issue #314, opened 2020-01-21
S2


gdmcbain:
Another couple of elements #23, the quadratic serendipty elements on quadrilaterals and hexahedra, with eight and twenty nodes respectively, are called S2 in the [Periodic Table of the Finite Elements](http://femtable.org/).  I think they're fairly common; the 20-node hexahedron seems to be the default element in ANSYS, e.g.

What might they be called here?  `ElementQuadS2` and `ElementHexS2`?

kinnala:
Simple implementation for `ElementQuadS2` would be
```
class ElementQuadS2(ElementQuad2):
    interior_dofs = 0
```
I didn't try it out but I don't see why it wouldn't work.

gdmcbain:
Neat!  Ta.

gdmcbain:
Actually I think the basis functions are different; they don't have to vanish at the interior node. 

A systematic procedure is outlined by 

* Zienkiewicz, O. C., Morgan, K. (2006). *Finite Elements and Approximation.* Mineola, New York: Dover, §4.6.1

I'm going through it now.

gdmcbain:
Viz., in case the reference isn't to hand:
* The midpoints have a quadratic across their edge by a linear the other way.
* The corners are the Q1 corners minus enough of the neighbouring midpoints to cancel their values there.

kinnala:
> Actually I think the basis functions are different; they don't have to vanish at the interior node.

Sorry, my mistake. Obviously my earlier hint cannot be correct.

Issue #318, opened 2020-01-21
mesh-convergence in structural vibration


gdmcbain:
I'm a bit puzzled at the eigenfrequencies in the ‘structural vibration’ [ex21](https://kinnala.github.io/scikit-fem-docs/examples/ex21.html) when the mesh is refined; they don't seem to converge before my machine runs out of memory.  This is using the geometry of ex21 but also other simple ones like flattish elongated boxes like beams or plates, including ones with multiple materials #315.

Of course boundaries aren't preserved during refinement #104, but in ex21 they're easily recovered with 
```python
m.boundaries['fixed'] = m.facets_satisfying(lambda x: x[2] == 0.)
```
after the call to `m.refine`.

Here's my minimally modified ex21: [beams.py.txt](https://github.com/kinnala/scikit-fem/files/4094189/beams.py.txt).  Calling it with `python beams.py -r 0`, `-r 1`, `-r 2` gives:

0. [ 80.17036319 179.80324866 214.18727331]
1. [ 61.02158326 135.15096073 149.1143932 ]
2. [ 49.63520068 110.2739215  118.82280969]

and `-r 3` runs out of memory.

The eigenfrequencies of beams &c. similarly don't converge to the known analytical solutions, whereas I've checked in another code (ANSYS) using the same mesh and they do.  (This was behind my interest in serendipity elements #314 but I subsequently found that ANSYS also got accurate answers on coarse meshes with linear 8-node hexahedra.  It's hard to know exactly what ANSYS is doing, of course, but I don't expect anything too fancy here.)

I'll checkout the WIP #317 without threading and try to see more clearly what the linear elastic forms are doing.

Most likely I'm making a small simple error somewhere, perhaps of interpretation, perhaps a little bug in the script.

gdmcbain:
I am getting much less mesh-dependent eigenfrequencies in the two-dimensional ex03:

0. (50, 10) [0.01029339 0.05554975 0.08188535 0.13264495 0.22105733 0.24485105]
1. (100, 20) [0.0102668  0.05535409 0.0818689  0.1320289  0.21972848 0.24472409]
2. (200, 40) [0.01025974 0.05530334 0.08186304 0.13187295 0.21939837 0.24468756]

kinnala:
This is related #75 . Apparently the name of the example changed?

gdmcbain:
Duplicate of #75.

PR #319, opened 2020-01-22
ElementQuadS2

gdmcbain:
This is to fulfil the two-dimensional part of #314.


gdmcbain:
I've tested it locally on the two-dimensional linear elastic eigenproblem: 
[ex03_S2.py.txt](https://github.com/kinnala/scikit-fem/files/4095397/ex03_S2.py.txt).  It gives similar answers (i.e. not evincing the difficulty of #318 which seems limited to three dimensions) but takes a little while to run so I don't think ex03 should be modified to use S2 (or Q2).

gdmcbain:
Could also drop the third of dofnames since there are no interior dofs.

gdmcbain:
I wasn't sure about maxdeg. It's 3 for ElementQuad2; is that right, rather than 4?

kinnala:
You could add a simple convergence test like this: https://github.com/kinnala/scikit-fem/blob/master/tests/test_convergence.py#L136

PR #324, opened 2020-01-28
Fix the order of Lame parameters

kinnala:
I ran into an issue with the convergence of ex04 and then realized that the Lamè parameters in `skfem.models.elasticity.lame_parameters` were defined wrong way around (when passed on directly to `skfem.models.elasticity.linear_elasticity`. Do you agree @gdmcbain?

How does this reflect with your earlier tests on the eigenvalues? Do the conclusions still hold? I had to modify the test result a bit. I suppose we need some sort of convergence test for this operator.

### DIFF
--- a/docs/examples/ex04.py
+++ b/docs/examples/ex04.py
@@ -82,19 +82,19 @@ for i in range(2):
 
         @BilinearForm
         def bilin_mortar(u, v, w):
-            ju = (-1.) ** i * dot(u, w.n)
-            jv = (-1.) ** j * dot(v, w.n)
+            ju = (-1.) ** j * dot(u, w.n)
+            jv = (-1.) ** i * dot(v, w.n)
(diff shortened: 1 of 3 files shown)


gdmcbain:
> How does this reflect with your earlier tests on the eigenvalues?

‘Luckily‘, my tests were at _&nu;_ = &frac;14; so the Lamé parameters were equal.

gdmcbain:
> I suppose we need some sort of convergence test for this operator.

Day & Romero's analytical eigenmodes for the cuboid might be good. 

Quadratic tetrahedra are O. K., but I think Q2 or S2 might be better if one has a MeshHex.

I couldn't get too far with ex21, just the received mesh and refine(1); the second refinement ran out of memory. Not really enough for a convergence study but I think P2 gives a reasonable answer, unlike P1.

PR #329, opened 2020-02-01
Add MINI element, and convergence test for RT0

kinnala:
Solves the same Poisson equation as the other convergence tests.


kinnala:
Added also MINI element for demonstrating simple Stokes elements. Didn't test with Stokes, but added convergence test for standard H1 case. Could also do some sort of automatized stability testing for the Stokes operator but it's maybe more related to #241 .

tests/test_convergence_hdiv.py
```diff
@@ -0,0 +1,152 @@
...
+    eps = 0.1
+    Hdivbound = 0.04
+    L2bound = 0.01
+
+    def runTest(self):
+
+        @BilinearForm
+        def bilinf_A(sigma, tau, w):
+            from skfem.models.helpers import dot
+            return dot(sigma, tau)
```
gdmcbain: This bilinear form arises in a few places; it might be a candidate for `skfem.models`.

https://github.com/kinnala/scikit-fem/blob/46a75bbf50185cabd3028c1f2c487d8b473ece88/docs/examples/ex03.py#L20-L22

https://github.com/kinnala/scikit-fem/blob/46a75bbf50185cabd3028c1f2c487d8b473ece88/docs/examples/ex21.py#L19-L21

https://github.com/gdmcbain/fenics-tuto-in-skfem/blob/119258655559ab7ac6b563bf75be6923d494d2be/07_navier_stokes_channel/st07_navier_stokes_channel.py#L12-L14

PR #332, opened 2020-02-07
Element composite

kinnala:
I thought I'd announce what I've been doing lately @gdmcbain .

So this is related to #241 . Allows defining, e.g.,
```python
e = ElementVectorH1(ElementTriP2()) * ElementTriP1()
```
and then in the bilinear form definition
```python
@BilinearForm
def bilinf(u, p, v, q, w):
    from skfem.models.helpers import *
    return ddot(u, v) + div(u) * q + div(v) * p
```

In practice this is achieved through `ElementComposite` and the multiplication of elements is just a shorthand for defining that.

I'm currently testing this in a use case related to shell theory. I'll polish it a bit, e.g., allow interpolating of both fields at the same time and make splitting the solution vector easier. Also add some tests.


gdmcbain:
Very nice.  That's a very concise, very clear, notation for the Stokes operator.

I wonder whether the approach taken in #170 to the CG–Uzawa method for the Stokes equation, which is somewhat opposite to the present, is old-fashioned.  In #170 the velocity and pressure (or momentum and continuity) blocks are treated very separately as in the paper by Glowinski & Pironneau, though that is from forty years ago.  I think other modern packages making use of PETSc for mixed finite element problems use something called [‘FIELDSPLIT’](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/PC/PCFIELDSPLIT.html) which forms the equations all together and then internally works out how to separate them.  It does seem easy enough in scikit-fem to keep track of where the blocks are in a block matrix; e.g. as exploited in the fluid mechanics examples from [ex18](https://kinnala.github.io/scikit-fem-docs/examples/ex26.html) through to [ex29](https://kinnala.github.io/scikit-fem-docs/examples/ex29.html).

One feature of the FEniCS/Firedrake DSL to avoid here is their two 'split' operators or methods.  These are a regular source of confusion and bug reports.  I don't foresee it being such an issue here as our vectors are just plain `numpy.ndarray` not wrappings of something else (PETSc vectors, I think).

kinnala:
This is again becoming a huge PR. I'll finish a test that I'm writing, add `ElementComposite.doflocs` and add entries to CHANGELOG. Then I promise to stop.

kinnala:
Solving lid-driven cavity in 2D looks now like this:
```python
        m = MeshTri()
        m.refine(4)
        e = ElementVectorH1(ElementTriP2()) * ElementTriP1()

        m.define_boundary('up', lambda x: x[1] == 1.)
        m.define_boundary('rest', lambda x: x[1] != 1.)

        basis = InteriorBasis(m, e)

        @BilinearForm
        def bilinf(u, p, v, q, w):
            from skfem.helpers import grad, ddot, div
            return ddot(grad(u), grad(v)) - div(u) * q - div(v) * p - 1e-2 * p * q

        S = asm(bilinf, basis)

        D = basis.boundary_dofs(skip=['u^2'])
        x = basis.zeros()
        x[D['up'].all('u^1^1')] = .1

        x = solve(*condense(S, basis.zeros(), x=x, D=D))

        (u, u_basis), (p, p_basis) = basis.split(x)

        from skfem.visuals.matplotlib import plot, show

        plot(p_basis, p, Nrefs=3, colorbar=True)
```

kinnala:
Moreover, splitting `S` into blocks `A`, `B` and `C` could be done via
```python
ix_u, ix_p = basis.split_indices()
A = S[ix_u].T[ix_u].T
```
and so on.

kinnala:
Had to do some refactoring to allow `ElementComposite` to deduce the number of edges/facets without having an explicit `Mesh` instance. Now elements have `Element.mesh_type` attribute.

kinnala:
Added
- `ElementComposite` for using many elements in one bilinear form
- `ElementQuadS2`, quadratic Serendipity element
- `Basis.boundary_dofs` for finding degree-of-freedom indices
- `Mesh.define_boundary` for defining named boundaries
- `Basis.split` for splitting multicomponent solutions

Deprecated
- `Basis.get_dofs` in favor of `Basis.boundary_dofs`

kinnala:
Note: API of `Basis.get_dofs` was becoming too complicated to comprehend.

kinnala:
I'm merging this so I can more easily base some future work on it. If you still have comments, we can revert etc.

PR #334, opened 2020-02-14
Change reference domain

kinnala:
Started work on changing quad and hex reference domains from `(-1, 1)^d` to `(0 ,1)^d`. Shouldn't take too long.


kinnala:
Everything should be now there.

PR #337, opened 2020-02-18
allow internal boundaries from meshio

### LINKED ISSUE #336: internal boundaries from meshio not in bndfacets
I was porting some old one-dimensional code to scikit-fem and found that whereas previously I had identified an internal vertex facet as a `Physical Point` in the Gmsh GEO, this wasn't picked up by `skfem.io.meshio.from_meshio`.

Roughly the code is like:
```python

import meshio
import skfem
from skfem.io import from_meshio

geom = Geometry()
points = []
lines = []
for x in range(3):
    points.append(geom.add_point([x, 0., 0.]))
    if len(points) > 1:
        lines.append(geom.add_line(*points[-2:]))                     
geom.add_physical(points[0], 'start')
geom.add_physical(points[1], 'middle')
geom.add_physical(points[2], 'end')
geom.add_physical(lines, 'domain')
mesh = from_meshio(generate_mesh(geom, dim=1))
``` 
Then `mesh.boundaries` is missing `'middle'`:
```python
{'start': array([0]), 'end': array([2])}
```

The reason is simple:

https://github.com/kinnala/scikit-fem/blob/61424c55508954e71302c78dc5e6637883f68a27/skfem/io/meshio.py#L101

`index` doesn't include `'middle'` because it's not in `bndfacets`.

I haven't decided whether the current behaviour is better or it should be changed to permit the definition of internal facets.  Here I was just going to use it as a probe point at which to simply sample the solution without having to interpolate.  I wonder whether the latter might be interesting for other applications (interfacial sources? domain decomposition into subdomains?).

gdmcbain:
Fixes #336.

This is an easy fix, though I haven't thought through all the consequences.  The immediately motivating problem #336 isn't urgent as it is easy to interpolate instead if there are qualms about this change.


kinnala:
So this causes `m.boundaries` to have interface facets also?

kinnala:
Just checked that `MeshTri._uniform_refine`, `MeshQuad._uniform_refine` should properly preserve interfaces. I think other mesh types don't support these yet. It might work with `MeshLine` if we're lucky since the order of indexing is preserved, we should modify the test `test_mesh.RefinePreserveSubsets` accordingly.

gdmcbain:
Abandoning to reimplement after #338 from meshio>=4.0.4.

Issue #321, opened 2020-01-23
meshio 4 cells


gdmcbain:
it looks as though the next release of meshio, meshio 4, will involve more than the transposition of `cell_data` keys #301; nschloe/meshio#629 replaces the old `cells` `Dict[str, ndarray]` and `cell_data` `Dict[str, [Dict[str, ndarray]]]` with `List[Tuple[str, ndarray]]` and `Dict[str, Tuple[str, ndarray]]`, respectively, the `str` in the `Tuple`s being the cell-type (`'triangle'`, &c.) and the `str` in the `cell_data.keys` being the names of the data (`'potential'`, &c.).  This has arrived in meshio master with nschloe/meshio#631.

Thus this subsumes #301.

I propose to extend PR #302 to adapt to this so that scikit-fem is ready when the next meshio is released.

gdmcbain:
nschloe/meshio#699 will ease this.  `meshio.Mesh` now has a new attribute `.cell_sets` and a more convenient property `.cell_sets_dict` which at least for the domain elements is a lot closer to `skfem.Mesh.subdomains`, being a dict keyed like (in ex17) `'convection'`, `'wire'` and `'insulation'` with values dicts from (meshio) cell-type ('line' or 'quad') to the `numpy.ndarray` which should match that in `subdomains`.

`skfem.Mesh.boundaries` is different of course since Gmsh doesn't have a concept of facets, unfortunately; these still have to be inferred by scikit-fem from the lower-dimensional cells ('line' in ex17).

gdmcbain:
[meshio 4](https://github.com/nschloe/meshio/releases/tag/4.0.0) was released yesterday and is available from [PyPI](https://pypi.org/project/meshio/).

gdmcbain:
With meshio 4.0.1 and pygmsh 6.0.4, examples like ex12 raise
```
NameError: free variable 'meshio_type' referenced before assignment in enclosing scope
```
This is because it is expected that `meshio_type` will have been set in 

https://github.com/kinnala/scikit-fem/blob/61424c55508954e71302c78dc5e6637883f68a27/skfem/io/meshio.py#L34-L42

which fails because `type(m.cells) is list` and `type(m.cells[0]) is meshio.Cells` whereas `type(k) is str`.

gdmcbain:
[...]

PR #338, opened 2020-02-19
from meshio >=4.0.4

gdmcbain:
This is for #321, attempting to also work for old meshio and Gmsh too.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ setup(
         'Topic :: Scientific/Engineering :: Mathematics',
     ],
     packages=find_packages(exclude=['tests']),
-    install_requires=['numpy', 'scipy', 'meshio'],
+    install_requires=['numpy', 'scipy', 'meshio>=4.0.4'],
     extras_require={
         'full': ['matplotlib'],
     },
--- a/skfem/io/meshio.py
+++ b/skfem/io/meshio.py
@@ -31,8 +31,7 @@ def from_meshio(m, force_mesh_type=None):
 
     """
 
-    # Is this is meshio >= 4 and Gmsh MSH >=4.1?
-    cells = m.cells_dict if hasattr(m, 'cells_dict') else m.cells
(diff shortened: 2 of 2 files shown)


gdmcbain:
From #302 

> Maybe we should also add to setup.py the explicit requirement of meshio>4 if that will be the case?

I think so, at least eventually, but for the moment I'll try to accommodate new & old.

gdmcbain:
The error is upstream in pygmsh; its release is behind its master against which this passes the tests.

gdmcbain:
Ah good, nschloe/pygmshw#313 overnight has given us pygmsh 6.0.6; this should pass now.

gdmcbain:
> I think so, at least eventually, but for the moment I'll try to accommodate new & old.

I abandoned this; meshio >=4.0.4 is easy to install via either pip or [conda](https://anaconda.org/conda-forge/meshio).

gdmcbain:
This does still support MSH 2.2, not least because of `docs/examples/beams.msh` used in ex21 and `docs/examples/{square,box}.msh` used in `docs/tutorials/walkthrough.rst` and `tests/test_mesh.py`.

I think the new `meshio.Mesh.cell_sets_dict` property introduced in meshio 4 is much nicer for scikit-fem; subdomains are now quite direct and boundaries also much easier.  However, meshio 4 only defines the `meshio.Mesh.cell_sets` attribute on reading Abaqus and Gmsh's MSH 4.1.  The latter is important as that's what's generated by default by Gmsh 4 and thus pygmsh as used in various examples.  Abaqus might be useful for porting examples from Calculix, but I haven't tried that yet.  I think there's still progress being made on `meshio.read`.  There is also a `meshio.Mesh.int_data_to_sets` method which I had hoped might make MSH 2.2 as directly readable as MSH 4.1, but it isn't quite yet.

kinnala:
Looks good. Will merge after the tests run.

PR #340, opened 2020-02-25
internal boundaries

### LINKED ISSUE #336: internal boundaries from meshio not in bndfacets

gdmcbain:
Fixes #336 for MSH 4.1 or MSH 2.2.

This is atop #338 which makes reading MSH 4.1 much easier.

### DIFF
--- a/skfem/io/meshio.py
+++ b/skfem/io/meshio.py
@@ -76,7 +76,6 @@ def from_meshio(m, force_mesh_type=None):
                     return key
             return None
 
-        bndfacets = mtmp.boundary_facets()
         if m.cell_sets:         # MSH 4.1
             subdomains = {k: v[meshio_type]
                           for k, v in m.cell_sets_dict.items()
@@ -87,7 +86,7 @@ def from_meshio(m, force_mesh_type=None):
                       if bnd_type in v}
             boundaries = {k: np.array([i for i, f in
                                        enumerate(map(tuple, mtmp.facets.T))
-                                       if f in v and i in bndfacets])
+                                       if f in v])
                           for k, v in facets.items()}
         else: # MSH 2.2?
             elements_tag = m.cell_data_dict['gmsh:physical'][meshio_type]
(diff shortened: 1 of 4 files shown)


gdmcbain:
- [ ] [`define_boundary`](https://github.com/kinnala/scikit-fem/blob/f116b1ac001ba83023a37665b97079f0633607e0/skfem/mesh/mesh.py#L405-L420)
- [ ] [`get_dofs`](https://github.com/kinnala/scikit-fem/blob/f116b1ac001ba83023a37665b97079f0633607e0/skfem/assembly/basis/basis.py#L195-L228)
- [ ] [`boundary_dofs`](https://github.com/kinnala/scikit-fem/blob/f116b1ac001ba83023a37665b97079f0633607e0/skfem/assembly/basis/basis.py#L170-L193)
- [ ] [`refine`](https://github.com/kinnala/scikit-fem/blob/f116b1ac001ba83023a37665b97079f0633607e0/skfem/mesh/mesh.py#L111) (with `arg=None` for uniform refinement)
  - [ ] `MeshTri`
  - [ ] `MeshQuad`

gdmcbain:
Covered MeshTri.refine by splitting the four refines into 4 = 1 + 3 and inserting the `define_boundary`.

https://github.com/kinnala/scikit-fem/blob/d6803356b312de2a1230e067a9628b557dc98da7/tests/test_basis.py#L13-L17

gdmcbain:
Ah, I see that idea was already used in test_mesh.RefinePreserveSubsets.  Good.

PR #342, opened 2020-02-27
block-preconditioned Stokes example

gdmcbain:
I've been working on a better Stokes solver than the Uzawa method of ex30, that better scales to reasonably fine three-dimensional problems.  I've run this one with up to 1 095 031 Taylor–Hood tetrahedra which it solved in 331 s.  The example here uses a coarser mesh, giving 41 301 tetrahedra and solving in 22 s.

For three dimensions, it uses an ellipsoid, generated in pygmsh, since there's a known exact solution; I'll add a comparison with that in due course.

The method is based on [ch. 4 of Elman, Silvester, & Wathen (2014)](https://www.oxfordscholarship.com/view/10.1093/acprof:oso/9780199678792.001.0001/acprof-9780199678792-chapter-5) and uses pyamg.smoothed_aggregation_solver.aspreconditioner for the velocity.

I haven't yet thought of a good way to automatically generate a figure from the script, so it's just saving the pressure and (nodal) velocity to VTK for rendinging in ParaView.

ex32


gdmcbain:
This doesn't use the new ElementComposite yet, as it's mostly a small step from the Uzawa scheme of ex30 and none of the previous Stokes examples (exx 18, 24, 27, 29, 30) use it (yet).

kinnala:
Cool. I think it's fine to not use `ElementComposite`.

I like to think of it as a syntactic tweak which might or might not make sense depending on if it simplifies the bilinear forms.

gdmcbain:
I found that 
```shell
pip install pyamgcl
```
didn't work under Windows whereas
```shell
pip install pyamg
```
does so I revised the example so that it can use either.  (Although I can't easily check pyamgcl at the moment.)  I'm not particularly comfortable under Windows, so I might be missing something simple, but then again it's important that examples work easily for new users.

Although there's a lot to multigrid and algebraic multigrid and so probably lots of differences between these two packages, for the purposes of this example (like ex09 and others), they're not important.  The point is that any decent implementation of AMG will lift the solver to the next level in terms of the size of problem that can be solved on modest hardware in a reasonable amount of time, even with (as here) the default options.  I'm sure that still better performance can be obtained by investigating these options, but again that's not needed in these examples.

gdmcbain:
Oh.  I wrote a test for the L2 error in the pressure (relative to the known analytic solution) but then realized that it's not going to work if neither pyamgcl nor pyamg are installed.  

Should one of these be installed in .travis.yml?

gdmcbain:
It would be nice to script an image, but I haven't looked into vtkplotter yet. #361  Apart from that, this is pretty much done, I think.

PR #344, opened 2020-02-27
Rest of the doflocs

### LINKED ISSUE #224: Add doflocs to all Element subclasses
Remaining work from #223 .

kinnala:
Fixes #224 .


tests/test_basis.py
```diff
@@ -55,3 +55,51 @@ def bilinf(u, p, v, q, w):
...
+        ElementTriRT0(),
+        ElementTriP2(),
+        ElementTriArgyris(),
+    ]
+
+    def runTest(self):
+        """Test that global and local doflocs match in ref. domain."""
+
+        for e in self.elems:
+            self.assertEqual(
```
gdmcbain: I wonder whether it would be worth making the test both clearer and more rigorous using [`numpy.testing.assert_allclose`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) or similar.

gdmcbain: Cancellation in the sum of positive and negative errors seems like a possible risk here; e.g. if two doflocs got swapped somehow.

kinnala: Fair enough. Seems that the test is actually failing. Thanks for realizing this.

kinnala: Seems that this test cannot work as such. Unintuitively, the basis functions in reference domains are not mapped one-to-one because of how `Mesh.t2f` is created via `np.unique` and how the global DOF numbering is subsequently created using `Mesh.t2f`.

Changing this would be a fairly large scale change (most of the elements, almost all of the mesh classes, some refinement functions) and it seems to me that there is not any other added benefit than this test would be possible as such.

Basically the numbering of global DOF's is arbitrary and is created in `Mesh._build_dofnum`. Currently this numbering logic simply does not reproduce the numbering that we're expecting for the reference triangle.

I have to think if this would be beneficial in the future or not. For now, I'm inventing another test.

kinnala: I figured out a way to preserve the order of facets but this seemed to broke some assumptions regarding how normal vectors are being mapped from the reference domain.

I'm not going to look into it any further and we should just make notes to ourselves that the global DOF numbering cannot be predicted from the local DOF numbering in any case.

kinnala: I'm now doing some hard thinking on how these things should be tested since there is obviously a possibility of introducing mistakes if the doflocs order does not match with the basis function order.

kinnala: This might be actually a really tough problem in general, because not all elements are nodal (so that `TestNodality` would make sense, e.g., `ElementQuadP` that are based on integrated Legendre polynomials). Do these elements even need doflocs?

kinnala: Earlier I was thinking that doflocs are needed to find dofs on facets, but now that we have `Basis.boundary_dofs` which does it based on the topological entities, and we can first use `Mesh.facets_satisfying` to find the correct entities, I think we don't need those.

kinnala: I now put corresponding `doflocs` rows to `np.nan` if it doesn't make sense to define a location for the particular DOF.

kinnala:
I'll push some additional commits to this soon.

kinnala:
Now I realized that actually DOF location doesn't make too much sense even for RT0 because the DOF is actually an integral over the edge and not flux at some specific point. This just happens to work here because linear flux can be integrated with one integration point in the middle of the edge.

I suppose in these finite elements, not defined by point evaluation functions, you should always resort to projecting the given data to the finite element basis.

kinnala:
I'll leave them there for now and think later what to do about it.

gdmcbain:
I wonder whether that sometimes the ‘DOF location doesn't make too much sense’ is related to the [quotation](https://github.com/kinnala/scikit-fem/issues/224#issuecomment-569540786) from M. Knepley in #224; I'd like to read a fuller exposition of that point of view some day.

Issue #209, opened 2019-06-22
Create example and test for ElementHcurl


kinnala:
At the same time fix subdomain assembly.

PR #346, opened 2020-02-28
Make type of derivative returned by Element explicit

kinnala:
I was working on #209 and realized that there isn't no way of detecting whether `Element.gbasis` already returns the curl of the vector field or some other derivative.

This PR makes it explicit and, hence, easier to detect in `skfem.helpers` what to do when user asks e.g. `div(u)` or `curl(u)` and so on. What do you think @gdmcbain ?


kinnala:
Going to add here an example how this benefits in electromagnetism.

kinnala:
field

I heard visualizing these is really tricky. Well...

gdmcbain:
Sigh. Yes. In the mid-1970s Graham de Vahl Davis (who passed away in December) wrote a seminal paper on simulating convection in three-dimensional cavities—like #342 but in a cuboid. He warned that three-dimensional solution might be demanding but the visualization took ten times as long again. Plus ça change, plus c'est la même chose

Envoyé depuis ProtonMail mobile

-------- Message d'origine --------

gdmcbain:
> This PR makes it explicit and, hence, easier to detect in skfem.helpers what to do when user asks e.g. div(u) or curl(u) and so on. What do you think @gdmcbain ?

Yes, very good idea.

gdmcbain:
A couple more thoughts on the visualization:

As [mentioned](https://github.com/kinnala/scikit-fem/issues/286#issuecomment-568728904) in #286 #287, I've been experimenting with other Pythonic backends for visualization.  Nothing to report yet, but eventually one of these might help.

Is *E* solenoidal here?  (I guess this is true only if *f* is, which I haven't checked.)  If so, integral curves are often better visualizations that vector-fields.  This is certainly true in two dimensions (the stream-function), but also in three.  Sometimes, e.g. in cases with some symmetry, an auxiliary problem can be solved to give a potential function in terms of which the integral curves can be plotted as level-sets; the stream-functions in two-dimensional and axially symmetric problems are obvious examples.  But at worst the integral curves of a vector field *E* can be computed in parameterized form *x* = *x* ( _t_) directly by solving the first-order ordinary differential equation d *x*  = *E* d _t_.  ParaView &c. can do this, although it would be nice to have the backend work with the full finite element representation of the vector field *E* rather than just a P1 approximation of it via its nodal dofs.

gdmcbain:
The Stokes equations can be written a bit like this, since there's the vector identity lap *u* = grad div *u* &minus; curl curl *u* so the momentum equation becomes curl curl *u* + grad p = *f* and the curl of that is curl curl curl *u* = curl *f*.  This is used in 'poloidal–toroidal’ decompositions of divergence-free vector fields in spherical geometries; i.e. putting *u* = curl curl (P *r*) + curl (T *r*) for scalars P and T and *r* = _x_ *i* + _y_ *j* + _z_ *k*.

gdmcbain:
In the sphere (special case of the ellipsoid in #342), the buoyancy force **f** = _x_ **j** is purely toroidal (well, no poloidal part, there is a ‘scaloidal’ part that contributes to the pressure gradient), so the velocity field is purely toroidal too (toroidal **f** → toroidal **u**, poloidal **f** → poloidal **u**), so **u** = curl (*T* **r**) and the toroidal Stokes equation is curl⁴ (*T* **r**) = curl (*x* **j**).  This is derived at

* http://gdmcbain.freeshell.org/papers/thesis/node132.html

Ha, I see though that I didn't plot the velocity field, just the force-lines and the pressure in the plane of dihedral symmetry *z* = 0.

In the general ellipsoidal case, the *z*-component of velocity vanishes everywhere so there is actually a stream-function in each plane of constant *z*.  An example with that property might be easier to plot.

PR #348, opened 2020-03-05
Combine MortarPair and MortarBasis to MappingMortar

kinnala:
Simplifies slightly by reusing `FacetBasis` although I had to do the ugly side effect via `MappingMortar.side`.


kinnala:
Alright, changed also ex33 to use a slightly coarser mesh.

Issue #23, opened 2018-07-30
More elements


kinnala:
More elements would be nice, e.g.,
 
 - [x] Simplical elements up to p=3
 - [x] Second-order hexahedron
 - [x] Curved elements through MappingIsoparametric
 - [x] MINI

gdmcbain:
I've got some Python code for one-dimensional Hermite elements of arbitrary order which are handy for fourth order equations like Euler–Bernoulli beam and Orr–Sommerfeld 

* Mamou, M., & M. Khalid (2004). [Finite element solution of the Orr–Sommerfeld equation using high precision Hermite elements: plane Poiseuille flow](http://dx.doi.org/10.1002/fld.661). *International Journal for Numerical Methods in Fluids* 44(7):721–735 

but also very accurate for smooth second order systems.  I haven't used them since discovering scikit-fem, but my plan is to rewrite them to fit in if possible, certainly before using them again.  I saw that `class Element` has a class variable `nodal_dofs` which was encouraging.

I was unable to do that in UFL at the time because it explicitly excluded one-dimensional Hermite elements.  Looking that up now though, I see that that was reversed earlier this month!  [‘Hermite makes sense in one dimension’](https://bitbucket.org/fenics-project/ufl/commits/6b09f30d34e7fb7ab95ab6920a3ff742e46b074c)

kinnala:
I can later give a short explanation how GlobalBasis uses the Element objects.

Hermite is not H1 conforming so the reference mappings in ElementH1 do not work directly. I think the derivative DOFs need a scaling of 1/h if a reference element is used.

Another approach is to assemble the element globally. Both approaches are viable

gdmcbain:
I've forgotten what H1 conforming means (but I'll look it up) and yes there was a row vector
```python
np.tile(h ** np.arange(self.dofs_per_node), 2)
```
that arose in a number of places.  

It hadn't occurred to me to ‘assemble the element globally’; I'll have to think about that too.

kinnala:
I'm sorry, don't worry about it, it's just math jargon. H1 elements typically have only continuity whereas H2 elements have also continuity of the derivative (which is the idea in Hermite, I suppose?). 

The reference element approach is totally fine also. I'll come back to this in the evening or tomorrow. It's 8.30 AM and I need a cup of coffee.

gdmcbain:
[...]

PR #349, opened 2020-03-08
ElementGlobal, more generic ElementH2

### LINKED ISSUE #39: extend ElementH2 to 1D and 3D
This should be quite easily done. I cannot currently figure any use case for 3D but at least in 1D you could create Hermite elements this way.

- [x] cubic Hermite 1d
- [x] BFS in 2D
- [ ] BFS in 3D for Cahn-Hilliard

kinnala:
This is related to #39 and #23 .

It already replaces the existing `ElementH2`.

My personal need is calculating higher order derivatives.

Fixes #39 .


kinnala:
I'm now looking into Bogner-Fox-Schmidt.

kinnala:
Alright, it seems to be working fine. Noticed one issue in how boundary conditions are set in test_convergence_h2, going to fix that later.

Issue #158, opened 2019-04-02
serialization of meshes and solutions


gdmcbain:
In working on combining a Navier–Stokes solution #145 with an advection–diffusion problem #152, I thought to first solve for the flow, then save it.  Then in a second script, for the advection–diffusion, load the mesh and velocity and solve for the temperature.

The motivation is that, depending on fineness of the mesh, the flow takes a few minutes to solve whereas the advection–diffusion should take much less and I might be making lots of little changes to the second script and don't want to have to rerun the first script every time.

The two hurdles I see are:
1. Just calling `save` doesn't save the `boundaries` attribute.   This attribute doesn't fit into either the `point_data` or `cell_data` optional arguments as I think these are supposed to have lengths matching `.shape[1]` of `.p` or `.t`, respectively.  Thus it's not obvious how to get meshio to save this, despite its various output formats.  It could be stored in the `field_data` of a `meshio.Mesh`, but I don't know whether this would be stored in an ouput mesh file, or whether it would subsequently be reread.  Maybe construct the (_n_ &minus; 1)-dimensional elements corresponding to the boundary facets and pack these into the `cells` variable passed to the `meshio.Mesh` constructor inside `skfem.Mesh.save`?  Along with the appropriate `cell_data`.
1. The velocity is typically solved as P2 if using Taylor–Hood elements, but mesh typically has `'triangle'` cells (in two dimensions), which only have three nodes.  It's not obvious how to get meshio to save a P2 solution on a mesh of simple triangles or tetrahedra.  Should the facet-midpoint nodes and then the six-node triangles be constructed?  (Which gets back to finding dof locations #84).

For the moment, it'll probably do to locally pickle the boundaries attribute and `numpy.save` the solution array, but serialization of meshes and solutions is likely to be something that's going to be wanted eventually, as more problems get more complicated and solution times grow.

gdmcbain:
For meshes created with  `classmethod from_meshio`, the `external` attribute should have everything required for resaving the mesh.  I didn't know about that.

kinnala:
You can now serialize meshes via `Mesh.to_dict` and then dump those to json:
```python
import json
json.dumps(m.to_dict)
```

PR #350, opened 2020-03-14
Move save to io

### LINKED ISSUE #291: Move Mesh.save logic to a separate module

kinnala:
Fixes #291

### DIFF
--- a/skfem/io/meshio.py
+++ b/skfem/io/meshio.py
@@ -134,21 +134,10 @@ def from_file(filename):
     return from_meshio(meshio.read(filename))
 
 
-def to_file(mesh, filename, point_data=None, cell_data=None, **kwargs):
+def to_meshio(mesh, point_data=None):
+    cells = {TYPE_MESH_MAPPING[type(mesh)]: mesh.t.T}
+    return meshio.Mesh(mesh.p.T, cells, point_data)
 
-    meshio_type = TYPE_MESH_MAPPING[type(mesh)]
 
-    if point_data is not None:
-        if not isinstance(point_data, dict):
-            raise ValueError("point_data should be "
-                             "a dictionary of ndarrays.")
-
-    if cell_data is not None:
-        if not isinstance(cell_data, dict):
-            raise ValueError("cell_data should be "
-                             "a dictionary of ndarrays.")
-        cell_data = {meshio_type: cell_data}
-
-    cells = {meshio_type: mesh.t.T}
-    mesh = meshio.Mesh(mesh.p.T, cells, point_data, cell_data)
-    meshio.write(filename, mesh, **kwargs)
+def to_file(mesh, filename, point_data=None, **kwargs):
+    meshio.write(filename, to_meshio(mesh, point_data), **kwargs)
--- a/skfem/io/meshio.py
+++ b/skfem/io/meshio.py
@@ -17,6 +17,8 @@ MESH_TYPE_MAPPING = OrderedDict([
     ('line', skfem.MeshLine),
 ])
 
+TYPE_MESH_MAPPING = {v: k for k, v in MESH_TYPE_MAPPING.items()}
+
 
 def from_meshio(m, force_mesh_type=None):
     """Convert meshio mesh into :class:`skfem.mesh.Mesh`.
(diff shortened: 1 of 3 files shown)


kinnala:
Also fixes bug in skfem.utils.

skfem/io/meshio.py
```diff
@@ -130,3 +132,23 @@ def find_tagname(tag):
...
+            raise ValueError("point_data should be "
+                             "a dictionary of ndarrays.")
+
+    if cell_data is not None:
+        if not isinstance(cell_data, dict):
+            raise ValueError("cell_data should be "
+                             "a dictionary of ndarrays.")
+        cell_data = {meshio_type: cell_data}
+
+    cells = {meshio_type: mesh.t.T}
```
gdmcbain: The `meshio.Mesh.cells` attribute has changed in meshio 4.  It's now `List[CellBlock]` where 
```python
CellBlock = collections.namedtuple("CellBlock", ["type", "data"])
```

https://github.com/nschloe/meshio/blob/add5d452f853cca3f2cb4a9151b297dda03442ce/meshio/_mesh.py#L5

and `type: str` ('line', 'triangle', …) and `data: ndarray` (`ndim == 2`, dtype some kind of int, with one cell per column).

This old type (`Dict[str, ndarray]`) is still tolerated and converted in the initializer but deprecated in comments.

https://github.com/nschloe/meshio/blob/add5d452f853cca3f2cb4a9151b297dda03442ce/meshio/_mesh.py#L22-L35

I think the up-to-date thing here would be
```python
cells = [(meshio_type, mesh.t.T)]
```

gdmcbain: The situation in meshio 4 with `meshio.Mesh.cell_data` is similar, except that it's not converted in the constructor; anything truthy is just accepted.

https://github.com/nschloe/meshio/blob/add5d452f853cca3f2cb4a9151b297dda03442ce/meshio/_mesh.py#L37

Assuming
```python
cells = [(meshio.type, mesh.t.T)]
```
that means something like
```python
cell_data = {k: [(meshio_type, v)] for k, v in cell_data.items()}
```

So far there are no examples of `cell_data` (or `**kwargs`) in scikit-fem/docs/examples, only point_data.

kinnala: I'll add a test for `cell_data` at least.

gdmcbain: We _could_ also start saving `skfem.Mesh.boundaries` as `meshio.Mesh.cell_sets`, although it's not yet used by `meshio.write` for many formats. Anyway, that needn't hold up this PR.

kinnala: I'll see if I have time to try it out. So I should be able to save and load named boundaries?

gdmcbain: Oops, sorry, I meant `skfem.Mesh.subdomains`, not `boundaries`; the latter are more complicated again as they involve facets rather than cells.  meshio doesn't support facets. nschloe/meshio#729

But even with `subdomains` it depends on one's interpretation of ‘should’.  I think that the intent of the newish attribute `meshio.Mesh.cell_sets` is very similar to `skfem.Mesh.subdomains`; see 

https://github.com/kinnala/scikit-fem/blob/ada7b469bd7ce5eab58677cf8a073d67e454b41c/skfem/io/meshio.py#L79-L82

for how we can read `skfem.Mesh.subdomains` from `meshio.Mesh.cell_sets_dict`, which is a property that restructures `meshio.Mesh.cell_sets`.  We should be able to invert this with something like
```python
cell_sets = {k: [(meshio_type, v)] for k, v in mesh.boundaries.items()}
```
however, as [mentioned](https://github.com/kinnala/scikit-fem/pull/350#discussion_r393537114), I  don't think `meshio.write` actually writes the `meshio.Mesh.cell_sets` to very many formats, maybe only Abaqus, which I haven't tried.  Not many mesh file formats could support the functionality of cell_sets; MSH 4.1 could but meshio doesn't really write that format.

But hopefully in the future meshio will write `cell_sets` to more formats (maybe XDMF, Exodus, …?) and that will give more options for serialization #158.

Without support from `meshio.write`, there might still be use in defining `cell_sets` in terms of `boundaries` if something else is to be done with the `meshio.Mesh`.  There might be external libraries that recognize and process `meshio.Mesh` (smoothers? hex-splitters? I don't know).  This would necessitate factoring a function `skfem.io.meshio.to_meshio` out of `to_file` in the same way that `from_meshio` was factored out of `from_file`.

Sorry, these are vague optimistic speculations on the future which aren't really related to this PR.

kinnala:
Decided to remove cell_data altogether. We can open separate issues and pull requests for the other issues.

PR #360, opened 2020-04-05
Release 1.0.0

### LINKED ISSUE #194: Roadmap to 1.0.0

### LINKED ISSUE #326: specify quadrature
One of the things arising from the recent investigation into the poor behaviour of elastic modal analysis with `ElementTetP1` #75 was the idea of reduced quadrature; e.g. in discussing `ElementHex1`, Petyt (1990, [_Introduction to Finite Element Vibration Analysis_](https://doi.org/10.1017%2fCBO9780511624292), Cambridge University Press, §5.8) recommends a rule with one abscissa at the centroid.  Reduced quadrature or ‘selective integration’ is also discussed by

* Benzley, S. E., Perry, E., Merkley, K., Clark, B. & Sjaardema, G. (1995). [A comparison of all hexagonal and all tetrahedral finite element meshes for elastic and elasto-plastic analysis](https://imr.sandia.gov/papers/abstracts/Be15.html). In T. Tautges (ed.), In Proceedings, 4th International Meshing Roundtable (pp. 179–191).

although their practical recommendation is for increasing the order of the element as done here in #322.

Currently I don't see an easy way to experiment with quadrature in scikit-fem.  I think though that enabling this would be simple generalization of the constructor for [`skfem.assembly.Basis`](
https://github.com/kinnala/scikit-fem/blob/89f726808f50bf9773f1e2c7d43248ccbd0714a0/skfem/assembly/basis/basis.py#L26).

https://github.com/kinnala/scikit-fem/blob/89f726808f50bf9773f1e2c7d43248ccbd0714a0/skfem/assembly/basis/basis.py#L26

or maybe

https://github.com/kinnala/scikit-fem/blob/89f726808f50bf9773f1e2c7d43248ccbd0714a0/skfem/assembly/basis/interior_basis.py#L35-L40

(and similarly for [`FacetBasis`](https://github.com/kinnala/scikit-fem/blob/89f726808f50bf9773f1e2c7d43248ccbd0714a0/skfem/assembly/basis/facet_basis.py#L36-L42)) so that either one specifies:
* the actual quadrature rule, i.e. `X` and `W` which become attributes of the `InteriorBasis`, or
* `intorder: int` as now, and the rule is computed from [`skfem.quadrature.get_quadrature`](https://github.com/kinnala/scikit-fem/blob/89f726808f50bf9773f1e2c7d43248ccbd0714a0/skfem/quadrature.py#L8), or
* nothing, also as now, and `intorder` is computed from [`self.elem.maxdeg`](https://github.com/kinnala/scikit-fem/blob/89f726808f50bf9773f1e2c7d43248ccbd0714a0/skfem/assembly/basis/basis.py#L52).

A possible further modification is that the type of the rule returned by `get_quadrature` is `Tuple[ndarray, ndarray]` from which it is unclear which is weights and which abscissae.  Perhaps a quadrature rule might be represented as in [`quadpy.nsimplex._helpers.NSimplexScheme`](https://github.com/nschloe/quadpy/blob/ef2fe0720c08a9a563c8f9ba678e0e44e190c89d/quadpy/nsimplex/_helpers.py#L8-L9) as an object with attributes `weights` and `points`, or more simply as a `NamedTuple` (or `dataclass`) with similarly named attributes; given Python's duck-typing, reference from within `skfem.assembly` only to attributes `weights` and `points` might enable direct use of one of the many schemes from quadpy while only importing it at the top-level user-script in which the basis is instantiated.

I don't think that this is urgent—I don't have any immediate plans to use it myself—but it seems like an easy addition and something that should be present in a general finite element library.  I think it's mostly a matter of deciding on the desired syntax.

kinnala:
I'd like to proceed with 1.0.0 release even if the documentation is still lacking behind a bit, i.e. still advocating the use of some older features from which improved versions are available. We can then begin fixing those but I think there is no hurry.

Is there anything you'd like us to address before that @gdmcbain?

I think we could now consider adding `DeprecationWarning` to the old style forms. What do you think?

I have now successfully written numerical examples to one unpublished scientific article using the new style forms and are pretty happy with them.

Edit: Fixes #326 . Fixes #194 .


kinnala:
We can also not deprecate the old style forms yet and simply first migrate all documentation to use them before adding `DeprecationWarning`.

kinnala:
It's probably some download step failing. I'll restart the build.

gdmcbain:
Hmm, I typed a review here, I thought, but now I don't see it; unfamiliarity with working under Windows...

Meanwhile, I'm starting to look into adapting the examples to use the new-style forms, both to familiarize myself and to facilitate the step to 1.0.0.

kinnala:
I added `DeprecationWarning` to this implicit casting procedure: https://github.com/kinnala/scikit-fem/pull/360/commits/5a7a9567f32845d23bb7eb7dc7813561a6df3482 .

I wanted to simplify what gets passed to `asm(..., w=w)` as `w`. Now it can be a bunch of different things which I find sometimes very confusing.

@gdmcbain Do you see any obvious downsides to this? It should be easy to fix all examples.

gdmcbain:
Yes, I see. I wonder whether we can just use `**kwargs` somehow.

kinnala:
I'm not sure how it'd go because we have to take two parameters for different `Basis` objects so that one can have different `Basis` for the solution and the test function (e.g. various Petrov-Galerkin methods).

kinnala:
Unless `Basis` and `DiscreteField` were interchangeable somehow. Let me look into it.

gdmcbain:
> I'm not sure how it'd go because we have to take two parameters for different Basis objects so that one can have different Basis for the solution and the test function (e.g. various Petrov-Galerkin methods).

Hmm, yes, I see, or if projecting between bases, e.g. between overlapping meshes.  I suppose it must happen that there can be coefficients (like thermal conductivity, whatever) on each side.  I mean in the abstract a bilinear form might look like (something with test functions including operators and coefficients, something with trial functions including operators and coefficients) and the two bases in general (except in the strict Galerkin case) differ.  How is that handled at the moment?  I think I've only had to put coefficients on the side of the test functions, as in `dot(grad(v), k * grad(u))` for thermal conduction with nonuniform thermal conductivity _k_.

kinnala:
Now fixes #326 . I needed that in a research problem.

kinnala:
Decided to remove the deprecation warning from `Basis.get_dofs` because it was heavily used downstream and currently easy to support.

In future we should make a more thorough plan for handling DOF numbering and "named boundaries". Hopefully in 2.0.0.

Issue #357, opened 2020-04-03
Matrix Transforms, Displacements, and Resultant Forces


mbatesole:
Hello All, 

I don't have an issue with scikit-fem so much as I have a general question of use. 

I'd like to take a beam, constrain one end then displace the other end, but not just in translation-- I'd like to perform a 4x4 matrix transform on that end consisting of both rotation and translation then determine the resultant force. 

Every package I've investigated so far only implements the translation without the rotation: Solidworks, Fusion 360, CalculiX, etc. 

Is this something that scikit-fem can do? Would you be so kind as to show me how? 

Thanks!

gdmcbain:
So the points _x_ on some face will be transformed to _x'_ = A _x_ + _b_ for matrix _A_ and vector _b_ independent of _x_, as in https://en.wikipedia.org/wiki/Affine_transformation#Augmented_matrix ?  Then the displacement is _u_ (_x_) = _x'_ &minus; _x_ = (A &minus; I) _x_ + _b_.

I think we can do this with a combination of
* the linear elasticity example, [ex11](https://kinnala.github.io/scikit-fem-docs/examples/ex11.html), except that that has trivially A = I, 

https://github.com/kinnala/scikit-fem/blob/de2f4acba96abd6acb716197330e24b7190618b4/docs/examples/ex11.py#L13-L19

and

* a way to have the essential boundary conditions depend on the coordinates, as in the parabolic velocity profile applied at the inlet to the backward facing step example, [ex24](https://kinnala.github.io/scikit-fem-docs/examples/ex24.html):

https://github.com/kinnala/scikit-fem/blob/de2f4acba96abd6acb716197330e24b7190618b4/docs/examples/ex24.py#L76-L81

gdmcbain:
I had a go at this.  There was a bit more mucking about with `numpy.einsum` than I'd like, but I may have overlooked an easier way.  It's basically ex11 but with a nonzero rotation matrix A about the _z_-axis.

[ex_affine.zip](https://github.com/kinnala/scikit-fem/files/4425178/ex_affine.zip)

![ex_affine](https://user-images.githubusercontent.com/1588947/78327245-26051e00-75c8-11ea-8e52-4408215e50cb.png)

In the course of this I did notice some simplifications in selecting the degrees of freedom for the essential boundary condition in ex24 (& ex27) which I'll push shortly (branch `essential-dofs`).

gdmcbain:
[...]

PR #362, opened 2020-04-07
new forms in examples

gdmcbain:
&hellip;as discussed in #360 .

### DIFF
--- a/docs/examples/ex27.py
+++ b/docs/examples/ex27.py
@@ -35,7 +35,8 @@ def acceleration(v, w):
 
     """
     u = w['w']
-    return np.einsum('j...,ij...,i...', u, grad(u), v )
+    return np.einsum('j...,ij...,i...', u, grad(u), v)
+
 
 @bilinear_form
 def acceleration_jacobian(u, du, v, dv, w):
--- a/docs/examples/ex27.py
+++ b/docs/examples/ex27.py
@@ -1,4 +1,5 @@
 from skfem import *
+from skfem.helpers import grad, dot
 from skfem.models.poisson import vector_laplace, laplace
 from skfem.models.general import divergence, rot
 from skfem.io import from_meshio
@@ -19,8 +20,8 @@ from pygmsh.built_in import Geometry
 from pacopy import natural
 
 
-@linear_form
-def acceleration(v, dv, w):
+@LinearForm
+def acceleration(v, w):
     """Compute the vector (v, u . grad u) for given velocity u
 
     passed in via w after having been interpolated onto its quadrature
@@ -33,9 +34,8 @@ def acceleration(v, dv, w):
         u_j u_{i,j} v_i.
 
     """
-    u, du = w.w, w.dw
-    return sum(np.einsum('j...,ij...->i...', u, du) * v)
-
+    u = w['w']
+    return np.einsum('j...,ij...,i...', u, grad(u), v )
 
 @bilinear_form
 def acceleration_jacobian(u, du, v, dv, w):
--- a/docs/examples/ex27.py
+++ b/docs/examples/ex27.py
@@ -38,8 +38,8 @@ def acceleration(v, w):
     return np.einsum('j...,ij...,i...', u, grad(u), v)
 
 
-@bilinear_form
-def acceleration_jacobian(u, du, v, dv, w):
+@BilinearForm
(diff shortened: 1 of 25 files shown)


docs/examples/ex27.py
```diff
@@ -51,8 +52,8 @@ def acceleration_jacobian(u, du, v, dv, w):
        (w_j du_{i,j} + u_j dw_{i,j}) v_i
 
     """
-    return sum((np.einsum('j...,ij...->i...', w.w, du)
-                + np.einsum('j...,ij...->i...', u, w.dw)) * v)
+    return dot(np.einsum('j...,ij...->i...', w['w'], grad(u))
+               + np.einsum('j...,ij...->i...', u, grad(w['w'])), v)
```
kinnala, line 56: Do you think there should be some special shorthand in `skfem.helpers` for matrix-vector products? I think `dot` in Fenics works for that purpose but I don't like mixing "dot" products and "matrix-vector" or "matrix-matrix" products, somehow it doesn't make sense to me because I always like to think "dot product" as an inner product.

E.g. `skfem.helpers.mul` could be a new helper which detects the dimension of the inputs and then acts similarly as `dot` in Fenics (https://fenicsproject.org/qa/595/difference-inner-dot/).

Currently you can also write this as
```
ddot(prod(w, v), grad(u)) + ddot(prod(u, v), grad(w))
```
but it might not be too obvious what is the meaning.

gdmcbain: Actually here I like the einsum for its clarity. FEniCS gets this wrong in my opinion; it's a real muddle in UFL working out which way around the gradient of a vector field is.

I think it's O. K. to use dot between vectors and ddot between second order tensors as there's no chance of confusion there, but otherwise I'm never quite sure whether it's the first or last or inside indices that are being summed over. 

Also generally maybe it's clearer keeping the test function v separate?

gdmcbain:
One place I thought might do with some syntactic sugar is getting at the coordinates, e.g. in the Legendre example, or the parabolic Poiseuille inlet profile, or the 4×4 matrix transformation in #357. I think I had to write something like `w['x'].values[1]` to get at _y_. Is there a better way that I've overlooked?

docs/examples/ex27.py
kinnala: Sure I think it's also fine.

docs/examples/ex28.py
```diff
@@ -71,16 +72,16 @@ def make_mesh(halfheight: float,  # mm
...
+def conduction(u, v, w):
+    return dot(w['w'] * grad(u), grad(v))
 
 
-@bilinear_form
-def advection(u, du, v, dv, w):
-    _, y = w.x
+@BilinearForm
+def advection(u, v, w):
+    y = w['x'].value[1]
```
kinnala, line 82: Can you write `_, y = w.x` ?

`w` is of type `FormDict(dict)` for which `__getattr__` has been overloaded as follows:
```python
    def __getattr__(self, attr):
        return self[attr].value
```

docs/examples/ex29.py
```diff
@@ -13,20 +14,14 @@
 U = Polynomial([1, 0, -1])      # base-flow profile
 
 
-@bilinear_form
-def divergence(u, du, v, dv, w):
-    """Can't use models.general.divergence in one dimension"""
-    return v * du[0]
+@BilinearForm
+def base_velocity(u, v, w):
+    return v * U(w['x'].value[0]) * u
```
kinnala, line 19: Also here it's possible to write `w.x[0]`

docs/examples/ex29.py
```diff
@@ -13,20 +14,14 @@
...
-def base_velocity(u, du, v, dv, w):
-    return v * U(w.x[0]) * u
-
-
-@bilinear_form
-def base_shear(u, du, v, dv, w):
-    return v * U.deriv()(w.x[0]) * u
+@BilinearForm
+def base_shear(u, v, w):
+    return v * U.deriv()(w['x'].value[0]) * u
```
kinnala, line 24: And here also `w.x[0]`.

docs/examples/ex28.py
gdmcbain: Oh, yes, that's much better.  I thought that I had tried that and it hadn't worked but it does.  That's fine.  Thanks.

PR #364, opened 2020-04-09
Use kwargs in asm for additional fields

kinnala:
As proposed in https://github.com/kinnala/scikit-fem/pull/360#issuecomment-611193927 .

After all it seemed to work fine. So the principle is now that
```python
y = basis.interpolate(x)  # type: DiscreteField
```
and this gets passed to `asm` as keyword arguments:
```python
asm(bilinf, basis, test=y)
```
This assumes e.g.
```python
@BilinearForm
def bilinf(u, v, w):
    return u * v * w.test
```

Makes sense?

### DIFF
--- a/skfem/assembly/basis/basis.py
+++ b/skfem/assembly/basis/basis.py
@@ -285,17 +285,18 @@ class Basis:
 
         Returns
         -------
-        Dict[str, DiscreteField]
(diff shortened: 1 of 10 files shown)


kinnala:
I'll add some additional tests for confirming that it works with multiple fields later today.

PR #367, opened 2020-04-27
convert from get_dofs to find_dofs in examples

gdmcbain:
I noticed that deprecating get_dofs was dropped from 1.0.0 because it's still too much in use so thought I'd make a start on replacing it.


gdmcbain:
There are a couple left (exx 02, 11) but they're examples that I'm less familiar with.

Issue #371, opened 2020-05-18
ElementTetMini


gdmcbain:
ElementTetMini should avoid #370 on `MeshTet`, for example for Stokes problems.

`ElementTriMini` was added in #329.  It looks as though `ElementTetMini` should be a straightforward generalization.  I did also initially think that about `ElementHex2` #327 for which it wasn't the case; care has to be taken with the sequence of the additional degrees there, but for a MINI there is only one additional degree of freedom, the interior dof.

gdmcbain:
I've just noticed that the introduction of the MINI element in
*  Arnold, D. N., Brezzi, F. & Fortin, M. (1984). A Stable Finite Element for the Stokes Equations. _Calcolo,_ 21, 337--344. doi: 10.1007/bf02576171 

and its treatment in

*  Braess, D. (2001). _Finite Elements._ Cambridge University Press, &sect;3.7

are both only two-dimensional, for triangular elements.

gdmcbain:
Neither the triangular nor tetrahedral MINI element seem to appear in the [Periodic Table of the Finite Elements](http://femtable.org).

kinnala:
I think the proof should be equivalent for the stability on triangles and
tetrahedra. Trying it out should be quick though.

I mean,

(div v, p) = -(v, grad p) = 0

and picking v to be the elementwise bubble, you get p = const. in each
element.

Thus, if p is the difference of two solutions, they are the same up to a
constant.

gdmcbain:
There is a _d_-dimensional treatment of &lsquo;The ℙ<sub>1</sub>-bubble/ℙ<sub>1</sub> finite element&rsquo; in &sect;4.2.4 of
*  Ern, A., Guermond, J.-L. (2004). _Theory and Practice of Finite Elements._ New York: Springer

gdmcbain:
How is the coeffiicient 27 in the interior_dof of ElementTriMini arrived at?  I assume the element will work regardless of what value is chosen, but that there's a sensible convention out there.

kinnala:
I think it was simply normalized so that the value is one at the element midpoint. But actually it doesn't matter because this basis doesn't form a partition of unity.

gdmcbain:
Ah, O. K.; even though `doflocs[-1] == [np.nan] * dim`, that seems reasonable.

I calculate 256 for the tet.

Issue #372, opened 2020-05-18
MINI example


gdmcbain:
The (triangular) MINI element `ElementTriMini` was introduced in #329 but without an example other than a couple of tests.  Should it be used in an example?

`ElementTetMini` is proposed in #371.

kinnala:
A simple example would be to study the null space of the discrete Stokes operator using a few different elements, e.g.

1. P_1 - P_1 (unstable)
2. P_1 + bubble - P_1 (stable)
3. P_2 + P_1 (stable)

For unstable elements, the null space (eigenfunction corresponding to the eigenvalue 0) should contain a checkerboard function which can be visualized.

There is also some quadrilateral element which is stable only if the mesh is perfectly aligned to a grid but cannot remember which pair it is.

gdmcbain:
I like this idea a lot.

I've just acquired a copy of

* Boffi, D., Brezzi, F. & Fortin, M. (2008). Finite elements for the Stokes problem. In Boffi & Gastaldi (eds) _Mixed Finite Elements, Compatibility Conditions, and Applications_, Springer. ([doi:10.1007/978-3-540-78319-0_2](https://doi.org/10.1007%2f978-3-540-78319-0_2))

which I'll have a look through and then put a script together.

PR #373, opened 2020-05-18
ElementTetMini

gdmcbain:
This is for #371.

Basically it's copied from `ElementTriMini`.

The cofficient on the bubble remains unclear.

The element has been added to `tests/test_elements.py` but not yet `test/test_convergence.py`.

### DIFF
--- a/skfem/__init__.py
+++ b/skfem/__init__.py
@@ -53,6 +53,7 @@ __all__ = [
     'ElementTetP1',
     'ElementTetP2',
     'ElementTetRT0',
+    'ElementTetMini',
     'ElementTriDG',
     'ElementTriP0',
     'ElementTriP1',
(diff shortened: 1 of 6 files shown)


gdmcbain:
The convergence test, copied from that for `ElementTriMini` fails as the element is of degree 1 + dim = 4 and so needs `get_quadrature_tet(2 * 4)` but it looks like only up to 4 is available.  (How does the test work for `ElementTriMini` work then?  Doesn't that need order 2 * (1 + dim) = 6?)

gdmcbain:
Ah, there's reduced integration in `tests.test_convergence.ConvergenceTetP1` and its descendents.

gdmcbain:
I think the only thing missing now is a sensible choice of coefficient for the bubble.

And maybe an example. #372

gdmcbain:
ex32 still works with the simple replacement `ElementTetP2` -> `ElementTetMini`.

kinnala:
Fantastic. I'll add a changelog entry. Otherwise it looks fine.

PR #374, opened 2020-05-18
Fix boundary edges

### LINKED ISSUE #370: Incorrect edges returned by `Mesh3D.boundary_edges`
I've been working on applying the block-diagonally preconditioned MINRES+AMG Stokes solver of #342 & ex32 to other MeshTet geometries and found some odd errors.  It seems to have nothing to do with the operators but something to do with `.find_dofs` when there are several Gmsh elementary entities in a single Gmsh `Physical Surface`.  The situation here is three Physical Surfaces: inlet, outlet, and wall, with uniform positive pressure applied at the inlet, zero pressure at the outlet, and zero velocity on the wall.  Typically the inlet and outlet are single elementary surfaces and the walls include everything else.  Example: a cube with inlet and outlet on an opposing pair of faces and the other four being walls.

I generate a tetrahedral mesh with marked boundaries in Gmsh, marking the boundaries in two different ways: first with all four walls combined, second with a separate Physical Surface for each.
```
SetFactory("OpenCASCADE");

Box(1) = {0, 0, 0, 1, 1, 1};

Physical Volume("fluid", 13) = {1};

Physical Surface("inlet") = {1};
Physical Surface("outlet") = {2};

Physical Surface("wall") = {3, 4, 5, 6};

Physical Surface("wall3") = {3};
Physical Surface("wall4") = {4};
Physical Surface("wall5") = {5};
Physical Surface("wall6") = {6};
```

I load this with 
```
import numpy as np
import skfem

basis = skfem.InteriorBasis(skfem.MeshTet.load("duct.msh"), skfem.ElementTetP2())

walls = {"separate": np.concatenate([b.all() for k, b in basis.find_dofs().items()
                                     if k.startswith("wall") and k[len("wall") :]]),
         "combined": np.concatenate([b.all() for k, b in basis.find_dofs().items()
                                     if k.endswith("wall")])}
```

Now I was thinking that `walls["separate"]` and `walls["combined"]` would be equal arrays of degrees of freedom on the walls but apparently not.

```python
walls["diff"] = np.setdiff1d(walls["combined"], walls["separate"])
print('In "separate" but not "combined"',  np.setdiff1d(walls["separate"], walls["combined"]))  # empty
print('In "combined" but not "separate"', walls["diff"])  # several
print('In "combined", not "separate", and not edge', np.setdiff1d(walls["diff"], basis.edge_dofs))  # empty
```

The "combined" approach has more degrees of freedom on the walls, with the extra ones all being in `basis.edge_dofs`.  (I haven't located them more precisely yet.)  The result is typically a singular matrix.  The "separated" approach seems to lead to a correct matrix

I haven't ascertained whether the discrepancy is in Gmsh, meshio, or `skfem.MeshTet.load` (or in my understanding of `basis.find_dofs`).

kinnala:
I think this fixes #370 . I need to add one more test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### [Unreleased]
 
+#### Added
+- `ElementTetMini`, MINI-element for tetrahedral mesh
+
+#### Fixed
+- `Mesh3d.boundary_edges` incorrectly returned all edges where both nodes are on
+the boundary
+
 ### [1.0.0] - 2020-04-22
 
 #### Added
--- a/docs/examples/ex04.py
+++ b/docs/examples/ex04.py
@@ -1,4 +1,4 @@
-"""Contact problem.
+r"""Contact problem.
 
 Mortar methods allow setting interface conditions on non-matching meshes.
 They are useful also when solving variational inequalities such as
--- a/docs/examples/ex05.py
+++ b/docs/examples/ex05.py
@@ -1,4 +1,4 @@
-"""Integral condition.
+r"""Integral condition.
 
 This short example demonstrates the implementation of an integral boundary
  condition
--- a/docs/examples/ex06.py
+++ b/docs/examples/ex06.py
@@ -1,4 +1,4 @@
-"""High-order plotting.
+r"""High-order plotting.
 
 This simple example demonstrates the usage and visualisation of biquadratic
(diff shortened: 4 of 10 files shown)


kinnala:
Wow. Apparently one convergence test with TetP2 fails (still H1 convergence rate is ~2 so it should be fine) and example 33 solution slightly differs. Good that we get this fixed.

PR #375, opened 2020-05-20
new-style LinearForm for rot

gdmcbain:
I found another old-style `linear_form`, in `skfem.models.general`.

Is there a better way to write this?

### DIFF
--- a/skfem/models/general.py
+++ b/skfem/models/general.py
@@ -1,7 +1,9 @@
 """Bilinear and linear forms too general to put into a specific model."""
 
-from skfem.assembly import BilinearForm, linear_form
-from .helpers import div
+from skfem.assembly import BilinearForm, LinearForm
+from .helpers import div, grad
+
+import numpy as np
 
 
 @BilinearForm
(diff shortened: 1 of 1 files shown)


kinnala:
I cannot figure out anything cleaner. I suppose we could define also `rot_grad` in `skfem.helpers` and use it here if there is further need for such operator. But this is fine for now.

gdmcbain:
Thank you.

Since posting this (and while working on #372) I have been reading about [finite element exterior calculus](https://doi.org/10.1017%2fs0962492906210018) and [compatible finite element spaces](https://doi.org/10.1093%2fclimsys%2fdzw005).  It will take a while to digest but I'm getting the idea that I should choose the finite element spaces for stream-function, velocity, and vorticity more thoughtfully.

I also wondered whether, if the finite element spaces were chosen properly, it would be possible to solve for the stream-function _&psi;_ directly from its definition **rot** _&psi;_ = **u** rather than the Poisson equation rot **rot** _&psi;_ = &minus;lap _&psi;_ = rot **u**.

scikit-fem is an excellent tool for exploring these concepts.

Issue #327, opened 2020-01-29
ElementHex2


gdmcbain:
The next element to be added under #23 is probably the 27-node quadratic three-dimensional `ElementHex2`.

Quadratic hexahedral elements were recommended over `ElementHex1` by Benzley et al (1995), as consulted in #75.

> The use of quadratic displacement formulated finite elements significantly improve the
performance of the tetrahedral as well as the hexahedral elements.

`ElementHex2` will also be required for Taylor–Hood element for Stokes or Navier–Stokes problems on a `MeshHex`.

gdmcbain:
I guess the  definition begins like:
```python
class ElementHex2(ElementH1):
    nodal_dofs = 1
    facet_dofs = 1
    interior_dofs = 1
    edge_dofs = 1
    dim = 3
    maxdeg = 6
    dofnames = ['u', 'u', 'u', 'u']
```
but what is the rule for how the dofs are ordered within each group (8 nodal, 6 facets, 1 interior, 12 edges)?  I assume the nodal dofs are the same as `ElementHex1` but how are the facets and edges numbered?  (And is the order indeed nodal, faces, interior, edges?)

kinnala:
Tricky question because it depends on how the mesh connectivity is defined in `MeshHex._build_mappings()`. E.g. by looking at this definition of edges:
```
        self.edges = np.sort(np.hstack((
            self.t[[0, 1]],
            self.t[[0, 2]],
            self.t[[0, 3]],
            self.t[[1, 4]],
            self.t[[1, 5]],
            self.t[[2, 4]],
            self.t[[2, 6]],
            self.t[[3, 5]],
            self.t[[3, 6]],
            self.t[[4, 7]],
            self.t[[5, 7]],
            self.t[[6, 7]],
        )), axis=0)
```
Basically you would need to make sure that the first edge basis function would be the one which gets value 1 between the nodes numbered 0 and 1 (rows 0 and 1 in `mesh.p[:, mesh.t]`), the second basis function gets value 1 between the nodes numbered 0 and 2, etc.

I hope it makes sense. If it doesn't, ask more info and I'll provide.

kinnala:
[...]

Issue #83, opened 2018-10-31
Renamed: Edges in GlobalBasis._expand_facets


kinnala:
This is remaining work from #38. Loading the Submesh edges from external formats is not done yet.

kinnala:
This functionality is not anymore within Mesh.load. Instead, facets are expanded to edges in GlobalBasis._expand_facets. There is now a crude implementation which does not work correctly in some edge cases (e.g. tetrahedron which has all nodes on the boundary.)

This should be fixed by first creating f2e type of mapping in MeshTet and then using that in GlobalBasis._expand_facets.

PR #377, opened 2020-05-23
20 node quadratic hexahedral element

kinnala:
This begins the work on #327 .

Edit: Includes also a feature for merging two meshes via `__add__`.


kinnala:
The convergence rate seems to be lower than I expect; I'm not sure what causes this. Need to investigate a bit further.

kinnala:
Added several new tests but couldn't find the culprit: still the new element converges as O(h^1/2) in H^1 norm and O(h) in L^2 norm which isn't O(h^2) and O(h^3) as I'd expect.

kinnala:
Figured out that there could still be a hidden bug in the facet->edge expansion. Added a test in the latest commit.

kinnala:
Yes! Figured it out. The facet->edge expansion that was fixed in #83 was working only in case of tetrahedral elements.

kinnala:
Now we have the first quadratic hexahedral element. Do we still need the 27 node element? Maybe for Stokes?

skfem/element/element_hex/element_hex_s2.py
```diff
@@ -0,0 +1,87 @@
+import numpy as np
+
+from ..element_h1 import ElementH1
+from ...mesh.mesh3d import MeshHex
+
+
+class ElementHexS2(ElementH1):
```
gdmcbain, line 7: `ElementQuadS2` is derived from `ElementQuad2` in which case for consistency `ElementHexS2` should be derived from `ElementHex2`, but perhaps it shouldn't have?   I've sometimes read that object-oriented inheritance is supposed to mean 'is a' but don't know how valid or important that is.  Anyway I think the tree should be consistent between quadrilaterals and hexahedra here.

skfem/element/element_hex/element_hex_s2.py
```diff
@@ -0,0 +1,87 @@
...
+    nodal_dofs = 1
+    edge_dofs = 1
+    dim = 3
+    maxdeg = 3
```
gdmcbain, line 11: Is maxdeg = 3?  I would have thought it was 4, like `ElementHex2`; the eight `nodal_dofs` are quartic.

gdmcbain:
> Do we still need the 27 node element? Maybe for Stokes?

Well we probably want it for completeness anyway and yes for Stokes, I think.  `ElementHex2` is often recommended for Stokes (e.g. Ern & Guermond 2004, _Theory and Practice of Finite Elements_, figure. 4.4), but I don't think I've seen serendipity elements used for incompressible fluids.

skfem/element/element_hex/element_hex_s2.py
kinnala: Increased this to 4. Also decreased to 3 in `ElementQuadS2`.

skfem/element/element_hex/element_hex_s2.py
kinnala: I changed how `ElementQuadS2` gets defined.

kinnala:
> Well we probably want it for completeness anyway and yes for Stokes, I think.

Sure. That can be my next follow up task.

PR #379, opened 2020-05-31
Update assembly/__init__.py: fix flake8 errors

ahojukka5:
Fixes the following flake8 errors in `skfem/assembly/__init__.py`:

- '.basis.Basis' imported but unused
- '.basis.InteriorBasis' imported but unused
- '.basis.FacetBasis' imported but unused
- '.dofs.Dofs' imported but unused
- '.form.BilinearForm' imported but unused
- '.form.LinearForm' imported but unused
- '.form.Functional' imported but unused
- '.form.bilinear_form' imported but unused
- '.form.linear_form' imported but unused
- '.form.functional' imported but unused

### DIFF
--- a/skfem/assembly/__init__.py
+++ b/skfem/assembly/__init__.py
@@ -62,3 +62,16 @@ def asm(form: Form,
 
     """
     return form.assemble(*args, **kwargs)
+
+
+__all__ = [
+    "Basis",
+    "InteriorBasis",
+    "FacetBasis",
+    "Dofs",
+    "BilinearForm",
+    "LinearForm",
+    "Functional",
+    "bilinear_form",
+    "linear_form",
+    "functional"]


gdmcbain:
Oh.  I guess that that makes sense.  Thank you.  There are lots of similar F401 in other subdirectories; I count 145 in total. We may as well fix them all at once if this is the correct thing to do.

gdmcbain:
I proposed them over in ahojukka5/scikit-fem#1.  (I wonder whether had I a better integrated developers' environment this kind of task could be automated.)

ahojukka5:
This is one way to fix them. Another way would be to add `# noqa` comment, see [this](https://stackoverflow.com/questions/31079047/python-pep8-class-in-init-imported-but-not-used) SO question.

kinnala:
I can definitely see the point of allowing a wildcard import from the top-level submodules (i.e. `skfem.assembly`, `skfem.element`, etc.) as proposed here for `skfem.assembly` but I'm not sure if it's needed recursively in all submodules as was done in ahojukka5#1. I'll merge this for now and pick some commits from ahojukka5#1 that are obviously for the better.

Note: I'm currently not running automatically any linters so I easily miss these kind of things. I'm using the same setup in larger legacy codebases where it causes lots of visual noise. We could consider linting this project in Github Actions but let's discuss that in another issue.

PR #380, opened 2020-06-01
miscellaneous simpler F401

kinnala:
Picking some commits from https://github.com/gdmcbain/scikit-fem/tree/379


kinnala:
Apparently that removed also stuff that's still needed. I'm fixing those.

kinnala:
I think the rest of the unused imports should be fixed with linter exceptions (i.e. `#noqa`).

PR #382, opened 2020-06-01
Add flake8 linting to CI and fix errors

kinnala:


kinnala:
This is a bit ugly as now all commits would be marked with the red cross. I was hoping to make the check initially optional.

kinnala:
Added ignore to E741 which makes flake8 angry about the index set variable `I` naming.

kinnala:
I decided to grind all of them at one go so now it will show green.

PR #384, opened 2020-06-02
Add test for ElementLineHermite

ahojukka5:
Add test to make sure that exceptions are thrown when calling `gdof` with index > 3. After merging this PR, file `element_line_hermite.py` is having 100 % test coverage.


ahojukka5:
Internally, `gdof` is calling `index_error` which raises `ValueError`. Is this intentional?

kinnala:
> Internally, `gdof` is calling `index_error` which raises `ValueError`. Is this intentional?

Maybe `IndexError` would be more appropriate? 

> exception ValueError
> Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

> exception IndexError
> Raised when a sequence subscript is out of range.

kinnala:
Note: Apparently the flake8 check I added using Github Actions yesterday is not running on pull requests. I need to look what's causing it.

kinnala:
Now it runs but there seems to be an issue with the Github Action I used: https://github.com/samuelmeuli/lint-action/issues/13

Issue #385, opened 2020-06-02
preserve subdomains & boundaries in MeshQuad._splitquads


gdmcbain:
In ex17, e.g., 

```python
mesh.subdomains.keys()
mesh.boundaries.keys()
```
are `['wire', 'insulation']` and `['convection']`, respectively, but `mesh._splitquads()` has `None` for  `subdomains` and `boundaries`.

I think there's a natural definition for the induced subdomains and boundaries; the former should be much like the preservation of cell data but inversely.

Issue #104, opened 2018-11-24
Subdomains & boundaries in refinement


kinnala:
The subdomains and boundaries are currently invalid if one refines the mesh. Create a method to Mesh which removes them on refine and override in MeshTri etc. to create new indexing

Boundaries:

- [x] MeshTri uniform
- [ ] MeshTri adaptive
- [x] MeshQuad
- [ ] MeshTet
- [ ] MeshHex

gdmcbain:
This would be perfect for the empirical convergence study of #13; thus far I've been remeshing the Gmsh GEO from scratch.

(What's ‘neq‛?)

gdmcbain:
Yes, and this would be even more comvenient in conjunction with the adaptive use of `Mesh.refine` demonstrated in #80.

kinnala:
For now I added a warning that the indexing is invalidated.

Subdomains are simple to maintain in refine, but now that I thought about this more carefully I think that maintaining the named boundaries is very hard using the current data structures. There are many edge cases so you have to be very careful or do something clever.

kinnala:
Uniform refine in MeshTri now should preserve named boundaries (stuff in Mesh.boundaries) in ec297c1f33be6839aaed47ffa854aaaa0eb60d5c. I was overly pessimistic in the previous comment: it was quite straightforward although it required some meditation.

kinnala:
I realized that there is still a bug in refine when called with integer larger than one. This is just a reminder for myself.

PR #387, opened 2020-06-02
preserve subdomains & boundaries in MeshQuad._splitquads

gdmcbain:
For #385


gdmcbain:
I noticed that `MeshQuad._splitquads` isn't covered by a test.  I'll see if I can add one.

kinnala:
This naming, `_splitquads`, seems to imply that it is a private method. Originally it was, as the reason for introducing it was to do plotting of solutions on quadrilateral meshes and it was used internally by `MeshQuad.plot3`.

Currently I can think of many other use cases so I wonder if we should add an alias, `MeshQuad.split_quads`, `MeshQuad.split`, or something. Note that we also have the variant `MeshQuad._splitquads_symmetric` which is supposed to split each quad into four triangles.

gdmcbain:
I see, yes, good point about the private-looking name.

On the two new names, the 'quads' in the first seems redundant but then there are lots of other functions and methods already called split.  And maybe it could be thought to be an inverse of the new `Mesh.__add__ `, so maybe the former.  Or what about `MeshQuad.to_meshtri`?

And for the symmetric variant, I hoping that they can be refactored later.  As I generally have to work with unstructured meshes, I haven't looked at it myself.

gdmcbain:
Oh, I see that `MeshQuad._splitquads_symmetric` introduces new points in the centroids; that'll break the implementation above.  And the tests, which are based on checking that the points supporting the subdomains and boundaries don't change.  Can that be deferred?

kinnala:
Yes, `to_meshtri` sounds reasonable. NumPy/SciPy uses `tolist`, `toarray`, etc. but I think `tomeshtri` sounds too terse. Also, we already have `to_dict` and `to_json` etc.

We can later add `m.to_meshtri(symmetric=True)` flag.

skfem/mesh/mesh2d/mesh_quad.py
```diff
@@ -244,21 +245,41 @@ def _uniform_refine(self):
 
         self._fix_boundaries(new_facets)
 
-    def _splitquads(self, x=None):
+    def to_trimesh(self, x=None):
```
kinnala, line 248: `s/trimesh/meshtri/`

tests/test_mesh.py
```diff
@@ -181,5 +181,21 @@ def runTest(self):
...
+
+        for s in mesh.subdomains:
+            self.assertEqual(np.setdiff1d(*[m.t.T[m.subdomains[s]]
+                                            for m in [mesh, tri_mesh]]).size,
+                             0)
+
+        for b in mesh.boundaries:
+            np.testing.assert_array_equal(*[m.facets.T[m.boundaries[b]]
+                                            for m in [mesh, tri_mesh]])
+
```
kinnala, line 198: Here is an additional test that should be resolved:
```python
>>> m = MeshQuad()
>>> m.define_boundary('left', lambda x: x[0] == 0)
>>> m.to_trimesh()
Triangular mesh with 4 vertices and 2 elements.
>>> m.to_trimesh().boundaries
{'left': array([1])}
>>> m.refine()
>>> m.to_trimesh()
  File "<stdin>", line 1, in <module>
  File "/src/scikit-fem/skfem/mesh/mesh2d/mesh_quad.py", line 269, in to_trimesh
    for f in self.facets.T[self.boundaries[k]]])
  File "/src/scikit-fem/skfem/mesh/mesh2d/mesh_quad.py", line 269, in <listcomp>
    for f in self.facets.T[self.boundaries[k]]])
StopIteration
```

skfem/mesh/mesh2d/mesh_quad.py
gdmcbain: 6e326be

tests/test_mesh.py
gdmcbain: Ugh.  Is this something to do with #104? That has `MeshQuad` ticked. 

I knew those functional programming people were onto something when they lauded immutability.

gdmcbain: `m.to_meshtri().boundaries['left'] == [6, 1]`, which isn't sorted which was assumed.  I'll sort.  Is it only `refine` that deranges boundaries?  I'll add the test.

It's also assumed that `.facets` is sorted lexicographically.

gdmcbain: I've sorted in _splitquads, but an alternative would be to sort in `Mesh._fix_boundaries`. 

https://github.com/kinnala/scikit-fem/blob/b5571dac8bad50aee2da1ad6a1146f488c49e400/skfem/mesh/mesh.py#L144-L145

gdmcbain: ...instead of `.flatten()`

kinnala: > I knew those functional programming people were onto something when they lauded immutability.

Yeah, it's an old mistake that `refine` does not return a new mesh. Sometimes I ran into an issue where I run `m.refine()` and then the respective `InteriorBasis` breaks.

I wonder if we should do something about it or not.

kinnala: > I've sorted in _splitquads, but an alternative would be to sort in `Mesh._fix_boundaries`.
> 
> https://github.com/kinnala/scikit-fem/blob/b5571dac8bad50aee2da1ad6a1146f488c49e400/skfem/mesh/mesh.py#L144-L145

I cannot currently recall a reason why there couldn't be a sort there.

I'll add an issue as a reminder.

PR #394, opened 2020-06-04
minor edits

gdmcbain:
Some of these are opinionated; e.g. PDEs v. PDE's for the plural.

### DIFF
--- a/paper/paper.bib
+++ b/paper/paper.bib
@@ -115,7 +115,7 @@
 
 @article{mcbain2018,
   title = {Three ways to compute multiport inertance},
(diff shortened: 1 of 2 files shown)


gdmcbain:
I couldn't find the Hannukainen & Juntunen preprint and the URL seems garbled.  Classic BiBTeX problem!

kinnala:
> I couldn't find the Hannukainen & Juntunen preprint and the URL seems garbled. Classic BiBTeX problem!

Yes, I actually contacted one of the authors due to link rot and asked them to upload to arXiv or something which can be cited. I'll send you email.

PR #410, opened 2020-06-17
Create tests/test_element_quad_bfs.py

ahojukka5:
Add a test for checking that ValueError is thrown. Now `element_quad_bfs.py` is 100 % test covered.


kinnala:
I think there is actually a bug in the implementation: to be consistent with the rest of the element classes it should raise the error when all DOF's are exhausted. So basically something like
```python
if i > 16 or i < 0:
    self._index_error()
```
I suppoes now the error won't be reached for large or negative integers.

kinnala:
Closing for inactivity, created new issue #467 on the bug.

Issue #413, opened 2020-06-29
ElementLineMini


gdmcbain:
In the context of the primitive Orr&ndash;Sommerfeld equation (ex29), I started wondering whether there was a one-dimensional MINI element.  I think there is; it's not very different from `ElementLineP2`.   It's got linear nodal dofs and a quadratic bubble. 
```python
    maxdeg = 1 + dim
```
I think it spans the same finite element function space but just with different basis functions.  I've implemented it and it seems to run ex29 just as well.  Would it be worth including for completeness or is it just redundant?  I haven't thought of any possible advantages thus far.

PR #414, opened 2020-06-29
ElementLineMini

gdmcbain:
As discussed in #413.

Is this just a curio?


gdmcbain:
Run ex29 using the MINI element with
```shell
python docs/examples/ex29.py --mini
```

With this, I checked the all-important `max(c['scikit-fem'].real)` (this is the growth-rate of the most unstable mode) and got:
| element | _c_ | 
| ---------- | ----- | 
| Taylor&ndash;Hood | 0.9646422554314331 |
| MINI | 0.9646422554314381 |

Not much difference, as expected since the two elements span the same function space.

kinnala:
Why not?

PR #418, opened 2020-07-09
Merge dofs

kinnala:
Allows combining two objects returned by `get_dofs` or `find_dofs`.

### DIFF
--- a/skfem/assembly/dofs.py
+++ b/skfem/assembly/dofs.py
@@ -66,3 +66,19 @@ class Dofs(NamedTuple):
                           "Returning an empty set of DOF's.")
 
         return output
+
+    @staticmethod
+    def _merge_dicts(d1, d2):
+        keys = set(d1).union(d2)
+        no = np.array([])
+        return dict((k, np.union1d(d1.get(k, no), d2.get(k, no)))
+                    for k in keys)
+
+    def __or__(self, other: Dofs):
+        """For merging two sets of DOF's."""
+        return Dofs(
+            nodal=self._merge_dicts(self.nodal, other.nodal),
+            facet=self._merge_dicts(self.facet, other.facet),
+            edge=self._merge_dicts(self.edge, other.edge),
+            interior=self._merge_dicts(self.interior, other.interior),
+        )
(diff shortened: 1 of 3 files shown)


tests/test_basis.py
```diff
@@ -89,3 +89,16 @@ def runTest(self):
...
+
+    def runTest(self):
+        m = MeshTri()
+        basis = InteriorBasis(m, ElementTriP2())
+        D1 = basis.get_dofs(lambda x: x[0] == 0)
+        D2 = basis.get_dofs(lambda x: x[0] == 1)
+        D3 = basis.get_dofs(lambda x: x[1] == 1)
+        D4 = basis.get_dofs(lambda x: x[1] == 0)
+        assert_allclose(D1 | D2 | D3 | D4,
+                        basis.get_dofs())
```
gdmcbain, line 104: I thought `.get_dofs` was to be deprecated and replaced by `.find_dofs`, as in #367?

skfem/assembly/dofs.py
```diff
@@ -66,3 +66,22 @@ def all(self, key: Union[List[str], str] = None):
                           "Returning an empty set of DOF's.")
 
         return output
+
+    @staticmethod
+    def _merge_dicts(d1, d2):
+        keys = set(d1).union(d2)
+        no = np.array([])
+        return dict((k, np.union1d(d1.get(k, no), d2.get(k, no)))
```
gdmcbain, line 74: We don't need to sort the merged degrees of freedom here do we?  I don't think so, just checking.

kinnala: Do you mean that something else than `union1d` would suffice?

gdmcbain: No, I worried for a moment that leaving the degrees of freedom unsorted here might be a problem.  I don't think it is, just checking.

tests/test_basis.py
kinnala: Yeah I thought so too but now looking at the current implementation it's pretty short and concise. Also `find_dofs` wasn't the ultimate solution as I thought it would be: we still need something like `find_subdomain_dofs` or `get_subdomain_dofs` to find DOF's corresponding to the subdomains. So still further improvements required on this regard.

Eventually I'd like to have something like
```
x, D1 = basis.set_boundary_condition(lambda x: x[0] == 0, 5.0)
x, D2 = basis.set_boundary_condition(lambda x: x[0] == 1, lambda x: x[0] ** 2, x)
...
```
which builds on top of the existing more lower level functionality, whatever that'll be.

gdmcbain: O. K.; interesting, thanks.

PR #419, opened 2020-07-11
Refactor Dofnum into a separate class

kinnala:
Taking some parts of #400 to make the work in smaller steps.

For high-order meshes a global DOF numbering, separate from the DOF numbering related to the finite element basis, is required. Thus, make `_build_dofnum` into a separate class.


kinnala:
Right now my plan is to incorporate this functionality in `Dofs` and filter that for further purposes.

kinnala:
The idea is now the following: when `Basis` is initialized, a `Dofs` object is created with all global DOF numbers. Then when `get_dofs` or `find_dofs` is called, a proxy object to `Dofs` with the type `DofsView` is created. This `DofsView` has all the traditional `nodal`, `edge`, `facet` etc. properties and the method `.all(['u'])`.

This currently does not have the functionality to "skip" `dofnames` as was in the previous implementation of `find_dofs`.

I wonder @gdmcbain if you use this keyword argument `skip` in `find_dofs` anywhere? It was used in one test case only. It could be implemented in `DofsView` so that you could filter also based on `dofnames` in addition to node, facet, etc. indices.

gdmcbain:
No, I don't think so; I've looked at `skip` but don't have it in use. I generally try to keep `docs/examples` up to date in covering all the features that I use.  Thanks for asking.

kinnala:
Technically the removal of `skip` is a backwards-incompatible change and dictates the release of 2.0.0. We could remove old school forms at the same time. What you think?

kinnala:
I'd try to include at last some cool new features before that though so that it's not only removing stuff.

kinnala:
Nevermind, I improved it so that we have backwards-compatibility.

kinnala:
- [ ] Docstrings (for `drop` and `keep` at least)
- [ ] Add `Dofs.get_element_dofs` (a feature request)

gdmcbain:
> We could remove old school forms at the same time. What you think?

Yeah, it would be good to do this eventually and maybe sooner rather than later.  I don't know how much legacy code is out there using the old ones, but for new arrivals looking into the code-base it's probably confusing to have the two.  I don't think the old ones offer any advantages over the new, so nothing's being lost on that count.

PR #420, opened 2020-07-11
Add new example (ex35)

ahhuhtal:
New example solving a simple electromagnetic problem. Problem statement is written in docstring and a simple test case is added.


kinnala:
Uploading the resulting figure here for archival purposes:
ex35_solution

gdmcbain:
Thanks for posting the figure; I was away from the terminal when the example came in and had to imagine it in my head.  This is very nice.  

I wonder whether more generally it might be good to have a gallery of figures from the suite of examples as an index into them, like https://matplotlib.org/gallery (this goes back to #31 #365 and I suppose further discussion should be taken up under the latter rather than here).

Issue #431, opened 2020-07-16
Figures of examples


kinnala:
Figures are stored here for the purpose of linking to them at https://scikit-fem.readthedocs.io/en/latest/

FYI @gdmcbain

kinnala:
ex2:
![ex02_solution](https://user-images.githubusercontent.com/973268/87659951-f50bbc00-c766-11ea-8c0e-7de0e9e83714.png)

kinnala:
ex03:
![Figure_1](https://user-images.githubusercontent.com/973268/87661134-cbec2b00-c768-11ea-81bc-f5455df7cc33.png)

kinnala:
ex03:
![ex04_solution](https://user-images.githubusercontent.com/973268/87661313-1372b700-c769-11ea-89ee-db144986a25a.png)

kinnala:
ex05:
![ex05_solution](https://user-images.githubusercontent.com/973268/87661575-7ebc8900-c769-11ea-8e51-07aeb49797f9.png)

kinnala:
ex06:
![ex06_solution](https://user-images.githubusercontent.com/973268/87661665-aa3f7380-c769-11ea-9677-8ec0ff27184c.png)

kinnala:
ex07:
![Figure_1](https://user-images.githubusercontent.com/973268/87662192-80d31780-c76a-11ea-9291-2d11920bc098.png)

kinnala:
ex08:
![Figure_1](https://user-images.githubusercontent.com/973268/87662432-e0c9be00-c76a-11ea-85b9-711c6b34791e.png)

kinnala:
ex09:
placeholder, waiting for paraview to compile

kinnala:
ex10:
![Figure_1](https://user-images.githubusercontent.com/973268/87663902-1c658780-c76d-11ea-9e00-324a18769ad2.png)

kinnala:
ex11:
![ex11_solution](https://user-images.githubusercontent.com/973268/87685532-31054800-c78c-11ea-9b89-bc41dc0cb80c.png)

kinnala:
ex12:
![Figure_1](https://user-images.githubusercontent.com/973268/87686059-bee13300-c78c-11ea-9693-727f0baf0433.png)

kinnala:
ex13:
[...]

PR #432, opened 2020-07-16
Add a gallery of examples

### LINKED ISSUE #428: Add a gallery of examples
From #426
> The authors claims that more than 30 examples illustrate the use of the library. I did not find any page that summarises those examples. It would be very useful for new users. I mean something similar to what the MFEM developers provide: https://mfem.org/examples/

Also related from https://github.com/kinnala/scikit-fem/pull/420#issuecomment-657289008
> I wonder whether more generally it might be good to have a gallery of figures from the suite of examples as an index into them, like https://matplotlib.org/gallery (this goes back to #31 #365 and I suppose further discussion should be taken up under the latter rather than here).

I propose we add this to Sphinx docs and also add a link to it in README to make it obvious.

kinnala:
Docs will be built at https://scikit-fem.readthedocs.io . Figures will be added to #431 as comments and linked to in the Sphinx doc.

Closes #428 .


kinnala:
12 out of 35 examples done.

kinnala:
Images added for all examples. What remains is to add a small description for all examples 13 onwards. I believe most of them I can just copy paste from the docstring.

PR #433, opened 2020-07-17
Fix for ElementComposite when used with FacetBasis

ahhuhtal:
Trying to initialize a FacetBasis when using an ElementComposite caused an error:
  File "mtl_coax.py", line 89, in <module>
    resistor_end_basis = FacetBasis(mesh, element, facets=mesh.facets_satisfying(lambda x: (x[0] == 1.0)))
  File "/home/antti/.local/lib/python3.8/site-packages/skfem/assembly/basis/facet_basis.py", line 97, in __init__
    self.basis = [self.elem.gbasis(self.mapping, Y, j, self.tind)
  File "/home/antti/.local/lib/python3.8/site-packages/skfem/assembly/basis/facet_basis.py", line 97, in <listcomp>
    self.basis = [self.elem.gbasis(self.mapping, Y, j, self.tind)
TypeError: gbasis() takes 4 positional arguments but 5 were given

In the above we have
mesh = MeshLine(np.linspace(0, 1, 100))
element = ElementLineP1() * ElementLineP1()
resistor_end_basis = FacetBasis(mesh, element, facets=mesh.facets_satisfying(lambda x: (x[0] == 1.0)))

I'm not very experienced with Python, so please check carefully my reasoning and proposed solution :-).

The problem seems to stem from FacetBasis calling gbasis of the Element with tind as a positional argument. The gbasis for other elements accept this as a positional argument, but ElementComposite expects a keyword argument.

My proposed fix is the following:
Change ElementComposite gbasis to accept tind as a positional argument, as is defined in Element, instead of **kwargs
Change FacetBasis to call Element.gbasis with tind as a keyword argument instead of a positional argument, as is done in InteriorBasis


kinnala:
Yes I think this is a correct fix. I'll add a simple test for it and merge.

PR #434, opened 2020-07-17
Add possibility to pass data type for assembly

ahhuhtal:
Bilinear and linear form assembly intializes the data array as np.float64.

If the bilinear or linear forms return complex values, a warning will be displayed and the imaginary part will be discarded.
To retain the wanted complex value, it should be possible to allow the user to specify the proper data type to use when assembling the matrix and vector.

This could be done automatically in the event that the form returns non-conforming data, but I propose the following:
Have assembly take as input an argument "dtype" for the data type, whose value defaults to np.float to allow existing syntax.


gdmcbain:
Interesting.  I guess I was just lucky not to hit this issue in developing the ex29 on the primitive Orr&ndash;Sommerfeld equation #237; it's complex but each bilinear form is real.

skfem/assembly/form/bilinear_form.py
```diff
@@ -42,6 +42,7 @@ class BilinearForm(Form):
     def assemble(self,
                  u: Basis,
                  v: Optional[Basis] = None,
+                 dtype=np.float,
```
gdmcbain, line 45: I wonder whether there's a type-hint for `dtype`.  I note that `isinstance(np.float, type)` so would it make sense to write
```python
dtype: type = np.float
```
?

(I don't know the answer here.)  

So far I don't think anything formal or automatic is being done with the type-hints in scikit-fem (except perhaps in users' IDEs?), but one day that might be useful; e.g. they might be exploited by optimizing compilers.

----

On looking further into this tangential question, I see that it's quite a topic. numpy/numpy#7370 numpy/numpy#16515

ahhuhtal: Unfortunately I'm very new to Python, so I'm not familiar with many things yet :-). Had to google type hinting.

It does seem like a good idea to have dtype be of type 'type'. Though I found that there is a legacy convention, which allows defining the type as a string (for instance dtype='f') for numpy arrays, but I don't think that needs to be supported.

ahhuhtal: Regarding the Orr-Sommerfeld example and complex systems. I was working on multiconductor transmission line computations, with a nonzero damping matrix from a resistive absorbing boundary condition. This results in a complex system, but with all assembled matrixes real. Thus condense and solve work fine with complex matrixes. It wasn't until I added a capacitive boundary condition, which made the damping matrix complex that I arrived to problems.

kinnala:
Does it make sense to use the same form definition for different dtypes?

Anyhow, I'd like to see at least a small test case utilizing this feature so that it won't be forgotten in possible later refactors.

ahhuhtal:
> Does it make sense to use the same form definition for different dtypes?
> 
> Anyhow, I'd like to see at least a small test case utilizing this feature so that it won't be forgotten in possible future refactorings.

I'm not sure what you mean by same form definition for different dtypes? You mean there would be e.g. a separate ComplexBilinearForm? (SesquilinearForm?) I think I'd rather just have the one form. It's your project though, so you have the final say.

kinnala:
> I'm not sure what you mean by same form definition for different dtypes? You mean there would be e.g. a separate ComplexBilinearForm? (SesquilinearForm?)

I was thinking more like parametrizing the decorator, e.g.,
```python
@BilinearForm(dtype=np.complex64)
def bilinf(u, v, w):
    ....
```
but I think the presented approach is also fine.

ahhuhtal:
That does look like a nice approach. I can't think of a situation where the same form would need to be assembled with different data types. Didn't know enough Python to know this was possible.

ahhuhtal:
I had a look at decorator arguments, but I can't figure out a way to retain the existing syntax for defining bilinear forms. In the decorator __init__, I think the callable needs to be the final argument, and there can't be optional arguments before a required argument, I don't think.
Thus we could have
@BilinearForm(dype = np.float)
def form(u,v,w)
but not
@BilinearForm
def form(u,v,w)

kinnala:
Try something like
```python
class Form:

    def __init__(self,
                 form: Callable = None,
                 dtype: type = None):
        self.form = form
        self.dtype = dtype

    def __call__(self, *args):
        if self.form is None:  # decorate
            return type(self)(form=args[0],
                              dtype=np.float64 if self.dtype is None\
                              else self.dtype)
        return self.assemble(self.kernel(*args))

    ...
```
and then use `self.dtype` in `Form.assemble` to initialize `data`.

kinnala:
I'll merge this with the above suggestions and modify the test accordingly.

Issue #426, opened 2020-07-15
Comments on your JOSS paper review


thelfer:
This documents is part of the review of the paper submitted to the Journal of Open Source Software. See <https://github.com/openjournals/joss-reviews/issues/2369>.

The paper describes the `scikit-fem` `python` package which is dedicated to the assembly of the matrices and vectors resulting from the finite element discretization of partial differential equations. More precisely, the library allows the user to focus on the weak form of the differential equation and the choice of the finite element library (several families are provided).

The paper illustrates the library with three examples, grouped in a single figure:

- a linear elastic contact problem.
- a H(curl)-conforming model problem.
- a resolution of the Navier–Stokes equations for different Reynolds numbers using Taylor–Hood element.

# General remarks and questions

- The paper is easy to read. I don't have particular remarks on the quality of english used.
- The paper is very short. I understand that the task is to library is very narrow. However, many things are missing to fully understand the interest and the limits of the library. For example:
  - The library do provide a way to handle Dirichlet boundary conditions by elimination. This shall be at least mentioned. I imagine that the treatment of more complex conditions, such as multiple points constraints in mechanics, have to be treated by hand after the assembly, isn't it ?
  - Following the previous remark, it is not clear how the library could be used in this way. For example, is the mesh connectivity clearly described (how the unknowns are numbered) ?
  - The authors shall describe how the library could be used in non linear cases. In particular, the contact problem given in the first example is indeed non linear.
  - Can non uniform coefficients be handled ? To make my question clearer, would I be able to treat:
    - a linear elastic problem with a non uniform Young Modulus.
    - a heat transfer equation with a non uniform heat source Those non uniform coefficients are dealt with in `UFL` using the notion of functions spaces. Is there anything equivalent in `scikit-fem` ?
- The authors claims that the assembly, while made in pure `python`, is  "fast enough" ? This shall be made more explicit. Do the authors mean that the assembly takes a negligible time compared to the resolution of the resulting linear system ?
- The authors claims that more than 30 examples illustrate the use of the library. I did not find any page that summarises those examples. It would be very useful for new users. I mean something similar to what the `MFEM` developers provide: <https://mfem.org/examples/>
- Would it work in MPI-like processes ? Do you plan any extension of the package in that matter or is it out of its scope ?.

# Specific comments

## Quadrature points

[...]

Issue #423, opened 2020-07-13
DOC: basis functions and weak formulation for those without a background in finite elements


gdmcbain:
> For someone who does not have a background in finite elements, they may not understand what basis functions or weak formulation means. (#422)

gdmcbain:
My initial thoughts are that (at present), scikit-fem is either for:
* those who do have a background in finite elements and know what basis functions and weak formulation mean, or
* for internal use in other packages in which the end-user need not have background in finite elements but doesn't have to look inside the functions in that package.

In the first case, the design idea of scikit-fem is that it adheres closely to the standard fundamental notions of the finite element method as understood and accepted by those with a background in finite elements, e.g. as set forth in the books of Braess, Ern & Guermond, Elman, Silvester, & Wathen, &c., &c., or as in other codes like FreeFEM, libMesh, FEniCS, &c.,, &c.; and expresses them in Python (in modern Python, Pythonically, and 'thinly' in the SciPy software stack).  scikit-fem is deliberately not (at present) innovative at the level of the basic finite element notions, anyone having read one of those books or used one of those codes shouldn't have much trouble seeing what MeshTet, ElementTetP1, InteriorBasis, BilinearForm, LinearForm, &c., are intending.

In the second case, the writers of those parts of the downstream packages need to be in the first case, but the users of those packages might not even know that scikit-fem or finite elements are being used at all.  This is like someone might use [`scipy.integrate.solve_ivp`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html) without understanding the theory & nomenclature of Runge&ndash;Kutta methods (if indeed that's what they use: that I don't actually know, which is just the point).  Actually more I'm thinking of higher-level application-specific packages like the battery-simulator, but I'm guessing that it would be possible to write a drop-in re-implementation of `scipy.integrate.solve_bvp` using scikit-fem without the user having to know about basis functions or weak formulations.

These two use-cases are both important but have to be distinguished from each other.  I think it is possible for a single package to serve both cases without compromise, largely thanks to the design of the Python language and the SciPy stack.

I'd say that (at present) scikit-fem is not for the in-between case of the end-user who wants to set up and solve a general mechanical stress problem or heat transfer problem using finite element analysis but without knowledge of finite elements.

Anyway, these are just my initial thoughts, without rechecking what was written in the submission.  Were they covered there?  Or do they need to be emphasized or added?  I'll recheck the text later.

[...]

PR #444, opened 2020-07-29
Update paper based on the review

kinnala:
Starting to update the paper (@gdmcbain FYI).


kinnala:
- #423 Added a citation to the FEniCS book which I think is a rather good introduction to weak formulations and basis functions given that it's also a freely available reference.
- #426 Added the governing equation of the second example to the caption. (Is this enough?)
- #426 & #428 Changed the wording so that it implies the "over 30 examples" can be now found in the documentation.

kinnala:
- #426 Added references to matplotlib and ParaView.

kinnala:
- Tried addressing many of the comments in #426 by adding another paragraph to "Purpose and prior art".

Any comments or more ideas?

paper/paper.md
```diff
@@ -69,7 +69,7 @@ results from `scikit-fem`, e.g., @mcbain2018, @gustafsson2019, and
...
 Python package for battery modelling [@pybamm].
 
-![(Top left.) A combination of triangular and quadrilateral elements is used to solve the linear elastic contact problem. (Top right.) The lowest order tetrahedral Nédélec element is used to solve the $H(\mathrm{curl})$-conforming model problem $\nabla \times \nabla \times E + E = f$. (Bottom.) The Taylor–Hood element is used to solve the Navier–Stokes flow over a backward-facing step for different Reynolds numbers.\label{fig:examples}](examples.png)
+![(Top left.) A combination of triangular and quadrilateral elements is used to solve the linear elastic contact problem. The figure was created with the help of matplotlib [@hunter2007]. (Top right.) The lowest order tetrahedral Nédélec element is used to solve the $H(\mathrm{curl})$-conforming model problem $\nabla \times \nabla \times E + E = f$. The figure was created with the help of ParaView [@ahrens2005] (Bottom.) The Taylor–Hood element is used to solve the Navier–Stokes flow over a backward-facing step for different Reynolds numbers. The figure was created with the help of matplotlib [@hunter2007].\label{fig:examples}](examples.png)
```
gdmcbain: Rather than duplicating 

> The figure was created with the help of matplotlib [@hunter2007].

it might be worthwhile taking the opportunity to point out that there for two-dimensional problems there is a very convenient wrapping of matplotlib whereas for three-dimensional problems, which is generally more specialized, there is output to VTK, XDMF, &c., via meshio, for subsequent inspection and rendering in, e.g., ParaView.

Let's see, maybe:

```
![(Top left.) A combination of triangular and quadrilateral elements is used to solve the linear elastic contact problem. (Top right.) The lowest order tetrahedral Nédélec element is used to solve the $H(\mathrm{curl})$-conforming model problem $\nabla \times \nabla \times E + E = f$. (Bottom.) The Taylor–Hood element is used to solve the Navier–Stokes flow over a backward-facing step for different Reynolds numbers.   The first and last two-dimensional figures were generated using scikit-fem's convenient flexible wrapping of matplotlib [@hunter2007]; three-dimensional postprocessing is generally more specialized and is usually left to export of VTK, XDMF, \&c., via meshio [@meshio] for subsequent rendering in, e.g., ParaView [@ahrens2005].
```

PR #448, opened 2020-07-30
Splu ex19

gdmcbain:


kinnala:
I suppose now we don't need `scikit-sparse` in the CI job definition?

PR #449, opened 2020-07-30
WIP: lagged ILU preconditioned JFNK for ex10

### LINKED ISSUE #119: scipy.optimize.root as a nonlinear solver

gdmcbain:
I've finally worked out how to use `scipy.optimize.root` for pure (sparse!) Newton iteration (which fixes #119), pure JFNK (superseding #442), and many options for in between.  

As Knoll & Keyes (2004, &sect;3.5 &lsquo;Matrix-free preconditioning approaches&rsquo;) put it:

> There is a continuum of choices ranging from forming no preconditioner to forming the complete Jacobian.

Basically `scipy.optimize.root(method="krylov")` runs the Newton iteration using finite difference approximations to the Jacobian.  Since it solves the Newton iteration equation with a Krylov method, it only needs the action of the Jacobian on a vector (basis vector of the Krylov space) and this is cheaply computed with the finite difference approximation to the Fr&eacute;chet derivative in the required direction.

But `root` also accepts a preconditioner and if that preconditioner were the actual Jacobian, the Krylov iteration would converge in a single step.  In between those two extremes we can approximate the Jacobian in the preconditioner using ILU and we can lag it, only updating every _p_ steps.  Here I take `p = np.inf`, so only the Jacobian of the initial guess is ever assembled and (incompletely) factorized.

This is still quite rough, but I think it works well and hopefully demonstrates some of the principles that I'd been talking about in connexion with hyperelasticity. #439


bhaveshshrimali:
Thanks a lot @gdmcbain ,
Just finished taking a look at the paper and also this. Quickly adapting to the hyperelasticity example, there seems to be a trade-off when trying to improve the inner preconditioner. The overall convergence improves dramatically: 
```cpp
0:  |F(x)| = 0.0194247; step 1
1:  |F(x)| = 0.00301762; step 1
2:  |F(x)| = 0.000207081; step 1
3:  |F(x)| = 3.60293e-06; step 1
4:  |F(x)| = 2.80194e-08; step 1
5:  |F(x)| = 1.39458e-10; step 1
6:  |F(x)| = 3.44131e-13; step 1
7:  |F(x)| = 2.86443e-16; step 1

```

vs 


```cpp
0:  |F(x)| = 0.0382248; step 1
1:  |F(x)| = 0.0102671; step 1
2:  |F(x)| = 0.00692618; step 1
3:  |F(x)| = 0.0017736; step 1
4:  |F(x)| = 0.000146487; step 1
5:  |F(x)| = 5.72064e-06; step 1
6:  |F(x)| = 1.98364e-06; step 1
7:  |F(x)| = 4.99592e-07; step 1
8:  |F(x)| = 5.5456e-08; step 1
9:  |F(x)| = 6.83088e-09; step 1
10:  |F(x)| = 8.3604e-10; step 1
11:  |F(x)| = 3.82326e-10; step 1
12:  |F(x)| = 2.96208e-11; step 1
13:  |F(x)| = 2.95233e-12; step 1
14:  |F(x)| = 5.11091e-13; step 1
15:  |F(x)| = 4.30558e-14; step 1


```

but the only assembly of the jacobian done to precondition the inner Krylov inside the `update` function increases the solve time significantly. I will do some more checks tomorrow, and then hopefully run `cProfile` over the weekend to isolate the bottleneck (it most likely is the calculation of jacobian)
```python3
J = asm(jacNewton, iBasis, w=iBasis.interpolate(u_full))
```

I will start back on completing #439  tomorrow. Many thanks, again!!

gdmcbain:
One possibility for the hyperelastic problem is what Knoll & Keyes call 'physics-based preconditioning', i.e. approximating the Jacobian that is used in the preconditioner, noting that this does not weaken the finite-difference Jacobian defined in terms of the residual as used in the Newton&ndash;Krylov iteration proper.

In this example, the Jacobian is only ever actually assembled once at the beginning, and based on the initial guess, so it's really the linear Laplace operator rather than the Jacobian of the nonlinear Young&ndash;Laplace operator.  This wasn't exploited in the present implementation but would involve :
* less derivation
* less coding
  * actually the `skfem.models.poisson.laplace` is predefined
* quicker assembly

Is there a 'physics-based preconditioner' for hyperelasticity, e.g. corresponding to linear elasticity?  If it omits some of the terms that proved tricky in #439, it might be quicker to assemble.

PR #452, opened 2020-08-11
pin at pacopy 0.1.2

### LINKED ISSUE #451: pacopy goes GPL
It looks like another outbreak of &lsquo;clarify the licensing of the examples&rsquo; #445 arose overnight; following pygmsh #446, pacopy went from [MIT](https://github.com/nschloe/pacopy/blob/2dd70b80ddc3cf8f1ef9328413f27e41f4016125/LICENSE) to GPL, under
* commit &lsquo;add setup.cfg&rsquo; https://github.com/nschloe/pacopy/commit/31369a497ea845e138ff30c53dcffa049b97c8f4#diff-9879d6db96fd29134fc802214163b95a 
* PR &rsquo;modernize&rsquo; nschloe/pacopy#16.

While we could just do as for pygmsh #446 and explicitly GPL exx 23 (Bratu&ndash;Gelfand, using `pacopy.euler_newton`) and 27 (steady Navier&ndash;Stokes, using `pacopy.natural`), I'm inclined to suggest a different path.

I assume that the relicensing is not retrospective, so we could pin to the last MIT version.  There has been [very little development](https://github.com/nschloe/pacopy/issues?q=is%3Aissue+is%3Aclosed) on pacopy since (my) addition of milestones in April 2019 in nschloe/pygmsh#5, which is essential to the practical application of `pacopy.natural` to nonlinear problems.

i had also recently already been thinking anyway of writing a new independent routines for natural parameter and false-transient continuation in the context of JFNK #442 #449 since I don't think `pacopy.natural` is compatible with JFNK, running the Newton iteration itself.  In simple cases like ex10, it's enough to use a simple `for`-loop in the natural parameter (amplitude of boundary deflexion in that case).  The adaptive natural parameter continuation of `pacopy.natural` in ex27 is nice though, so I'll probably copy that from the last MIT version.  Were I to write my own continuation routine, it would almost certaintly be an `Iterator`, avoiding the need for callbacks.

gdmcbain:
Fixes #451.

This is a quick interim measure, avoiding having to specially license exx 23, 27; later I envisage writing a fresh natural parameter continuation function that is compatible with the use of `functools.partial(scipy.optimize.root(method="krylov")` as nonlinear solver #119. 

ex23 is not particularly important; it could be dropped if it came to it.  It was primarily developed as a first step in using scikit-fem with pacopy.

ex27 however is very important.  It's the first example solving the Navier&ndash;Stokes equation.  It contains much that is original that should be reused and expanded in subsequent examples.  It should not be brought under GPLv3+.

This PR specifies all references to pacopy as being to pacopy 0.1.2, https://pypi.org/project/pacopy/0.1.2/, except for those in the paper

https://github.com/kinnala/scikit-fem/blob/80a4f7d38f7aa3a9972969110ae5e9b8e6672efd/paper/paper.md#L48

https://github.com/kinnala/scikit-fem/blob/80a4f7d38f7aa3a9972969110ae5e9b8e6672efd/paper/paper.bib#L111-L116

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
   - conda info -a
   - conda create -q -n scikit-fem python=$TRAVIS_PYTHON_VERSION nomkl numpy scipy matplotlib ipython pip
   - source activate scikit-fem
-  - pip install meshio dmsh pacopy pygmsh pyamg flake8
+  - pip install meshio dmsh pacopy==0.1.2 pygmsh pyamg flake8
 script:
   - flake8 skfem || travis_terminate 1;
   - python -m unittest discover
--- a/docs/examples/ex10.py
+++ b/docs/examples/ex10.py
@@ -30,7 +30,7 @@ x = np.zeros(basis.N)
 
 I = m.interior_nodes()
 D = m.boundary_nodes()
-x[D] = np.sin(np.pi * m.p[0, D]) 
+x[D] = np.sin(np.pi * m.p[0, D])
 
 for itr in range(100):
     w = basis.interpolate(x)
--- a/docs/examples/ex23.py
+++ b/docs/examples/ex23.py
@@ -2,9 +2,9 @@ r"""Bratu–Gelfand.
 
 .. note::
 
-   This example requires the external package `pacopy <https://github.com/nschloe/pacopy>`_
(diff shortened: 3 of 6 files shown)


gdmcbain:
I see that ex27 was already put under GPLv3 in https://github.com/kinnala/scikit-fem/commit/6391a87f4c4ed457ad22d5364db646b2a65f55cd because of pygmsh's licence-switch.

PR #454, opened 2020-08-11
WIP: ex27-continuation

gdmcbain:
The idea here is to break ex27 away from pacopy.  The immediate motivation is that it has switched licences #451 and we want to keep ex27 under the same licence as the rest of scikit-fem #453 but also longer term, it would be good to have a more flexible natural parameter continuation scheme, e.g. permitting `functools.partial(scipy.optimize.root(method="krylov")` as a solver #119 rather than insisting on its own implementation of Newton's method.


gdmcbain:
This includes #459, since we want to use the static mesh in ex27.

gdmcbain:
This is in a working state and free from pacopy but I'll continue modifying it to:
* adopt `partial(root, method="krylov")` #119, if possible 
* make the `natural` function included here within the example generally useful, perhaps moving it to `skfem.utils`

gdmcbain:
The failure is due to scipy/scipy#12813 which should be fixed upstream by scipy/scipy#12814.

gdmcbain:
However, this approach (JFNK via `scipy.optimize.root`) #119 after quite a bit of effort is not improving over the current implementation using pacopy 0.1.2 so I'm inclined to withdraw this (along with #442 and #449 which are mostly of interest as stepping-stones towards this).

Issue #456, opened 2020-08-12
Assigning dirichlet BC to a component of a vector field 


bhaveshshrimali:
In a way this is related to #455 but may be relevant in general too. What's the best way to select dofs that belong to a specific component of the solution? 

For instance, consider a unit cube where I want to specify u_1 = 0 on the left (x=0), u_2 = 0 on the bottom (y=0) and u_3 = 0 on the back (z=0) faces. Note that the remaining two components on each face are left unconstrained (physically, like in uniaxial tension). Currently I am tempted to think of it in a brute force way: 

```python3

from numpy import hstack, unique
stretch_ = 1.5
dofs = {
    "left":basis["u"].get_dofs(lambda x: x[0] == 0.),
    "bottom":basis["u"].get_dofs(lambda x: x[1] == 0.),
    "back":basis["u"].get_dofs(lambda x: x[2] == 0.),
    "front":basis["u"].get_dofs(lambda x: x[2] == 1.)
}

du[dofs["left"].nodal["u^1"]] = 0.
du[dofs["left"].edge["u^1"]] = 0.

du[dofs["bottom"].nodal["u^2"]] = 0.
du[dofs["bottom"].edge["u^2"]] = 0.

du[dofs["back"].nodal["u^3"]] = 0.
du[dofs["back"].edge["u^3"]] = 0.

du[dofs["front"].nodal["u^3"]] = stretch_
du[dofs["front"].edge["u^3"]] = stretch_

I = basis["u"].complement_dofs(dofs) #this selects the complement of everything in `dofs`

I = hstack((
    I, dofs["left"].nodal["u^2"],
    dofs["left"].nodal["u^3"],
    dofs["left"].edge["u^2"],
    dofs["left"].edge["u^3"],
    dofs["bottom"].nodal["u^1"],
    dofs["bottom"].nodal["u^3"],
    dofs["bottom"].edge["u^1"],
    dofs["bottom"].edge["u^3"],
    dofs["back"].nodal["u^1"],
    dofs["back"].nodal["u^2"],
    dofs["back"].edge["u^1"],
    dofs["back"].edge["u^2"],
    dofs["front"].nodal["u^1"],
    dofs["front"].nodal["u^2"],
    dofs["front"].edge["u^1"],
    dofs["front"].edge["u^2"]
))

I = unique(I)

```

I am still thinking if the above is correct (as in it doesn't duplicate dofs or selects dofs that have dirichlet boundary specified). 

[...]

PR #446, opened 2020-07-29
Add explicit GPLv3 license to the files depending on pygmsh

kinnala:
Closes #445

PR #457, opened 2020-08-13
Remove old style forms

kinnala:
I think we are soon in the position to remove these and release 2.0.0.

@gdmcbain Do you think we should do it before or after JOSS review has officially completed or does it matter?


gdmcbain:
What else would be in mind for 2.0.0?  Resolving `get_dofs` and `find_dofs`, as arose today in #456 ?

gdmcbain:
Not sure whether this is the place to start the 2.0.0 wishlist (also I don't know whether it matters: push out 2.0.0 with the obsolete form-style banished, then push and 3 and 4 with whatever: i can see the sense in adhering to semantic versioning but I don't know whether there are meaningful constraints on frequency at this stage).

But back to the wishlist.  Following #446 and #453, I've been wondering whether perhaps indeed for 2.0.0, i.e. really quite soon, we should pull other examples back under the common licence as examples do contain important snippets which are partial extensions of the library which might make their way into the library (e.g. handing of Dirichlet conditions, building preconditioners with pyamg or amgcl, specific preconditioners like those developed so far for the Stokes equation, ...).  Should most of those examples from #446 be provided with static meshes?  And then maybe provide one example that just shows how to use pygmsh with skfem and just put that example under the pygmsh licence; say it might just redo ex01 but with (not particularly necessary but valid) explicit .boundaries and .subdomains.  I'm thinking that a number of the examples might share static mesh files, so that there aren't too many lying around.  A nice disk, for example, would cover a few and perhaps even a disk with distinguished subdomains for core and annulus might also serve as a simple disk if those subdomains are simply ignored.  (I think that works and should be harmless enough.)  Sorry, this should be two new issues: (i) 2.0.0 wishlist, (ii) shared static meshes to reclaim examples from pygmsh licence.

PR #459, opened 2020-08-18
static meshes

### LINKED ISSUE #458: avoid pygmsh in examples
As part of a discussion in #457  of what might be done before 2.0.0:

> Following #446 and #453, I've been wondering whether perhaps indeed for 2.0.0, i.e. really quite soon, we should pull other examples back under the common licence as examples do contain important snippets which are partial extensions of the library which might make their way into the library (e.g. handing of Dirichlet conditions, building preconditioners with pyamg or amgcl, specific preconditioners like those developed so far for the Stokes equation, ...). Should most of those examples from #446 be provided with static meshes? And then maybe provide one example that just shows how to use pygmsh with skfem and just put that example under the pygmsh licence; say it might just redo ex01 but with (not particularly necessary but valid) explicit .boundaries and .subdomains. I'm thinking that a number of the examples might share static mesh files, so that there aren't too many lying around. A nice disk, for example, would cover a few and perhaps even a disk with distinguished subdomains for core and annulus might also serve as a simple disk if those subdomains are simply ignored. (I think that works and should be harmless enough.) Sorry, this should be two new issues: (i) 2.0.0 wishlist, (ii) shared static meshes to reclaim examples from pygmsh licence.

gdmcbain:
Fixes #458 .


gdmcbain:
The static disk mesh is also adopted for ex18, removing the dependency on dmsh (which also has its own licence, hitherto overlooked in LICENSE.md).

Issue #455, opened 2020-08-11
Mixed formulation for Incompressible Hyperelasticity


bhaveshshrimali:
Following #439  and inputs from [ex18](https://github.com/kinnala/scikit-fem/blob/master/docs/examples/ex18.py), I started working yesterday on a [mixed formulation](http://pamies.cee.illinois.edu/Publications_files/IJNME_2015.pdf#page=4&zoom=100,312,414) for incompressible hyperelasticity using the Taylor-Hood elements. The idea is to work on imposing Dirichlet BC's later today and then test a simple uniaxial tension later in the week. This should be fairly self contained. This is what I have so far. Please let me know whenever you get time to take a look. No rush from my end 

```python3
from numpy import einsum, linalg as nla, zeros, zeros_like, concatenate, abs as npabs, split as npsplit
from scipy.sparse import bmat
from sympy import Symbol, simplify, lambdify, diff, sqrt
from skfem.helpers import grad, transpose
from typing import List, Optional
from numba import jit, prange
from skfem import *

@jit(cache=True, nopython=True, nogil=True, parallel=True)
def vdet(A):
    detA = zeros_like(A[0, 0])
    detA = A[0, 0] * (A[1, 1] * A[2, 2] -
                    A[1, 2] * A[2, 1]) -\
        A[0, 1] * (A[1, 0] * A[2, 2] -
                    A[1, 2] * A[2, 0]) +\
        A[0, 2] * (A[1, 0] * A[2, 1] -
                    A[1, 1] * A[2, 0])
    return detA

@jit(cache=True, nopython=True, nogil=True, parallel=True)
def vinv(A):
    invA = zeros_like(A)
    detA = vdet(A)
    invA[0, 0] = (-A[1, 2] * A[2, 1] +
                A[1, 1] * A[2, 2]) / detA
    invA[1, 0] = (A[1, 2] * A[2, 0] -
                A[1, 0] * A[2, 2]) / detA
    invA[2, 0] = (-A[1, 1] * A[2, 0] +
                A[1, 0] * A[2, 1]) / detA
    invA[0, 1] = (A[0, 2] * A[2, 1] -
                A[0, 1] * A[2, 2]) / detA
    invA[1, 1] = (-A[0, 2] * A[2, 0] +
                A[0, 0] * A[2, 2]) / detA
    invA[2, 1] = (A[0, 1] * A[2, 0] -
                A[0, 0] * A[2, 1]) / detA
[...]

PR #462, opened 2020-08-23
Hyperelasticity example

bhaveshshrimali:
This PR adds an example illustrating a mixed finite element implementation of (nearly) incompressible hyperelasticity #455 using the Taylor-Hood element. It also adds two helper functions `inv` and `det` to `skfem.helpers.py` to compute the inverse and determinant of higher order tensors over trailing axes.


bhaveshshrimali:
trialFigureWhiteBG

A tiny bit more discernible?

gdmcbain:
Is the solution supposed to be isobaric?  If so, plotting the pressure seems redundant.

I think the local magnitude of the velocity is more often written |_u_|, reserving ‖_u_‖ for a norm of the velocity field.

bhaveshshrimali:
> Is the solution supposed to be isobaric? If so, plotting the pressure seems redundant.

The pressure for a homogeneous domain under uniaxial loading should be equal to `-mu/l`, which is exactly recovered by the `P1` elements in this case. But I agree the plot doesn't reveal any more insight than this.

bhaveshshrimali:
> I think the local magnitude of the velocity is more often written |_u_|, reserving ‖_u_‖ for a norm of the velocity field.

Ah, that was a bit sloppy. I will change this now. Thanks!!

As an aside, do you know the reason behind the failing build? I saw `pacopy` was pinned to `0.1.2` due to the licensing issue recently.

gdmcbain:
> As an aside, do you know the reason behind the failing build? I saw pacopy was pinned to 0.1.2 due to the licensing issue recently.

Hmm.  I'll look into it.  I think the pacopy stuff was all passing when #452 #453 #457 #459 #461 were merged, but let me check.

An aside to aside: I'm doing much work on re-implementing natural parameter continuation without pacopy, as the latter gets in the way of JFNK #119.  I expect that some kind of continuation (natural parameter or perhaps pseudo-transient) will be helpful in some more nonlinear problems in elasticity.

bhaveshshrimali:
> expect that some kind of continuation (natural parameter or perhaps pseudo-transient) will be helpful in some more nonlinear problems in elasticity.

Yeah some cool nonlinear buckling problems that often arise in rubber-like hyperelastic materials . I haven't gone to solving those yet either so definitely a nice problem to try out in the future. Thanks

gdmcbain:
I see, yes: https://travis-ci.com/github/kinnala/scikit-fem/jobs/376714751 has

> ERROR: Could not find a version that satisfies the requirement pacopy==0.1.2 (from versions: none)

> ERROR: No matching distribution found for pacopy==0.1.2

> The command "pip install meshio pacopy==0.1.2 pygmsh pyamg flake8" failed and exited with 1 during .

I'll raise an issue for this: #463.

gdmcbain:
In looking into the Travis CI failure #463, it so happened that I saw something else that needed fixing (unnecessary apt install of libsuitesparse #464), so I submitted a PR to omit that #465 and it passed.  I'm pretty sure that (_a_) your new example here is independent of pacopy and (_b_) #465 is independent of pacopy, so I'm puzzled that #465 passes but this PR fails.

bhaveshshrimali:
> ERROR: Could not find a version that satisfies the requirement pacopy==0.1.2 (from versions: none)

I think this is no longer occuring, but I'm confused as to why it was before. I am addressing the whitespace/newline issues.

bhaveshshrimali:
trialFigureWhiteBGdispOnly

Removed the pressure contour and edited the legend for the displacement plot. I will resume this later tomorrow to check for typos etc. in the description, but feel free to suggest changes. 

I have tried to keep the description short and yet not excessively cut on essential details that would describe the problem sufficiently well.

skfem/helpers.py
```diff
@@ -97,3 +97,63 @@ def eye(w, n):
     """Create diagonal matrix with w on diagonal."""
     return np.array([[w if i == j else 0. * w for i in range(n)]
                      for j in range(n)])
+
+
+def det(A):
```
kinnala, line 102: Should be easy to make it work also for 2-by-2 matrices? `if A.shape[0]==2: ...`

skfem/helpers.py
```diff
@@ -97,3 +97,63 @@ def eye(w, n):
...
+    ----------
+    A : (N, N,...) numpy.ndarray
+        Input array whose inverse is to be computed
+    Returns
+    -------
+    Ainv : (N, N,...) numpy.ndarray
+        Inverse of `A`.
+    """
+    invA = zeros_like(A)
+    detA = A[0, 0] * (A[1, 1] * A[2, 2] - A[1, 2] * A[2, 1]) -\
```
kinnala, line 137: Can this use `detA = det(A)` ?

skfem/helpers.py
```diff
@@ -97,3 +97,63 @@ def eye(w, n):
...
+    detA = A[0, 0] * (A[1, 1] * A[2, 2] -
+                      A[1, 2] * A[2, 1]) -\
+        A[0, 1] * (A[1, 0] * A[2, 2] -
+                   A[1, 2] * A[2, 0]) +\
+        A[0, 2] * (A[1, 0] * A[2, 1] -
+                   A[1, 1] * A[2, 0])
+    return detA
+
+
+def inv(A):
```
kinnala, line 124: It should also be easy to make this work for 2x2 matrices.

docs/examples/ex36.py
```diff
@@ -0,0 +1,221 @@
+r"""Incompressible Hyperelasticity
+This example solves the governing equations describing the mechanical
```
kinnala, line 2: Can you also add a short version of the description to `docs/listofexamples.rst` together with the image?

The image can be uploaded to this issue https://github.com/kinnala/scikit-fem/issues/431 after which you can get an url with right click.

kinnala (changes requested):
Some improvements.

skfem/helpers.py
bhaveshshrimali: Ah I see, completely forgot that these were to be modified to include the `2x2` cases as well.

skfem/helpers.py
bhaveshshrimali: I made these changes. Would test a couple more things. For instance, now that I remember it should be possible to write the inverse entirely using detA and `trace(A^2)` and `trace(A` and implementing these using `np.einsum`. If this is performant then would be valid for matrices of any rank (although I don't know where would these cases arise)

Same goes for `det(A)`.

bhaveshshrimali: Just tested the Cayley-Hamilton version

```python3
def finv(A):
    detA = vdet(A)
    I1 = einsum("ii...", A)
    I2 = -0.5 * (einsum("ik...,ki...", A, A) - I1**2)
    Asq = einsum("ik...,kj...->ij...", A, A)
    eye = zeros_like(A)
    eye[0, 0] = 1.
    eye[1, 1] = 1.
    eye[2,2] = 1.
    return 1./detA * (Asq - I1 * A + I2 * eye)
```

It's significantly (`~10x`) slower and more complicated than your version. Still faster than `numpy.linalg.inv` on my machine, which for some reason is extremely slow

kinnala:
Fantastic. I think this is ready. I'll merge it later today and check that the documentation recompiles properly.

bhaveshshrimali:
Thanks a lot for your help. I did open up an [issue on numpy](https://github.com/numpy/numpy/issues/17166) regarding `inv` being very slow for such cases :)

kinnala:
The figure was created with vedo?

bhaveshshrimali:
No, Paraview.

kinnala:
Thanks. I made some small code formatting changes and also decided to remove the saddle point problem from the short description because I didn't feel like the notation was familiar to non-experts. Better to read the entire description from ex36.py.

Thanks a lot for your contribution to scikit-fem. You are now entitled to have your name and ORCiD added to the Zenodo archive: https://github.com/kinnala/scikit-fem/blob/master/.zenodo.json If you wish to be included, please open another PR that adds your information.

bhaveshshrimali:
> If you wish to be included, please open another PR that adds your information.

Thanks a lot for considering it. I will open a PR for the same.

PR #465, opened 2020-08-24
omit libsuitesparse

gdmcbain:


gdmcbain:
A partial motivation for this submission was to see whether pinning pacopy 0.1.2 #452 had broken something; it seemed to trip Travis CI in the new hyperelasticity example #462.

PR #468, opened 2020-09-12
Add ElementHex2

### LINKED ISSUE #327: ElementHex2

kinnala:
Here is some work-in-progress how ElementHex2 could look like (not fully working yet I'm afraid). There are some optimizations possible, but I need to measure it first. Closes #327 when ready.


kinnala:
hexa2
Something's wrong, most likely ordering of the DOF's.

kinnala:
I'm slightly confused that none of the unit tests catch this issue and only the convergence test fails.

kinnala:
Yeah, had doflocs in the wrong order. :facepalm:

kinnala:
I'm now wondering if this way of constructing the basis should be a general strategy: start with `ElementGlobal` on the reference element and then transform that into `ElementH1`.

kinnala:
I'll leave it like this for now; now that the tests exist and the implementation is correct, we can optimize it later if needed.

Issue #481, opened 2020-09-29
Indexing error in find_dofs


mjziebarth:
Hi,

first off, great library: Indeed fills an important niche, thanks for creating!

While using it, being new to finite element methods, I have encountered an error that I am not sure how to fix (whether it's a wrong use on my end or a problem with the code). When trying to implement Dirichlet boundary conditions (following example 14) in a MeshTet, I am encountering a numpy indexing error.

My script condenses to the following:
```python
m = MeshTet3D(p=some_nodes, t=some_elements, boundaries=some_boundaries)
e = ElementTetP1()
basis = InteriorBasis(m, e)
surface_basis = FacetBasis(m, e, facets=m.boundaries["top"])
surface_dofs = surface_basis.find_dofs()['all'].all()
```
The `boundaries` dict contains, among others, an array of faces belonging to the top side. Things run smoothly up to the `surface_basis.find_dofs()` call, where the following error is raised:
```
~/.local/lib/python3.8/site-packages/skfem/mesh/mesh3d/mesh3d.py in boundary_edges(self)
     39                               facets]))
     40                    for itr in range(self.facets.shape[0])])).T, axis=1)
---> 41         return np.nonzero((self.edges.T[:, None] == boundary_edges)
     42                           .all(-1).any(-1))[0]
     43 
AttributeError: 'bool' object has no attribute 'all'
```
The error occurs because the shapes of `self.edges.T[:,None]`, `(180751, 1, 2)`, and `boundary_edges`, `(107076, 2)`, do not match for an elementwise comparison. From a quick skim of the code, I assume that some sort of element comparison is taking place in that function, but I am not sure about that (without some in depth study of the source). I thought to open this issue because maybe there is something fundamentally wrong in my code snipped above and I hope it's not too stupid, or otherwise it might be a bug.

Thanks in advance,
Malte

kinnala:
I might have an idea what it is: can you try `basis.find_dofs` instead?

mjziebarth:
Thanks for the quick reply!
I tried it out, unfortunately the same error appears.

I will try to dig deeper to find the root of this. Is there some way that I could validate that the objects are properly initialized?

kinnala:
Can you run this with the Python debugger and print the call stack with `where`?

Something like `ipython test.py -m pdb` then in the resulting prompt press `c` and wait until the execution halts to the error. Then write `where` and paste the output here.

kinnala:
[...]

PR #485, opened 2020-09-30
Optimize boundary_edges

### LINKED ISSUE #483: Memory runs out when finding dofs in a large 3D mesh
> I am running the latest version (2.0.0) from PyPI. Your example also runs fine:
> > The following runs fine when using the master branch:
> > 
> > ```python
> > from skfem import *
> > m = MeshTet()
> > m.define_boundary("top", lambda x: x[2] == 1.0)
> > e = ElementTetP1()
> > basis = InteriorBasis(m, e)
> > surface_basis = FacetBasis(m, e, facets=m.boundaries["top"])
> > surface_dofs = surface_basis.find_dofs()
> > ```
> > 
> > Does it run on your computer?
> 
> While building a smaller example to share here (which runs fine!), I think I might have found the root of the error:
> An array of size (180751, 107076, 2) is just to large for my RAM (that would be about 36GB for a bool array and I'm working > on a notebook with 8GB). Instead of raising a memory error, however, the comparison just silently fails and returns a single bool. Not something I would have expected, else I would have checked that earlier..
> 
> I will find a way to get along with that and at some point switch to a larger machine. Thanks for the quick help and feedback!
> 
> _Originally posted by @mjziebarth in https://github.com/kinnala/scikit-fem/issues/481#issuecomment-701214738_

I think the above is due to the logic how facets get expanded to edges: it currently creates a huge array. A quick workaround would be to disable facet expansion when using low-order elements where edge DOF's are not needed.

A long term approach would be to reduce the amount of comparisons done but not sure yet how to do that in practice.

kinnala:
Fixes #483 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Unreleased
 
+### Fixed
+
+- `Mesh3D.boundary_edges` (and, consequently, `Basis.find_dofs`) was slow
+  and used lots of memory due to an exhaustive search of all edges
+
 #### Added
 - `ElementHex2`, a triquadratic hexahedral element
 - `MeshTri.init_circle`, constructor for a circle mesh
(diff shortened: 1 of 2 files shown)


kinnala:
@mjziebarth here is a possible upcoming fix for your issue in #481

Issue #484, opened 2020-09-30
Support pygmsh 7.0.0 in the examples


kinnala:
I was unable to run pygmsh 7.0.0 on my laptop so I didn't do this while looking into #482 .

I suppose now is also the time to do #333 because otherwise I might be unable to run the tests locally unless they support both 6.x and 7.x.

kinnala:
I think this package `gmsh` is again a good example of why you shouldn't distribute binaries in a PyPI package unless you know what you're doing: https://gitlab.onelab.info/gmsh/gmsh/-/tree/master/utils/pypi/gmsh

ahojukka5:
If you need to just read gmsh file formats, consider using gmshparser. It has zero dependencies (the point of the package is not to introduce any complicated dependencies to FEM projects).

https://github.com/ahojukka5/gmshparser

gdmcbain:
Oh, O. K.

I had been relying routinely upon `pip install gmsh-sdk` to get a working recent Gmsh on Windows (especially during the lockdown) and then as it was so easy even back on Linux, until I was informed last week in pymor/pymor#1093 that it had been deprecated (in favour of `gmsh` or `gmsh-dev`).  Now that pygmsh depends on gmsh and not gmsh-sdk, I can't get pygmsh to install at all on Windows: nschloe/pygmsh#370.  This puzzled me a bit though as when I looked at the code for `pip install gmsh`, it seemed identical to `gmsh-sdk`, having been copied from it.

gdmcbain:
> If you need to just read gmsh file formats

pygmsh doesn't read gmsh file formats.  meshio does that.  pygmsh generates Gmsh GEO code.

gdmcbain:
There's been an upstream change in the PyPI package gmsh; `pip install gmsh` seems to work O. K. now (Ubuntu 20.04.1 LTS and MS-Windows 10).  It's a little slow on the step &lsquo;Building wheel for gmsh (setup.py)&rsquo; but gets there in the end.  Similarly `pip install pygmsh` works, bringing in `pygmsh-7.0.0` and `gmsh-4.6.0.post3`.

(Not that I'm pushing to unpin pygmsh just yet.)

kinnala:
[...]

Issue #482, opened 2020-09-30
Tests using pygmsh fail due to upstream changes


kinnala:
This is due to the commit: https://github.com/nschloe/pygmsh/commit/ed581be373ed7357b4bb83183749a10760034892#diff-bc0dadc6113a14c1c7891320765ee767

kinnala:
Seems to be a simple enough fix: `generate_mesh` was turned into a method of `Geometry`.

kinnala:
Wow, actually this new `pygmsh` release using `gmsh` package fails pretty fatally on my operating system:
```
        if handle is None:
>           self._handle = _dlopen(self._name, mode)
E           OSError: libXcursor.so.1: cannot open shared object file: No such file or directory
```

kinnala:
#365 causes the tests to fail but #360 is the PR which causes this to fail on my laptop.

kinnala:
I need to look into this later, I'm now a bit busy with the grant application.

kinnala:
Ok, the grant deadline was extended so I can now look into this.

kinnala:
```
(vedoenv) conda-shell-chrootenv:tom@tunkki:~/src/scikit-fem$ python -c "import gmsh"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/tom/.conda/envs/vedoenv/lib/python3.8/site-packages/gmsh-4.6.0-Linux64-sdk/lib/gmsh.py", line 39, in <module>
    lib = CDLL(libpath)
  File "/home/tom/.conda/envs/vedoenv/lib/python3.8/ctypes/__init__.py", line 373, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libXcursor.so.1: cannot open shared object file: No such file or directory
```

kinnala:
I'm getting some of the required libraries from nixpkgs but it seems that the binaries distributed in `gmsh` package are compiled against some newer versions of xorg-related libraries that I can easily to install.

Workaround is to restrict to an older version of pygmsh which I'm going to do now. Maybe this is finally the reason to do #333 .

kinnala:
Pinned pygmsh to 6.1.1. Making another issue to support 7.0.0.

PR #486, opened 2020-10-01
mesh ex13 ourselves

gdmcbain:
With another round of pygmsh misery upon us #482 #484, I thought it time to lift one more of the examples out of it and mesh it ourselves.


gdmcbain:
ex13

The solution is hardly changed, but the slight artifacts of the mesh are different, so here's a new figure.

PR #489, opened 2020-10-06
Pygmsh 7 support

### LINKED ISSUE #484: Support pygmsh 7.0.0 in the examples

kinnala:
Fixes #484 . Let's see if it runs. Github Actions is now running only test_examples.


kinnala:
There are some hints in Makefile if you want to run this container locally. I'll later add instructions to README when it's more mature.

PR #493, opened 2020-10-10
Preliminary implementation of MeshTri2 and MeshQuad2

kinnala:
Basic features such as assembly of interior forms works.

Many features do not work properly or have not been tested such as:
- boundary assembly
- refines
- alternative constructors

A warning is given which explains that this is an experimental feature not subject to the semantic versioning.


kinnala:
Figure_1

PR #494, opened 2020-10-12
Add mypy support

kinnala:
Run mypy in CI.

### DIFF
--- a/skfem/assembly/form/linear_form.py
+++ b/skfem/assembly/form/linear_form.py
@@ -16,30 +16,30 @@ class LinearForm(Form):
     """
 
     def assemble(self,
-                 u_basis: Basis,
-                 v_basis: Optional[Basis] = None,
+                 ubasis: Basis,
+                 vbasis: Optional[Basis] = None,
                  **kwargs) -> ndarray:
 
-        assert v_basis is None
-        v_basis = u_basis
+        assert vbasis is None
+        vbasis = ubasis
 
-        nt = v_basis.nelems
-        dx = v_basis.dx
-        w = FormDict({**v_basis.default_parameters(), **self.dictify(kwargs)})
+        nt = vbasis.nelems
+        dx = vbasis.dx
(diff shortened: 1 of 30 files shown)


skfem/element/discrete_field.py
```diff
@@ -14,39 +14,21 @@ class DiscreteField(NamedTuple):
     hess: Optional[ndarray] = None
     hod: Optional[ndarray] = None
 
-    @property
```
kinnala, line 17: Backwards incompatible

skfem/assembly/form/linear_form.py
```diff
@@ -16,30 +16,30 @@ class LinearForm(Form):
     """
 
     def assemble(self,
-                 u: Basis,
-                 v: Optional[Basis] = None,
+                 ubasis: Basis,
+                 vbasis: Optional[Basis] = None,
```
kinnala, line 20: Backwards incompatible

kinnala:
There are two "technically backwards-incompatible" changes highlighted above. @gdmcbain do you think such small changes warrant for a major version increment or not? `DiscreteField.f/df/ddf` should have been removed already before in `2.0.0` but seems I have forgotten them. There was also a typing-related bug I fixed in `Form` which revealed another issue in the argument naming. This forced me to rename `u` to `ubasis` because the keyword argument `u=...` was used in example 13, line 65. The change is basically renaming `v=...` keyword argument to `vbasis=...`. I have never used this keyword argument explicitly so that's why I think it's a minor change, not warranting for a major version increment.

kinnala:
Decided to revert `DiscreteField.f/df/ddf` because that was easy enough.

gdmcbain (approved):
Great! I'm very pleased to see type-checking added to the CI.  

I note that no changes were required to `docs/examples`, so don't really think the changes to the API should effect any users; otherwise the examples should be expanded.  There are lots of changes in `tests` like `.f` &rarr; `.value` for a `DiscreteField`; I wonder why that's is exercised so much in the tests but not at all in the examples.

PR #498, opened 2020-10-16
Add MeshTet.init_ball

### LINKED ISSUE #490: generate ellipsoidal ex32.mesh in other than pygmsh
Given the awkwardness of pygmsh 7 support #489, the idea of generating the ellipsoidal mesh for ex32 using something else, e.g. [ngsimple](https://github.com/kinnala/ngsimple), was raised in #484.

kinnala:
@gdmcbain you think this closes #490 ?

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -34,7 +34,7 @@ later in this document or at the beginning of the respective files.
 Some examples under `docs/examples/` are licensed differently.  In particular,
 they have a GPL-licensed dependency and, therefore, are also GPL-licensed.
 
-- docs/examples/ex{28,32}.py
+- docs/examples/ex{28}.py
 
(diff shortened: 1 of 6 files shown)


gdmcbain:
Great! init_ball is very clever and very useful. May I ask whether you came up with this trick or where you found it?

I'm not sure about the triaxial scaling though. It might be better to change the example to the special case of the sphere—until we have some kind of tetrahedral smoother or optimizer. Are the elements quite pointy? I look forward to sketching out the action of init_ball when I get back to the desk and pencil and paper. A sphere would be fine anyway for demonstrating the effectiveness of the block diagonal preconditioner.

kinnala:
@gdmcbain will you make a new PR when it's ready?

Issue #505, opened 2020-11-04
Computing curl for ElementTriMorley


YerbaPage:
Hi, It's me again! Thanks so much for this library and your help!

Recently I'm trying to assemble BilinearForm like (curl u, curl v) for Morley element. I tried to implement it this way:
``` python
m = MeshTri()
m.refine(5)

@BilinearForm
def zv_load(u, v, w):
    return dot(curl(u), curl(v))

element = ElementTriMorley()
basis = InteriorBasis(m, element)

K = asm(zv_load, basis)
```

But the following error is raised:
```
~\Anaconda3\lib\site-packages\skfem\helpers.py in curl(u)
     31     if u.curl is not None:
     32         return u.curl
---> 33     raise NotImplementedError
     34 
     35 

NotImplementedError: 
```

I'm wondering whether there are some problems in my codes or there are some other things that we could implement? Could you give me any hint of what I could do?

Thanks in advance!

kinnala:
The helper "curl" is implemented currently only for H(curl)-conforming
elements. But since Morley is a two-dimensional element, what do you
exactly mean by "curl" in 2D?

gdmcbain:
Maybe something like **rot** _&psi;_ = **curl** (_&psi;_ **k**)?  That gives the two-dimensional velocity from its stream-function, as in

https://github.com/kinnala/scikit-fem/blob/4ba1e06a01856e825d8d2f9cdfcd5742c5376f66/docs/examples/ex20.py#L73-L74

gdmcbain:
&hellip;or, as in #506,

https://github.com/kinnala/scikit-fem/blob/dece91061fec31e0dbe13d69bf549e17cc195e0b/skfem/models/general.py#L20-L23

https://github.com/kinnala/scikit-fem/blob/dece91061fec31e0dbe13d69bf549e17cc195e0b/docs/examples/ex20.py#L71-L75

kinnala:
> Maybe something like **rot** _ψ_ = **curl** (_ψ_ **k**)? That gives the two-dimensional velocity from its stream-function, as in
> 
> https://github.com/kinnala/scikit-fem/blob/4ba1e06a01856e825d8d2f9cdfcd5742c5376f66/docs/examples/ex20.py#L73-L74

So basically extending https://github.com/kinnala/scikit-fem/blob/master/skfem/helpers.py#L31 by something like
```python
if u.grad is not None:
    rot_grad = np.zeros_like(u.grad)
[...]

PR #506, opened 2020-11-05
skfem.models.general.vrot as adjoint of rot

gdmcbain:
Inspired by the discussion in #505, I defined `skfem.models.general: LinearForm`, being the adjoint (is that the right word) of `skfem.model.general.rot` in the sense that (**u**, **vrot** _&psi;_) = (rot **u**, _&psi;_), where everything is two-dimensional and _&psi;_ and **u** are scalar and vector fields, respectively.

This is used to simplify the calculation of the velocity from the stream-function in ex20, where the stream-function is defined using `ElementTriMorley`, as in the question #505 (though the actual question was about a *bi*-linear form).

I'm not at all sure about the name!  Sometimes I've seen this pair distinguished by bold and roman type (rot **u** is a scalar, the vorticity; **rot** _&psi;_ is a vector, the velocity, and rot **rot** _&psi;_ = &minus;&Delta; _&psi;_); here the bold one gets a v for vector at the front, but alternative suggestions are welcome.

### DIFF
--- a/skfem/models/general.py
+++ b/skfem/models/general.py
@@ -15,3 +15,9 @@ def divergence(u, v, w):
 def rot(v, w):
     return np.einsum('i...,ij...,j...',
                      w.w, np.array([[0, 1], [-1, 0]]), grad(v))
+
+
+@LinearForm
+def vrot(v, w):
+    return np.einsum('i...,ij...,j...',
+                     v, np.array([[0, 1], [-1, 0]]), grad(w['w']))
--- a/skfem/models/general.py
+++ b/skfem/models/general.py
@@ -1,9 +1,7 @@
 """Bilinear and linear forms too general to put into a specific model."""
 
 from skfem.assembly import BilinearForm, LinearForm
-from .helpers import div, grad
-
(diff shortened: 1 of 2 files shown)


gdmcbain:
ex20_velocity-vectors

*Figure:&mdash;* The `ElementVectorH1(ElementTriP1())` velocity vectors computed as the **vrot** linear form of the `ElementTriMorley` stream-function

skfem/models/general.py
```diff
@@ -15,3 +15,9 @@ def divergence(u, v, w):
 def rot(v, w):
     return np.einsum('i...,ij...,j...',
                      w.w, np.array([[0, 1], [-1, 0]]), grad(v))
+
+
+@LinearForm
+def vrot(v, w):
+    return np.einsum('i...,ij...,j...',
```
kinnala, line 22: Using the extended definition of `curl` in #507, can this be written as `dot(curl(v), w.w)`?

gdmcbain: Yes, that's much nicer!  Thanks!

PR #507, opened 2020-11-05
Implement rotated gradient special case

kinnala:
See #505 .


gdmcbain (approved):
This makes #506 much better and the old `skfem.models.general.rot` (which maps plane velocity to vorticity) too.

Issue #508, opened 2020-11-09
ElementLineP0


gdmcbain:
We have `ElementTriP0` , `ElementQuad0` #266, `ElementTetP0`, but not `ElementLineP0` (or the hexahedron).

gdmcbain:
An anomaly:
* these are O. K.:
  * `FacetBasis(MeshTri(), ElementTriP0())`
  * `FacetBasis(MeshTet(), ElementTetP0())`
  * `FacetBasis(MeshLine(), ElementLineP0())`, from #509 at 25a636a2f6957769cc1eb46c0fcdd1daeb1de4f3
* but this isn't
  * `FacetBasis(MeshQuad(), ElementQuad0())`

raising `ValueError: operands could not be broadcast together with remapped shapes [original->remapped]: (4,) and requested shape (4,2)`

kinnala:
Seems there is actually a bug in `element_quad0.py`. `np.ones(X.shape[1])` should be replaced with, e.g., `np.zeros_like(X[0]) + 1`. Maybe `np.ones(X.shape[1:])` would also work. The thing is that `X` has different shape in `InteriorBasis` and `FacetBasis`.

PR #509, opened 2020-11-09
ElementLineP0

gdmcbain:
I think this is right but it needs more tests and an example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Unreleased
 
+- Added: `ElementLineP0`
 - Added: `skfem.helpers.curl` now calculates the rotated gradient for
   two-dimensional elements.
 - Added: `MeshTet.init_ball` for meshing a ball.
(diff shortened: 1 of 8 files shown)


gdmcbain:
This is for #508.

kinnala:
I think it's fine. You can add it to `TestDerivatives` and then add also a changelog entry to `README.md`.

gdmcbain:
Wow, adding `ElementLineP0` to `elems` in `TestDerivatives` causes it to hang!

gdmcbain:
It wasn't tripping the `ValueError` in

https://github.com/kinnala/scikit-fem/blob/e4c0a4f0e5593502a7cebf45461e176f9668ff91/tests/test_elements.py#L193-L196

e.g. `ElementLineP0.lbasis(y, 33)` returns the same as `ElementLineP0.lbasis(y, 0)`; indeed the second argument was ignored.

This was copied from other P0 elements so I've fixed this for all of them in b8d96fb.

kinnala:
Great! Ready to merge?

gdmcbain:
Yeah.  I'd prefer to have an example demonstrating the use of the new element but don't have an easy one in mind.

kinnala:
Ok. Thanks.

PR #511, opened 2020-11-09
Fix ElementQuad0 for FacetBasis

### LINKED ISSUE #510: Bug in ElementQuad0
Seems there is actually a bug in `element_quad0.py`. `np.ones(X.shape[1])` should be replaced with, e.g., `np.zeros_like(X[0]) + 1`. Maybe `np.ones(X.shape[1:])` would also work. The thing is that `X` has different shape in `InteriorBasis` and `FacetBasis`.

_Originally posted by @kinnala in https://github.com/kinnala/scikit-fem/issues/508#issuecomment-724058188_

kinnala:
Fixes #510

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Unreleased
 
-- Added: `ElementLineP0`
+- Fixed: `ElementQuad0` was not compatible with `FacetBasis`
+- Added: `ElementLineP0`, one-dimensional piecewise constant element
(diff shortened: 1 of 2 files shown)


gdmcbain (approved):
I'm not exactly sure how to interpret `doflocs` for a piecewise-constant element, but if it has one, should it be in the corner?   

https://github.com/kinnala/scikit-fem/blob/4e6ca0e448e4370ef022cb179d2a36555caf189c/skfem/element/element_quad/element_quad0.py#L13

I'm also not sure whether this quibble is related to the issue, so I won't hold up approval.

kinnala:
No. It should be in the middle. What a buggy element this is.

PR #516, opened 2020-11-13
Redundant preallocation

gdmcbain:
A couple of minor fixes in ex22.
* First I noticed that *u* was set to zero before the `solve`, but this hasn't been necessary since #228.
* Then I noticed that the first two meshes were the same; perhaps instead of `itr > 1`, `>= 1` was intended, or `> 0`, as suggested here.


gdmcbain:
I think the failure of 6fa649a was because one more refinement was made, slightly improving the solution.

PR #521, opened 2020-12-01
allow vectorial functionals

### LINKED ISSUE #520: vectorial functionals
(Very low priority, this one.)   I wonder whether it makes sense to extend `Functional` to functionals which evaluate to a vector (of length `dim`) in the same way that linear and bilinear forms can refer to vectorial function spaces.  For example, the aerodynamic force per unit span on an aerofoil bounded by a contour *c* and immersed in an inviscid fluid of pressure *p* is &minus;&oint; <sub><i>c</i></sub> *p* d **n**, which could be written
```python
@Functional
def force(w):
   return -w.n * w['p']
```
but that does something a bit unexpected at the moment.

Of course it's possible to define a `Functional` for each Cartesian component, but I wonder whether the generalization is rational and might involve only a minor change.

gdmcbain:
Fixes #520 .

Is this harmless enough?  

It don't think it would work for tensorial functionals, but I haven't thought of an example of one of those yet.

### DIFF
--- a/skfem/assembly/form/functional.py
+++ b/skfem/assembly/form/functional.py
@@ -35,4 +35,4 @@ class Functional(Form):
                  **kwargs) -> float:
         assert vbasis is None
         vbasis = ubasis
-        return np.sum(self.elemental(vbasis, **kwargs))
+        return sum(self.elemental(vbasis, **kwargs).T)
--- a/skfem/assembly/form/functional.py
+++ b/skfem/assembly/form/functional.py
@@ -21,7 +21,7 @@ class Functional(Form):
                 dx: ndarray) -> ndarray:
         if self.form is None:
             raise Exception("Form function handle not defined.")
-        return np.sum(self.form(w) * dx, axis=1)
+        return (self.form(w) * dx).sum(axis=-1)
 
     def elemental(self,
                   v: Basis,
@@ -35,4 +35,4 @@ class Functional(Form):
                  **kwargs) -> float:
         assert vbasis is None
         vbasis = ubasis
-        return sum(self.elemental(vbasis, **kwargs).T)
+        return self.elemental(vbasis, **kwargs).sum(axis=-1)
--- a/skfem/assembly/form/functional.py
+++ b/skfem/assembly/form/functional.py
@@ -1,7 +1,6 @@
 from typing import Dict, Optional
 
 from numpy import ndarray
-import numpy as np
 
 from .form import Form, FormDict
 from ..basis import Basis
@@ -21,7 +20,7 @@ class Functional(Form):
                 dx: ndarray) -> ndarray:
         if self.form is None:
             raise Exception("Form function handle not defined.")
-        return (self.form(w) * dx).sum(axis=-1)
+        return (self.form(w) * dx).sum(-1)
 
     def elemental(self,
                   v: Basis,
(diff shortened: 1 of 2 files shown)


kinnala:
Can we add a test which makes it clear how it should work?

kinnala:
I'm a bit worried about the performance implications if there is a large amount of elements. It might be nothing but that should be checked out.

gdmcbain:
> Can we add a test which makes it clear how it should work?

Good idea.  How about: Consider a vessel filled with an incompressible fluid of density *&rho;* in a uniform gravitational field &minus; *g* **j**, open at the top, *y* = *y*<sub>max</sub>.  The hydrostatic pressure is *p* = *&rho;g* (*y*<sub>max</sub> &minus; *y*) and the weight per unit span on the vessel is either the boundary integral &oint;<sub>&part; *&Omega;*</sub> *p* d **n** = *&rho;g* &oint;<sub>&part; *&Omega;*</sub>  (*y*<sub>max</sub> &minus; *y*) d **n** or the domain integral &minus; *&rho;g* ∬ <sub>*&Omega;*</sub> d *A* **j** = &minus; *&rho;gA* **j**, where *A* is the volume per unit span.

More simply: test that &oint;<sub>&part; *&Omega;*</sub> *y* d**n** = *A* **j**.  I'll code that up shortly.  (I still haven't thought of an example of a second-order tensorial functional, but I think the concept makes sense too.)

gdmcbain:
> I'm a bit worried about the performance implications if there is a large amount of elements. It might be nothing but that should be checked out.

O. K.

I wonder whether using `numpy.sum` and defining `axis` according to the shape of the summand would be more efficient.  That might also allow tensorial functionals of arbitrary tensorial order.

gdmcbain:
Actually yes, using `.sum(axis=-1)` is always the right thing to do, for scalar, vectorial, or tensorial functionals.

I also noticed that it should be `axis=-1` in `Functional._kernel` rather than `axis=1`, which would only sum over the quadrature absciss&aelig; for scalar functionals.

gdmcbain:
On the performance, I'm having difficulty measuring `Functional.assemble`; it's just too fast.  The summand of `Functional._kernel` has shape `(..., vbasis.nelems, len(vbasis.W))`, where the leading dimensions are none for scalar, `vbasis.mesh.dim()` for vector, &c.  The summand of `assemble` is `(..., vbasis.nelems)`.  Transposing `np.ones((2, 2**20))`, say, doesn't take very long; nor does `np.ones(2**20).sum(-1)`, or `np.ones((2, 2**20)).sum(-1)`, or `np.ones((2, 2, 2**20)).sum(-1)`.   (I'm assuming Python isn't smart enough to take short cuts here?)

gdmcbain:
With something like:
```python
from skfem import *

import numpy as np

import logging
from pathlib import Path

logging.basicConfig(
    format="%(levelname)s %(asctime)s %(message)s",
    filename=Path(__file__).with_suffix(".log"),
    level=logging.INFO,
)

@Functional
def measure(w):
    return np.ones_like(w.x[0])

repeats = 2**6

mesh = MeshTri.init_tensor(*(np.arange(2**11),)*2)
ib = InteriorBasis(mesh, ElementTriP1())

elemental = measure.elemental(ib)

logging.info(elemental.shape)

for _ in range(repeats):
    st = sum(elemental.T)

logging.info(st)

for _ in range(repeats):
    s1 = elemental.sum(-1)

logging.info(s1)
```
The latter, `elemental.sum(-1)`, is much faster:
```
INFO 2020-12-02 15:48:44,056 (8380418,)
INFO 2020-12-02 15:50:15,727 4190209.0
INFO 2020-12-02 15:50:16,496 4190209.0
```
That's the implementation in the latest version of this PR.

kinnala:
Looks good. I'll add a changelog entry myself.

PR #524, opened 2020-12-10
Add tests for adaptive refinement

kinnala:


kinnala:
I figured we need something other than the one example, so I added these sanity checks. In both we refine a "corner element" only and check that the number of elements behave as it should.

Issue #411, opened 2020-06-28
wedge cells and elements


gdmcbain:
Would it be worth adding wedge cells and elements? I recently received a mesh of these (too big to attach here, but I think a simple one could be constructed by extruding a triangular mesh in the third dimension).

Four isoparametric wedge elements with 6, 15, 9, and 12 nodes, respectively, are listed in

* Kardestuncer, H. (ed.) (1988). *Finite Element Handbook*. New York: McGraw-Hill, p. 2.130

gdmcbain:
A difficulty: the facets of a wedge aren't all of the same topological type, there are three quadrilaterals and two triangles.  (I don't know the name for the subclass of polyhedra whose facets are the same kind of polygon, like tetrahedra and hexahedra, or for its complement.)  This complicates the `Mesh.facets` attribute; it can't be simple a rectangular array.  (Unless padded with a sentinel value like the &minus;1 in  `Mesh.f2t` for facets on the boundary?)

gdmcbain:
Ern & Guermond (2004, &sect;1.2.5) refer to wedge elements as &lsquo;prismatic Lagrange finite elements&rsquo; and (table 1.3) depict ℙℝ<sub><i>k</i></sub> for _k_ = 1, 2, and 3.

* Ern, A., Guermond, J.-L. (2004). _Theory and Practice of Finite Elements_. Springer, &sect;1.2.5 &lsquo;Prismatic Lagrange finite elements&rsquo;

gdmcbain:
Zienkiewicz & Taylor (2000, &sect;8.13) call them &lsquo;triangular prism elements&rsquo; and draw the 6, 15, and 26-node &lsquo;serendipity&rsquo; elements (figure. 8.23).

*  Zienkiewicz, O. C., Taylor, R. L. (2000). _The Finite Element Method. I: The Basis_. Butterworth-Heinemann.

gdmcbain:
There's some discussion of &lsquo;vtkWedge&rsquo; and Wedge6 in nschloe/meshio#863.

kinnala:
I've been thinking about this every now and then and still a proper solution escapes me. If only `CellBasis` assembly is needed for nodal elements then I think this can be implemented by hacking `Mesh.facets`, essentially disabling it.

gdmcbain:
[...]

PR #530, opened 2020-12-23
FacetBasis.trace()

### LINKED ISSUE #514: export trace
This is a vague feature request; vague because I'm still thinking about how best to formulate the problem, but let's see if I can already make sense.

Say in ex02, the square plate subject to normal pressure, clamped on the left and pinned right and top, that I wanted to export the deflexion `x` along the `{'bottom': m.facets_satisfying(lambda x: x[1] == 0)}`.  Say I wanted to couple the scikit-fem finite element model to something external.  Say, for a definite example, I had a soap-film (&Delta; *w* = 0) pinned to the bottom of the plate `x[1] == 0` and pinned (*w* = 0) around the adjacent square `0 < x[0] < 1, 1 < x[1] < 0` with boundary condition on `x[1] == 0` prescribed by the deflexion of the plate.  That's one-way coupling, but also in other cases two-way coupling might be of interest, and we might use a Lagrange multiplier as [suggested](https://github.com/kinnala/scikit-fem/issues/163#issuecomment-717019908) for coupling subdomains #163. 

I can extract the dofs with

```python
dofs['bottom'] = ib.find_dofs({'bottom': m.facets_satisfying(lambda x: x[1] == 0)})['bottom']
np.vstack([ib.doflocs[0, dofs['bottom'].all()], x[dofs['bottom'].all()]])
```

or the ordinates
```python
ib.interpolator(x)(ib.doflocs[:, dofs['bottom'].all()])
```
(Here the dofs aren't the same as the ordinates: the last eight of `dofs['bottom'].all()` are the `dofs['bottom'].facet['u_n']` since `x` is built with `ElementTriMorley`.)

But how is the connectivity of `dofs['bottom']` to be extracted and exported?

Does it make sense to export some kind of facet mesh corresponding to the `FacetBasis`?
```python
bb = FacetBasis(ib.mesh, ib.elem, facets=m.facets_satisfying(lambda x: x[1] == 0))
```
on which the dofs from the `ib` are also dofs?  (I'm not sure what this means when `isinstance(ib.elem, ElementTriMorley)`.)

Or if the external model doesn't have Morley elements, does it make sense to project the trace of `x` along `'bottom'` onto some other facet finite element space, e.g. facets with domain element `ElementTriP2` or `ElementTriP1` or `ElementTriP0`?  Or onto a lower (`dim = ib.elem.dim - 1`) dimensional finite element space defined on the part of the boundary; e.g. `ElementLineP0` #508, `ElementLineP1`, or `ElementLineP2`?  If the deflexion of the plate were projected onto `ElementTriP2` is there a sense in which the `FacetBasis` is like `ElementLineP2`?

So obviously it can all be done by interpolation (`ib.interpolator`) and reinterpolation on the other side, but are there more direct ways, e.g. involving projection?

Some related issues: #150, #158, #163, #261, #271.

kinnala:
This aims to address #514 by adding a helper method to `FacetBasis`. `FacetBasis.trace` will create a trace mesh and restrict a solution vector to that trace mesh. Optionally it performs L2 projection onto a new finite element basis on the trace mesh. 

Because we do not have `Mesh` classes defined for embedded meshes (surfaces etc.), `FacetBasis.trace` will return a pair `(p, t)`, i.e. an array of nodes and their connectivity. It is up to the user to, e.g., save them to an external format or project them to lower dimensional space so that we can initialize one of the existing `Mesh` classes.

Here is an usage example:
```python
from skfem import *
from skfem.visuals.matplotlib import *
m = MeshQuad()
m.refine(3)
basis = FacetBasis(m, ElementQuad1(), facets=m.facets_satisfying(lambda x: x[1]==0.0))
p, t, y = basis.trace(m.p[0], ElementQuad0())
plot(InteriorBasis(MeshLine(p[0], t), ElementLineP0()), y)
show()
```

Fixes #514


kinnala:
@gdmcbain Do you feel it addresses #514 ?

gdmcbain:
Oh, no.  Sorry, I don't think I'll be able to review this before next year.  It looks excellent, reading through the code.  I'm very pleased at how concise it is; that's a good testament to the design of what's gone before, I think.  I won't have time to test it before the break.

Would it be cleaner to omit the optional preliminary projection?  The user can just do that themselves if they need to, can't they?  Actually I envisage projecting afterwards, which should be much cheaper since the trace is smaller than the domain.  In the motivating application, I solve the Navier&ndash;Stokes equation with quadrilateral Taylor&ndash;Hood elements and want to (i) extract the velocity along a part of the boundary, (ii) project it to piecewise constant for use as an inlet velocity boundary condition in a finite volume code (OpenFOAM).  Projecting the entire field to ElementVectorH1(ElementQuad0) seems wasteful, but this might be premature optimisation.

kinnala:
> Would it be cleaner to omit the optional preliminary projection? The user can just do that themselves if they need to, can't they?

Sure, but note that many times there is no trivially corresponding boundary element, e.g., in the `ElementTriMorley` example, and then projection may be required.

> Actually I envisage projecting afterwards, which should be much cheaper since the trace is smaller than the domain.

If you read carefully, the L2 projection is performed only on the boundary.

> In the motivating application, I solve the Navier–Stokes equation with quadrilateral Taylor–Hood elements and want to (i) extract the velocity along a part of the boundary, (ii) project it to piecewise constant for use as an inlet velocity boundary condition in a finite volume code (OpenFOAM). Projecting the entire field to ElementVectorH1(ElementQuad0) seems wasteful, but this might be premature optimisation.

See above.

kinnala:
In principle, each `Element` could carry a reference to the "best possible" boundary element, e.g., `ElementTriMorley` would reference `ElementTriDG(ElementTriP2())` which would transform to something like `ElementLineDG(ElementLineP2())` (I think?). Thus, one wouldn't be required to supply this optional argument even in complicated cases. What do you think about that?

gdmcbain:
Thanks for the further comment on Morley.  (Actually the use of Morley in the example was accidental; it was just that ex02 was the first example with a nonzero trace and it uses Morley.)  I was puzzled at the idea that &lsquo;there is no trivially corresponding boundary element&rsquo; to the Morley space but am not really too familiar with the element.  My successful use of it in ex18 was just by analogy: it works for the clamped plate therefore it works for the creeping streamfunction since they're governed by the same biharmonic equation and boundary conditions&mdash;but those do involve the trace&hellip;  So I went looking for a reference and found [a very good one](https://scicomp.stackexchange.com/questions/34553/morley-element-implementation-reference/34568#34568) on scicomp.stackexchange.com.   Thanks!

I do like the idea that

> each Element could carry a reference to the "best possible" boundary element,

or just a `NotImplemented` for exotic cases, either to be filled in or avoided by preliminary projection.

And on 

> If you read carefully, the L2 projection is performed only on the boundary.

Sorry, yes, this is indeed obvious.

I'll get back to this properly in the new year but thanks again meanwhile for the explanations of Morley.

kinnala:
I concluded that this will not work if the provided `elem` has more than one interior DOF. In such a case interpolation/linear combination of basis functions is required to restrict to the boundary.

skfem/element/element_hex/element_hex0.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+
+from ..element_h1 import ElementH1
+from ...mesh.mesh3d import MeshHex
+
+
+class ElementHex0(ElementH1):
```
kinnala, line 7: TODO: add to changelog

kinnala:
Now `FacetBasis.trace` will return `InteriorBasis` and the projected solution vector. It requires that the user provides a function which defines how the points on the boundary of a `d`-dimensional mesh are projected onto `d-1`-dimensional space.

skfem/assembly/basis/facet_basis.py
```diff
@@ -108,10 +115,102 @@ def default_parameters(self):
...
+        y = project(x, self, fbasis, I=I)
+
+        # build connectivity for lower dimensional mesh
+        ix = self.mesh.facets[:, self.find]
+        ixuniq = np.unique(ix)
+        b = np.zeros(np.max(ix) + 1, dtype=np.int64)
+        b[ixuniq] = np.arange(len(ixuniq), dtype=np.int64)
+
+        return InteriorBasis(
+            target_meshcls(projection(self.mesh.p[:, ixuniq]), b[ix]),
```
kinnala, line 214: Now the questions remains: is this sufficient or should we support more general use cases, e.g., sometimes the user might actually not want to initialize `InteriorBasis` or even `target_meshcls` but simply feed `self.mesh.p[:, ixuniq]` and `b[ix]` to some external tool together with `y`.

kinnala: One solution would be to split everything else but the return statement into a separate method which can be called for different use cases. I'll try that.

kinnala: Now it should be flexible enough for different use cases (by direct use of `FacetBasis._trace`) and reasonable defaults (in `FacetBasis.trace`).

gdmcbain:
O. K.  The example from [2020-12-23](https:/https://github.com/kinnala/scikit-fem/pull/530#issue-544759340/github.com/kinnala/scikit-fem/pull/530#issue-544759340) is modified to:
```python
from skfem import *
from skfem.visuals.matplotlib import *

m = MeshQuad()
m.refine(3)
basis = FacetBasis(m, ElementQuad1(), facets=m.facets_satisfying(lambda x: x[1] == 0.0))
tb, y = basis.trace(m.p[0], lambda x: x[0])
plot(tb, y)
show()
```
and produces 
trace

which looks good.  I'll try it out on the motivating OpenFOAM example and report back.  I won't worry about ex02 as it's not immediately obvious what to do with Morley elements and they're not of immediate interest.

gdmcbain:
I'm getting some discrepancies going from `ElementQuad2` to `ElementQuad0` using this new `.trace` compared to the explicit code that I had previously; I assume that the bug is in the code on my side, but I'll continue investigating.

kinnala:
What kind of discrepancy? The trace values are different?

gdmcbain:
Yes, slightly different values, about 10%, which is weird.  Sorry for such an incomplete report. That was just on closing yesterday. I'm sure the error is on my side but today I'll craft a minimal working example; yesterday's was part of an extended version of ex27, 660 lines of code with lots of irrelevance.

gdmcbain:
Sorry, I haven't been able to reproduce the discrepancy in an MWE.  I'm even more sure that the error isn't in the new `FacetBasis.trace` code but haven't quite been able to pin it down.

The MWE is something like
```python
from skfem import *

import numpy as np


def f(y: np.ndarray) -> np.ndarray:
    return y * (1 - y)


mesh = MeshQuad.init_tensor(*np.tile(np.linspace(0, 1, 9), [2, 1]))
mesh.define_boundary("left", lambda x: x[0] == 0)
ib = InteriorBasis(mesh, ElementQuad2())
fb = FacetBasis(ib.mesh, ib.elem, facets=mesh.boundaries["left"])

y = fb.doflocs[1, fb.get_dofs(fb.find).nodal['u']]
left_mesh = MeshLine(y)
left_basis = InteriorBasis(left_mesh, ElementLineP0())

u_old = project(lambda y: f(y[0]), basis_to=left_basis)

_, u_new = fb.trace(lambda x: f(x[1]), lambda x: x[1])
print(np.linalg.norm(u_new - u_old))
```
except that that works perfectly!

I'm pretty sure the PR is fine.  Sorry for this terrible bug report!

gdmcbain:
O. K., I've finally found the cause of the discrepancy and as expected it was completely in the extraneous code.  (I had inferred the limits of the patch of boundary from the coordinates of the points passed to `fun` in `skfem.utils.project`; wrong because they're not doflocs but quadrature points and usually don't include the termini!)  Easily fixed in the extraneous code using, e.g., 
```python
mesh.p[:, mesh.facets[:, mesh.boundaries["patch"]]]
```
or `basis.doflocs`.

I'll now look at the actual PR.

skfem/assembly/basis/facet_basis.py
```diff
@@ -108,10 +115,110 @@ def default_parameters(self):
...
+            MeshQuad: ElementLineP0,
+            MeshTet: ElementTriP0,
+            MeshHex: ElementQuad0,
+        }
+
+        meshcls = type(self.mesh)
+        if meshcls not in DEFAULT_TARGET:
+            raise NotImplementedError("Mesh type not supported.")
+        if target_elem is None:
+            target_elem = DEFAULT_TARGET[meshcls]()
```
gdmcbain, line 197: Would it be more Pythonic to have the `DEFAULT_TARGET` as an attribute or property of `self.mesh`?  And EAFP over LBYL?  Define as `NotImplemented` or similar for `Mesh` then override in subclasses?  The resulting default error message should be almost as obvious.

Could the key use `self.mesh.brefdom`? 
```python
{'line': ElementLineP0, 'tri': ElementTriP0, 'quad': ElementQuad0}
```
(and worry about wedges #411 &c. later&hellip;)

skfem/assembly/basis/facet_basis.py
```diff
@@ -108,10 +115,110 @@ def default_parameters(self):
...
+            (ElementLineP1, MeshQuad): (ElementQuad1, MeshLine),
+            (ElementLineP2, MeshQuad): (ElementQuad2, MeshLine),
+            (ElementTriP0, MeshTet): (ElementTetP0, MeshTri),
+            (ElementTriP1, MeshTet): (ElementTetP1, MeshTri),
+            (ElementTriP2, MeshTet): (ElementTetP2, MeshTri),
+            (ElementQuad0, MeshHex): (ElementHex0, MeshQuad),
+            (ElementQuad1, MeshHex): (ElementHex1, MeshQuad),
+            (ElementQuad2, MeshHex): (ElementHex2, MeshQuad),
+        }
+
```
gdmcbain, line 213: Is the second term of the value redundant?
```python
all(k[0].mesh_type is v[1] for k, v in TRACE_RESTRICT_MAP.items())
```

gdmcbain (approved):

skfem/assembly/basis/facet_basis.py
```diff
@@ -108,10 +115,110 @@ def default_parameters(self):
...
+        y = project(x, self, fbasis, I=I)
+
+        # build connectivity for lower dimensional mesh
+        ix = self.mesh.facets[:, self.find]
+        ixuniq = np.unique(ix)
+        b = np.zeros(np.max(ix) + 1, dtype=np.int64)
+        b[ixuniq] = np.arange(len(ixuniq), dtype=np.int64)
+
+        return self.mesh.p[:, ixuniq], b[ix], y
+
```
gdmcbain, line 152: `_trace` looks as though it could be split into two functions, one returning the points and cells and the other returning the values. The parts of the body either side of the &lsquo;build connectivity&rsquo; comment are independent.

gdmcbain (approved):

skfem/assembly/basis/facet_basis.py
gdmcbain: Yep.

```python
 def _trace_points_cells(self) -> Tuple[ndarray, ndarray, ndarray]:
        """Trace mesh reindexing and basis projection."""

        # build connectivity for lower dimensional mesh
        ix = self.mesh.facets[:, self.find]
        ixuniq = np.unique(ix)
        b = np.zeros(np.max(ix) + 1, dtype=np.int64)
        b[ixuniq] = np.arange(len(ixuniq), dtype=np.int64)

        return self.mesh.p[:, ixuniq], b[ix]

    def _trace_values(
        self, x: ndarray, elem: Element
    ) -> Tuple[ndarray, ndarray, ndarray]:
        """Trace mesh reindexing and basis projection."""
        from skfem.utils import project

        fbasis = FacetBasis(
            self.mesh, elem, facets=self.find, quadrature=(self.X, self.W)
        )
        I = fbasis.get_dofs(self.find).all()
        if len(I) == 0:  # special case: no facet DOFs
            if fbasis.dofs.interior_dofs.shape[0] > 1:
                # no one-to-one restriction: requires interpolation
                raise NotImplementedError
            # special case: piecewise constant elem
            I = fbasis.dofs.interior_dofs[:, self.tind].flatten()
        y = project(x, self, fbasis, I=I)

        return y

    def _trace(self, x: ndarray, elem: Element) -> Tuple[ndarray, ndarray, ndarray]:
        """Trace mesh reindexing and basis projection."""
        from skfem.utils import project

        return *self._trace_points_cells(), self._trace_values(x, elem)
```

Not to say that that's how it should be written, just to demonstrate the factorization.

gdmcbain: Apart from the use of `self.find`, `_trace_points_cells` doesn't have too much to do with the trace.  It sort of looks like some kind of submesh function, if it were passed `ix`.  I wonder whether such a function (or rather method of class Mesh) would be useful for anything else.  Untested:

```python
def submesh(self: Mesh, ix: ndarray) -> Tuple[ndarray, ndarray]:
    ixuniq = np.unique(ix)
    b = np.zeros(np.max(ix) + 1, dtype=np.int64)
    b[ixuniq] = np.arange(len(ixuniq), dtype=np.int64)
   
    return self.p[:, ixuniq], b[ix]
```

kinnala: It seems that similar logic to `submesh` is used at least in `Mesh.remove_elements`. Thus, it could make sense to have it in `Mesh`. The same function could work for subsets of elements, facets or edges. I'll look into refactoring it.

skfem/assembly/basis/facet_basis.py
kinnala: Sure, thanks. I'll fix it.

PR #537, opened 2021-01-11
WIP: method-chaining for Mesh

gdmcbain:
In partial fulfillment of #536.

PR #538, opened 2021-01-11
Method chaining

### LINKED ISSUE #536: Support method chaining in Mesh
A proposition:
- [x] return `self` in `Mesh.scale`, `Mesh.translate`, and `Mesh.refine` to support chaining them
- [x] deprecate `Mesh.remove_elements` and implement `Mesh.prune` with similar functionality which returns `self`
- [ ] deprecate `Mesh.mirror`
- [ ] implement `Mesh.flip_vertical` and `Mesh.flip_horizontal` that return `self`

kinnala:
Fixes #536. Rebased #537 onto master to make the web interface visualize the changes better.

### DIFF
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -28,12 +28,15 @@ class Mesh:
         to a point.
     t
         The element connectivity (`dim x Nelements`). Each column corresponds
-        to a element and contains column indices to ``self.p``.
+        to an element and contains column indices to ``self.p``.
     subdomains
-        Named subsets of elements. Empty if not loaded from an external format.
+        Named subsets of elements. Can be provided to the constructor or loaded
+        from an external format. Otherwise the value is ``None``.
     boundaries
-        Named subsets of boundary facets. Empty if not loaded from an external
-        format or defined via :meth:`~skfem.mesh.Mesh.define_boundary`.
+        Named subsets of boundary facets. Can be provided to the constructor as
+        a keyword argument, loaded from an external format or defined via
+        :meth:`~skfem.mesh.Mesh.define_boundary`. Otherwise the value is
+        ``None``.
 
     """
 
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -130,7 +130,7 @@ class Mesh:
                                   "for this mesh type!")
 
     def refine(self: MeshType,
-               arg: Optional[Union[int, ndarray]] = None) -> MeshType:
+               arg: Optional[Union[int, ndarray]] = None):
         """Refine the mesh.
 
         Parameters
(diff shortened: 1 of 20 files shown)


skfem/mesh/mesh.py
```diff
@@ -24,11 +24,11 @@ class Mesh:
...
-        The vertices of the mesh (dim x Nvertices). Each column corresponds to
-        a point.
+        The vertices of the mesh (`dim x Nvertices`). Each column corresponds
+        to a point.
     t
-        The element connectivity (dim x Nelements). Each column corresponds to
-        a element and contains column indices to `self.p`.
+        The element connectivity (`dim x Nelements`). Each column corresponds
+        to a element and contains column indices to ``self.p``.
     subdomains
```
gdmcbain, line 32: a element -> an element

kinnala:
I'm still wondering `Mesh2D.mirror` which doesn't seem to be used anywhere and it's also a bit unexpected because it combines a reflection and joining of two meshes. I think I'll deprecate it and implement `Mesh.reflect` which works for any dimension.

kinnala:
I noticed that the support for subdomains/boundaries in `Mesh.__add__` was actually not working. For now decided to remove it and add a warning if subdomains/boundaries get discarded. Will fix this in future when revamping the internal data structure for subdomains/boundaries.

skfem/mesh/mesh.py
```diff
@@ -147,50 +152,45 @@ def refine(self, arg: Optional[Union[int, ndarray]] = None):
             self._adaptive_refine(arg)
         else:
             raise NotImplementedError("The parameter type not supported.")
+        return self
```
gdmcbain, line 155: Hmm, following the discussion in #536, I wonder whether having `Mesh.refine` 
* mutate `self` *and* 
* return the mutated self 

could cause just the kind of confusion that I wish to avoid with mutation in general; i.e. if the only usage of `Mesh.refine` is
```python
m.refine()
```
then it's pretty clear that it's mutating but if I see
```python
m_new = m_old.refine()
```
then I might forget that it's impure and forget about the mutation and get surprised if I re-access `m_old` and find it changed.  (Maybe in some kind of manual multigrid?)

Could this be why the Python Standard Library offers the pure function [`sorted()`](https://docs.python.org/3/library/functions.html#sorted) and the in-place method [`list.sort()`](https://docs.python.org/3/library/stdtypes.html#list.sort)?  Should the in-place method `Mesh.refine` be retained (for full backward-compatibility) and a pure method `Mesh.refined` be introduced?   (If so, in due course, the examples, docs, tests, &c., might migrate to the latter, with the former perhaps eventually being scheduled for deprecation.)

(In the case of `refine`, [I don't think](https://github.com/kinnala/scikit-fem/issues/536#issuecomment-758351972) it gains any efficiency by working in place so it might well be redefined in terms of the proposed pure `Mesh.refined`.)

kinnala: I kind of like this naming: `refine` vs. `refined`, `scale` vs. `scaled`, `translate` vs. `translated`. I'll look into it after I stop being busy with a peer review appointment.

kinnala: For now I created `refined` basically as
```python
mesh = self.copy()
mesh.refine()
return mesh
```
because we still aim to be backwards-compatible.
Later when we are ready to drop `Mesh.refine` we can refactor it differently.

kinnala:
Decided to keep the name `Mesh.remove_elements` because it is more descriptive. The initial idea was only to have a shorter variant but now it doesn't make much sense.

Issue #548, opened 2021-02-03
vrot broken


gdmcbain:
Running
```shell
python docs/examples/ex20.py
```
 raises
```
psi0 = 0.015846568093739626 (cf. exact = 1/64 = 0.015625)
Traceback (most recent call last):
  File "/home/gmcbain/src/scikit-fem/docs/examples/ex20.py", line 71, in <module>
    velocity = asm(
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/skfem/assembly/__init__.py", line 62, in asm
    return form.assemble(*args, **kwargs)
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/skfem/assembly/form/linear_form.py", line 40, in assemble
    data[ixs] = self._kernel(vbasis.basis[i], w, dx)
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/skfem/assembly/form/linear_form.py", line 45, in _kernel
    return np.sum(self.form(*v, w) * dx, axis=1)
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/skfem/models/general.py", line 19, in vrot
    return dot(v, curl(w.w))
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/skfem/helpers.py", line 31, in curl
    if u.curl is not None:
AttributeError: 'numpy.ndarray' object has no attribute 'curl'
```

This is missed by the tests since it's only run `if __name__ == "__main__"`.

I think the problem is that `w.w` is an `ndarray` whereas we were expecting a `DiscreteField` like `w["w"]`.

gdmcbain:
A simple fix is in #549, but it's a bit subtle.  On first looking at `skfem.models.general.vrot` and seeing `w.w`, I had thought that the same fix would be required for `rot`

https://github.com/kinnala/scikit-fem/blob/2957f0443be445d975c0299b9bef02ed4e2f5886/skfem/models/general.py#L12-L14

but no, because it doesn't have its curl taken there, so exx 18, 27, 30 run fine.

I see that 

https://github.com/kinnala/scikit-fem/blob/2957f0443be445d975c0299b9bef02ed4e2f5886/skfem/helpers.py#L29-L36

is expecting a `DiscreteField`.  Should `w.w` be one?

PR #549, opened 2021-02-03
access w as an item DiscreteField not an attribute ndarray

gdmcbain:
For #548.

### DIFF
--- a/skfem/models/general.py
+++ b/skfem/models/general.py
@@ -16,4 +16,4 @@ def rot(v, w):
 
 @LinearForm
 def vrot(v, w):
-    return dot(v, curl(w.w))
+    return dot(v, curl(w["w"]))
--- a/skfem/models/general.py
+++ b/skfem/models/general.py
@@ -11,7 +11,7 @@ def divergence(u, v, w):
 
 @LinearForm
 def rot(v, w):
-    return dot(curl(v), w.w)
+    return dot(curl(v), w["w"])
 
 
 @LinearForm


gdmcbain:
I see that the item-style access is consistent with

https://github.com/kinnala/scikit-fem/blob/2957f0443be445d975c0299b9bef02ed4e2f5886/docs/examples/ex10.py#L14-L23

https://github.com/kinnala/scikit-fem/blob/2957f0443be445d975c0299b9bef02ed4e2f5886/docs/examples/ex13.py#L54-L57

https://github.com/kinnala/scikit-fem/blob/2957f0443be445d975c0299b9bef02ed4e2f5886/docs/examples/ex22.py#L47-L54

skfem/models/general.py
```diff
@@ -11,9 +11,9 @@ def divergence(u, v, w):
 
 @LinearForm
 def rot(v, w):
-    return dot(curl(v), w.w)
+    return dot(curl(v), w["w"])
```
kinnala, line 14: Now that we have the option to use pretty much any name in place of `"w"` (I belive when this was originally implemented, it was only `"w"` allowed), the choice of `"w"` feels rather arbitrary. I wonder if we could improve these forms somehow by keeping that in mind?

kinnala: See #550

gdmcbain: I see, yes, that's a good idea, a less arbitrary name than `"w"`.  But then how is the user to know or remember it?  Perhaps better still is indeed the [tacit](https://en.wikipedia.org/wiki/Tacit_programming) style enabled by #550. Does that mean deprecating `rot` and `vrot`?  I suppose that if they're left in `skfem.models.general` at least temporarily during deprecation then the attribute &rarr; item fix is still required.

I think `rot` needs a different `BilinearForm` to `vrot`.

gdmcbain:
This is already covered by e19e5810771b00c9145b69b21715ac0fe267372e, though the [ideas](https://github.com/kinnala/scikit-fem/pull/549#discussion_r569272259) of either coming up with a better name for `"w"` or avoiding having to name it are worth considering.

PR #550, opened 2021-02-03
Add Basis.with_element and Form.partial

### LINKED ISSUE #548: vrot broken

kinnala:
@gdmcbain I'm proposing the addition of `Form.partial` which uses `functools.partial` on the actual form function and, thus, allows specializing them for different use cases. I also added `Basis.with_element` because it is a common pattern to initialize the same `Basis` but using different element.

Thus, if you have form such as
```python
@BilinearForm
def curluv(u, v, w):
    return dot(curl(u), v)
```
you can make it equivalent to
```python
@LinearForm
def rot(v, w):
    return dot(curl(w['w']), v)
```
by first casting it to `LinearForm` and then applying `Form.partial`:
```python
LinearForm(curluv).partial(basis.interpolate(w))
```

Also fixes #548.

Remaining work:
- [x] Add changelog entry

### DIFF
--- a/skfem/assembly/basis/interior_basis.py
+++ b/skfem/assembly/basis/interior_basis.py
@@ -144,3 +144,13 @@ class InteriorBasis(Basis):
             return w
 
         return interpfun
+
+    def with_element(self, elem: Element) -> 'InteriorBasis':
+
+        return type(self)(
+            self.mesh,
+            elem,
+            mapping=self.mapping,
+            quadrature=self.quadrature,
+            elements=self.tind,
+        )
--- a/README.md
+++ b/README.md
@@ -214,6 +214,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added: `InteriorFacetBasis` for integrating over the interior facets, e.g.,
   evaluating error estimators with jumps and implementing DG methods.
 - Added: `MortarFacetBasis` for integrating over the mortar mesh.
+- Added: `InteriorBasis.with_element` for reinitializing an equivalent basis
+  that uses a different element.
+- Added: `Form.partial` for applying `functools.partial` to the form function
+  wrapped by `Form`.
 - Fixed: Include explicit Python 3.9 support.
 
 ### [2.4.0] - 2021-01-20
--- a/docs/examples/ex18.py
+++ b/docs/examples/ex18.py
@@ -78,8 +78,7 @@ uvp = solve(*condense(K, f, D=basis['u'].find_dofs()))
 
 velocity, pressure = np.split(uvp, [A.shape[0]])
 
-basis['psi'] = InteriorBasis(mesh, ElementTriP2(),
-                             quadrature=basis['u'].quadrature)
+basis['psi'] = basis['u'].with_element(ElementTriP2())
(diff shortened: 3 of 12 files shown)


skfem/assembly/basis/interior_basis.py
```diff
@@ -144,3 +144,13 @@ def interpfun(x):
...
+
+    def with_element(self, elem: Element) -> 'InteriorBasis':
+
+        return type(self)(
+            self.mesh,
+            elem,
+            mapping=self.mapping,
+            quadrature=self.quadrature,
+            elements=self.tind,
+        )
```
gdmcbain, line 156: Were `InteriorBasis` a [`dataclass`](https://docs.python.org/3/library/dataclasses.html?highlight=dataclass#dataclasses.dataclass) (python>=3.7), this would be handled by [`replace`](https://docs.python.org/3/library/dataclasses.html?highlight=dataclass#dataclasses.replace): 
```python
        return replace(self, elem=elem)
```

kinnala: Will that also recompute `self.basis`? I thought `dataclass` is for those cases where you have only a trivial constructor. However, the main point of `Basis` is to cache this line: https://github.com/kinnala/scikit-fem/blob/master/skfem/assembly/basis/interior_basis.py#L67

kinnala: Would it be possible to not have `self.basis = ...` in constructor, but rather as a `property` which is cached and then invalidated during a call to `replace`?

kinnala: I really need to look into `dataclass` soon because Python 3.6 EOL is coming near and it becomes a viable option for us.

PR #554, opened 2021-02-04
Splitquads to meshtri

gdmcbain:
More work on using preferred new names instead of old ones, even if `MeshQuad._splitquads` isn't officially deprecated.


skfem/visuals/matplotlib.py
```diff
@@ -214,10 +214,9 @@ def plot_meshquad(m: MeshQuad, z, **kwargs):
 
     """
     if len(z) == m.t.shape[-1]:
-        m, z = m._splitquads(z)
+        return plot(*m.to_meshtri(z), **kwargs)
     else:
-        m = m._splitquads()
-    return plot(m, z, **kwargs)
+        return plot(m.to_meshtri(), z, **kargs)
```
kinnala, line 219: A typo here I suppose?

gdmcbain: Indeed.  Thanks.  The CI exceptions were all about types though rather then typos.  I thought it was saying it didn't like that to_meshtri sometimes returned cell_data as well as `MeshTri`; not that that was introduced here.  Let's see.

kinnala: Yeah, there are obviously several typing-related issues remaining in the codebase but mypy is not always detecting those. Now that numpy 1.20.0 has more typing-related stuff, mypy seems to detect them more often.

gdmcbain: That's good, I guess; I wasn't expecting it here though.  I intended merely a simple `_splitquads` &rarr; `to_meshtri`; apart from that the other changes were just attempts at avoiding type-warnings. 

Perhaps this PR should be closed and the typing worked out first.

Is the CI more strict on code that's changed than code that's already present?

kinnala: No, I'm not aware of any differences regarding stuff that's already present. It's just a run of `mypy skfem`.

gdmcbain: O. K., thanks.  I'll close this one and try again with a fresh start from master.

kinnala: I've noticed that mypy has become now very strict on things like `np.sum(...)` that might result in different outputs depending on the input array shape and flags such as `axis=...`. That was really tedious to fix in #547

kinnala: E.g., I think `solve` does not yet pass `mypy` checks and that's why I introduced `solve_linear` and `solve_eigen` that should pass `mypy` checks.

kinnala: Let me add that if you don't feel like fixing types right now, it's fine to put `# type: ignore` for now.

kinnala: This is more like a long term goal to have it correctly. Of course helps if you have time to look into it right now.

kinnala: Sometimes the issue (like in `solve`) is that the interface has not been really planned for type stability. Theoretically the user can put in a valid set of types that give a complete nonsense output or even Exception. E.g., one can feed in a linear system `Ax=b` but then give `solver` which is meant for eigenproblems.

gdmcbain: Yes, right.  I like type theory, for sure.  And if NumPy is going that way, it definitely makes sense to follow.  I'll have a look around for release-notes on that.  Maybe using type:ignores here just for deprecated alternatives.  Meanwhile, I'll retry the more minimal name-updating which I don't think should change anything type-related. #555

PR #557, opened 2021-02-06
functools.lru_cache

gdmcbain:
This is just a little experiment to see how `functools.lru_cache` looks compared to explicit memoization.  It's not WIP but not I'm not necessarily arguing for its adoption either; is there a symbol like WIP for marking that?

### DIFF
--- a/skfem/mesh/graph.py
+++ b/skfem/mesh/graph.py
@@ -34,38 +34,38 @@ class Graph:
     def nedges(self):
         return self.edges.shape[1]
 
-    @lru_cache(maxsize=1)
     @property
+    @lru_cache(maxsize=1)
     def _facets_and_connectivity(self):
         return Graph.build_entities(self.t, self._facet_indices)
 
-    @lru_cache(maxsize=1)
     @property
+    @lru_cache(maxsize=1)
     def _edges_and_connectivity(self):
         return Graph.build_entities(self.t, self._edge_indices)
 
-    @lru_cache(maxsize=1)
     @property
+    @lru_cache(maxsize=1)
     def facets(self):
         return self._facets_and_connectivity[0]
 
-    @lru_cache(maxsize=1)
     @property
+    @lru_cache(maxsize=1)
     def t2f(self):
         return self._facets_and_connectivity[1]
 
-    @lru_cache(maxsize=1)
     @property
+    @lru_cache(maxsize=1)
     def f2t(self):
         return Graph.build_inverse(self.t, self.t2f)
 
-    @lru_cache(maxsize=1)
     @property
+    @lru_cache(maxsize=1)
     def edges(self):
         return self._edges_and_connectivity[0]
 
-    @lru_cache(maxsize=1)
     @property
+    @lru_cache(maxsize=1)
     def t2e(self):
         return self._edges_and_connectivity[1]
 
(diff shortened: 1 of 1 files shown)


gdmcbain:
Here's the slightly modified and expanded example script.
```python
from skfem.mesh.graph import Graph, Grid
from skfem import *
import numpy as np

m = MeshTri()
m2 = MeshTri2.load("docs/examples/meshes/quadratic_tri.msh")
m3 = MeshTet()
m4 = MeshHex()
m5 = MeshQuad()

M = Grid(m.t, m.p, ElementTriP1())
M2 = Grid(m2.t, m2._mesh.p, ElementTriP2())
M3 = Grid(m3.t, m3.p, ElementTetP1())
M4 = Grid(m4.t, m4.p, ElementHex1())
M5 = Grid(m5.t, m5.p, ElementQuad1())

basis = InteriorBasis(m, ElementTriP1())
Basis = InteriorBasis(M, ElementTriP1())

Basis2 = InteriorBasis(M2, ElementTriP1())
basis2 = InteriorBasis(m2, ElementTriP1())

Basis3 = InteriorBasis(M3, ElementTetP1())
basis3 = InteriorBasis(m3, ElementTetP1())

Basis4 = InteriorBasis(M4, ElementHex1())
basis4 = InteriorBasis(m4, ElementHex1())

Basis5 = InteriorBasis(M5, ElementQuad1())
basis5 = InteriorBasis(m5, ElementQuad1())

from skfem.models.poisson import laplace

A = laplace.assemble(Basis)
a = laplace.assemble(basis)
assert (a-A).nnz == 0

A2 = laplace.assemble(Basis2)
a2 = laplace.assemble(basis2)
assert (a2-A2).sum() < 1e-10

A3 = laplace.assemble(Basis3)
a3 = laplace.assemble(basis3)
assert (a3-A3).nnz == 0

A4 = laplace.assemble(Basis4)
a4 = laplace.assemble(basis4)
assert (a4-A4).nnz == 0

A5 = laplace.assemble(Basis5)
a5 = laplace.assemble(basis5)
assert (a5-A5).nnz == 0
```

gdmcbain:
Oops, forgot that it's no long possible to `pip install -e .`, hence running this locally wasn't using it!

gdmcbain:
Nope, no good.
> TypeError: unhashable type: 'numpy.ndarray'

I thought I'd dealt with similar cases previously; I'll have to check my notes on Monday.

kinnala:
I tried also `lru_cache` but bumped to the same issue. Apparently there are some workarounds but I prefer not to use any extra dependencies.  Would it work by defining custom `__hash__` for `Graph` using some of the ideas in the following url? https://stackoverflow.com/questions/16589791/most-efficient-property-to-hash-for-numpy-array

gdmcbain:
While hashing the array properly is unappealing (i assume that that must take a nontrivial amoubt of work and scale at least like O(*n*), an improper custom hash might do if the connectivity array's never going to be changed. That could be arbitrarily cheap, O(1).

kinnala:
We can safely assume that. So you think the cache is local to the object and not spanning over multiple instances?

gdmcbain:
I hadn't thought of that. Yes, I assumed the caching is per instance rather than per class.

gdmcbain:
Oh, I see. So if there are situations later where we want to use .replace but keep some of the cached return-values?

PR #560, opened 2021-02-13
Add support for visualizing 2D meshes as svg

kinnala:
This PR enables visualizing 2D meshes as SVG without matplotlib and its now used by default in a Jupyter notebook:
Kuvakaappaus - 2021-02-13 15-21-34

### DIFF
--- a/skfem/mesh/mesh2d/mesh2d.py
+++ b/skfem/mesh/mesh2d/mesh2d.py
@@ -55,10 +55,10 @@ class Mesh2D(Mesh):
         return np.max(np.linalg.norm(np.diff(self.p[:, self.facets], axis=1),
                                      axis=0))
 
-    def _repr_svg_(self) -> str:
+    def _repr_svg_(self) -> Optional[str]:
         from skfem.visuals.svg import draw
         if self.t.shape[1] > 5000:
-            return None
+            return
         return draw(self)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


gdmcbain:
How does this compare with [meshio.svg](https://github.com/nschloe/meshio/tree/main/meshio/svg)?

kinnala:
Some points:
- This outputs `str` whereas `meshio.svg` seems to write to file; this is needed in Jupyter object integration, see https://ipython.readthedocs.io/en/stable/config/integrating.html.
- This supports high-order meshes, curved boundaries, etc.
- I plan to extend this so that we have `skfem.visuals.svg.plot` for visualizing solutions.

gdmcbain:
Thanks.

I didn't know about the [rich display](https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display) feature of notebooks; that is an enticing feature.  I wonder whether `_repr_javascript_` would extend that to something like [femgl](https://mikolalysenko.github.io/femgl/index.html) for three-dimensional meshes.

gdmcbain:
It might be possible to factor `meshio.svg.write` (over there) to construct the string and then write that string to a file, as we did here in `skfem.io.meshio`, but I see that that's partly hindered by `mesho.svg.write` using [`xml.etree.ElementTree.write`](https://docs.python.org/3/library/xml.etree.elementtree.html?highlight=xml%20etree#xml.etree.ElementTree.ElementTree.write) which writes to &lsquo;a file name, or a file object opened for writing&rsquo;, although perhaps the latter could be an [`io.StringIO`](https://docs.python.org/3/library/io.html?highlight=stringio#io.StringIO).

However, extension to &lsquo;high-order meshes, curved boundaries, etc.&rsquo; I would think rather beyond meshio; I don't think there's enough common ground to support these features between mesh file formats.

Thanks.

PR #563, opened 2021-02-15
Make all components of DiscreteField.hod separate

kinnala:
I was trying to use `ElementGlobal.derivative > 2` with `ElementVector` and it failed. By splitting `DiscreteField.hod` into `DiscreteField.grad3`, `DiscreteField.grad4`, and so on will make it hopefully more compatible with other features of the package.


kinnala:
Maybe we should add
```python
@property
def hod(self):
    return np.array([self.grad3, self.grad4, ...], dtype=object)
```
for backwards compatibility.

PR #569, opened 2021-03-08
Discrete field operator overloads

topikuu:
Adds the remaining relevant operator overloads for nicer syntax in forms.

### DIFF
--- a/skfem/element/discrete_field.py
+++ b/skfem/element/discrete_field.py
@@ -21,13 +21,48 @@ class DiscreteField(NamedTuple):
     def __array__(self):
         return self.value
 
+    def __add__(self, other):
+        if isinstance(other, DiscreteField):
+            return self.value + other.value
+        return self.value + other
+
+    def __sub__(self, other):
+        if isinstance(other, DiscreteField):
+            return self.value - other.value
+        return self.value - other
+
     def __mul__(self, other):
         if isinstance(other, DiscreteField):
             return self.value * other.value
         return self.value * other
 
+    def __truediv__(self, other):
+        if isinstance(other, DiscreteField):
+            return self.value / other.value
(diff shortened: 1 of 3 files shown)


topikuu:
I'm not sure if addition and subtraction are really sensible operators unless the 'other' is also a DiscreteField made with the same quadrature.

Should I also add a test for these?

kinnala:
> Should I also add a test for these?

Sure, at least to have an idea what kind of use cases you had in mind for these. Probably very useful when defining different `@Functional`s? I recall sometimes having to use an explicit `u.value` and this probably circumvents that in many cases.

Some calculations currently done inside the forms are dealt by `__array__` which I think NumPy uses internally to transform an object into an array.

topikuu:
My aim is to be able to compute along the lines of
```
@BilinearForm
def bilinf_mixed(u, U, v, V, w):
    return (u - U) * (v - V)

UEl = ElementMacroBoundary(m, facet_sets=macroelems)
comp_el = ElementComposite(ElementTetP2(), UEl)
comp_fbasis = ExteriorFacetBasis(m, comp_el, intorder=3, facets=np.unique(np.concatenate(macroelems)))

C = asm(bilinf_mixed, comp_fbasis, comp_fbasis)
```

I'm not immediately aware of other applications so I'm holding this PR until I finish with the other parts required for this.

kinnala:
Yes that sounds like a good use case.

A test could simply be something like
```python
@Functional
def sum(w):
    return w.x[0] + w.x[1]
sum.assemble(basis)
```
which should evaluate to the integral of `f(x,y) = x + y`. I think `w.x[0]` have the type `DiscreteField`.

kinnala:
I'll merge this, I can add some tests in #556.

PR #573, opened 2021-03-12
flip interpolator as probes

### LINKED ISSUE #572: flip interpolator
Currently [`InteriorBasis.interpolator`](https://github.com/kinnala/scikit-fem/blob/d14d6d8398c48c86c86d353b8822acbfef830be0/skfem/assembly/basis/interior_basis.py#L131-L146) is curried, taking first the `ndarray` of a finite element function's degrees of freedom and then the `ndarray` of the coordinates of some absciss&aelig;.

If one runs a long unsteady simulation with many time-steps on the same mesh (i.e. same `InteriorBasis`), it's common to &lsquo;probe&rsquo; a fixed array of points and record the interpolated solution there; e.g. gdmcbain/fenics-tuto-in-skfem#26 which is a two-dimensional incompressible Navier&ndash;Stokes problem with 5000 time-steps in which the `ElementTriP1` pressure is logged at two points.  Using `InteriorBasis.interpolator` would require a new interpolator at each time-step.  If one could [flip](https://toolz.readthedocs.io/en/latest/api.html?highlight=flip#toolz.functoolz.flip) `interpolator` so that it was defined for its `InteriorBasis` on a fixed array of points and then took a solution `ndarray`, it would just be a matter of calling the same flipped interpolator on each new snapshot.

I think most of the uses in the suite of examples to date just call it once, so it wouldn't make much difference to them whether the method were curried either way or uncurried, but I can imagine cases in which the current currying is desirable; e.g. tracing an [integral curve](https://en.wikipedia.org/wiki/Integral_curve) on a vector solution.

The current version of the unsteady Navier&ndash;Stokes problem has a hardcoded flip of `interpolate`, a function [`probe`](https://github.com/gdmcbain/fenics-tuto-in-skfem/blob/3c1b7910286847d8964d041b10c9c622ef95383f/08_navier_stokes_cylinder/st08_navier_stokes_cylinder_pyamg.py#L49-L64), but since it seems like a reasonably common thing to want to do, I intend to think about how to refactor the existing method in scikit-fem to provide both orders of currying.

gdmcbain:
For #572.

This is quite one-sided, basically moving everything to the new flipped version; most of the work does seem to be related to the absciss&aelig; (`element_finder`, `invF`, `gbasis`) rather than the ordinates.  Maybe `interpolator` is now just for backward-compatibility?


kinnala:
Looks good. Can you remove those empty png files?

gdmcbain:
> skfem/assembly/basis/interior_basis.py:146: error: Item "None" of "Optional[Any]" has no attribute "flatten"

I don't understand this one.

kinnala:
This will fix it:
```
modified   skfem/element/discrete_field.py
@@ -8,7 +8,7 @@ from numpy import ndarray
 class DiscreteField(NamedTuple):
     """A function defined at the global quadrature points."""
 
-    value: Optional[ndarray] = None
+    value: ndarray
     grad: Optional[ndarray] = None
     div: Optional[ndarray] = None
     curl: Optional[ndarray] = None
```
I think we always want to have at least `value` when creating `DiscreteField`?

Another option would be to check the contents of `phis` before using it but I think the above change is completely fine. At least all tests run if you apply the change.

gdmcbain:
> I think we always want to have at least value when creating DiscreteField?

I think so.   I can't think of a counterexample; were one to arise I guess the value could have zero components? Anyway I won't worry about that bit.

The bit that I didn't understand though was that I don't think that this was triggered by anything I added or deleted; the `.flatten()` was already there.  Anyway, I won't worry about that either.

I would quite like to get rid of the `for`-loop.  I'll take the opportunity to look at that.  (Unless there's a reason for it.)

kinnala:
> The bit that I didn't understand though was that I don't think that this was triggered by anything I added or deleted; the `.flatten()` was already there. Anyway, I won't worry about that either.

I have noticed that mypy seems to spot these mistakes very randomly. The change, for an unknown reason, allowed mypy to detect the typing mistake.

gdmcbain:
Ah, O. K., thanks.  I should study mypy; I know it's newish but it is very good to have any kind of support for types in numerical Python.

gdmcbain:
> I would quite like to get rid of the `for`-loop. 

https://github.com/kinnala/scikit-fem/blob/8763b9e415035acb8de406bd7e35ba89d7290f29/skfem/assembly/basis/interior_basis.py#L144-L146

I think this also unnecessarily restricts `InteriorBasis.interpolator` to scalars;

https://github.com/kinnala/scikit-fem/blob/8763b9e415035acb8de406bd7e35ba89d7290f29/skfem/assembly/basis/interior_basis.py#L143

In an example like gdmcbain/fenics-tuto-in-skfem#26, referred to in #572, one might well want to interpolate the velocity rather than the pressure; a more general NumPy `sum` might facilitate that generalization.

Issue #579, opened 2021-03-15
new Mesh class lacks meshio_type attribute


gdmcbain:
A couple of my scripts broke after #556, raising, e.g.,

> AttributeError: 'MeshLine1' object has no attribute 'meshio_type'

I see that this is still handled in the usual cases by [`to_meshio`](https://github.com/kinnala/scikit-fem/blob/e0703948a5d5ed18f78c935a9d6929f76a104648/skfem/io/meshio.py#L143) or [`to_file`](https://github.com/kinnala/scikit-fem/blob/e0703948a5d5ed18f78c935a9d6929f76a104648/skfem/io/meshio.py#L153) which use

https://github.com/kinnala/scikit-fem/blob/e0703948a5d5ed18f78c935a9d6929f76a104648/skfem/io/meshio.py#L11-L22

The failing scripts weren't using either of these functions from `skfem.io.meshio` because they needed to geometrically transform the `skfem.Mesh`; e.g. the `MeshLine` was laid along the _z_-axis with

```python
 meshio.Mesh(
                     np.array([[0., 0., z] for z in mesh.p[0]]),
                     ...
```
and a `MeshQuad` in the _yz_-plane with
```python
          meshio.Mesh(
                 (self.transform @ self.mesh.p).T,
                 ...
```
where `self.transform` is loaded from
```json
   "transform": [[0, 0],
		  [-1, 0],
		  [0, -1]],
```

As I'm writing custom output functions for these cases anyway I can just hard-code `"line"`, `"quad"`, &c., or
```python
from skfem.io.meshio import TYPE_MESH_MAPPING
```
and then use `TYPE_MESH_MAPPING[type(m)]` for `m: skfem.Mesh`, but I wanted to check whether this excision was deliberate or an oversight.

kinnala:
Basically anything that's not tested may be missing and apparently `meshio_type` was not included in the test suite. Do you think we should include it? If yes, we should implement a property which fetches it from `TYPE_MESH_MAPPING`. On the otherhand, next release will be 3.0.0 anyways so we can do small backwards-incompatible modifications if required.

gdmcbain:
There are a failed and a successful attempt at fixing this in #580 and #584 respectively, but as discussed under the latter I'm inclined to leave this to user-level makeshifts.

PR #580, opened 2021-03-17
reinstate Mesh.meshio_type

gdmcbain:
As discussed in #579.

### DIFF
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -399,6 +399,12 @@ class Mesh:
         return to_file(self, filename, point_data, **kwargs)
 
     @classmethod
+    @property
+    def meshio_type(cls) -> str:
+        from skfem.io.meshio import TYPE_MESH_MAPPING
+        return TYPE_MESH_MAPPING[cls]
+
+    @classmethod
     def load(cls, filename):
         from skfem.io.meshio import from_file
         return from_file(filename)


gdmcbain:
```
==================================== ERRORS ====================================
_____________________ ERROR collecting skfem/mesh/mesh.py ______________________
skfem/mesh/mesh.py:405: in meshio_type
    return TYPE_MESH_MAPPING[cls]
E   KeyError: <class 'skfem.mesh.mesh.Mesh'>
=========================== short test summary info ============================
ERROR skfem/mesh/mesh.py - KeyError: <class 'skfem.mesh.mesh.Mesh'>
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 0.85s ===============================
```

I'm not sure what this is and can't reproduce it here with Python 3.9.0 under Ubuntu 20.04.2 LTS; `pytest -n auto tests` is clear.

kinnala:
Strange stuff. I read that something was changed in Python 3.9 in relation to classmethods: https://bugs.python.org/issue42832

kinnala:
Btw, this should run exactly the same setup as the test suite: https://github.com/kinnala/scikit-fem/blob/master/Makefile#L42

gdmcbain:
Thanks, `make test_py39` looks good; unfortunately it's hard to run here because of an `SSLCertVerificationError` but that's an all too well known local problem that I labour under.

I think I'll assume that chaining decorators as in

https://github.com/kinnala/scikit-fem/blob/bac63ba847f6db6e913ff85bebccb4371c4e988c/skfem/mesh/mesh.py#L401-L403

isn't a good idea and will try to think of something else.

PR #584, opened 2021-03-18
WIP: meshio_type attribute

gdmcbain:
This is another attempt at #579, avoiding the problematic (not to mention questionable) chained decorators of #580.

### DIFF
--- a/skfem/io/meshio.py
+++ b/skfem/io/meshio.py
@@ -8,9 +8,9 @@ import numpy as np
 import skfem
 
 MESH_TYPE_MAPPING = {
-    e.meshio_type: getattr(skfem.mesh, e)
-    for e in dir(skfem.mesh)
-    if hasattr(e, "meshio_type")
+    t.meshio_type: t
+    for t in (getattr(skfem.mesh, e) for e in dir(skfem.mesh))
+    if hasattr(t, "meshio_type")
 }
 
 TYPE_MESH_MAPPING = {v: k for k, v in MESH_TYPE_MAPPING.items()}
--- a/skfem/io/meshio.py
+++ b/skfem/io/meshio.py
@@ -140,7 +140,7 @@ def to_meshio(mesh, point_data=None):
     if isinstance(mesh, skfem.MeshHex):
         t = t[[0, 3, 6, 2, 1, 5, 7, 4]]
 
-    cells = {TYPE_MESH_MAPPING[type(mesh)]: t.T}
+    cells = [(mesh.meshio_type, t.T)]
     return meshio.Mesh(mesh.p.T, cells, point_data)
 
 
--- a/skfem/io/meshio.py
+++ b/skfem/io/meshio.py
@@ -7,16 +7,10 @@ import numpy as np
 
 import skfem
 
-
 MESH_TYPE_MAPPING = {
-    'tetra': skfem.MeshTet,
-    'hexahedron': skfem.MeshHex,
-    'triangle': skfem.MeshTri,
-    'quad': skfem.MeshQuad,
-    'line': skfem.MeshLine,
-    'tetra10': skfem.MeshTet2,
-    'triangle6': skfem.MeshTri2,
-    'quad9': skfem.MeshQuad2,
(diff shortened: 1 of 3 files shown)


gdmcbain:
This is experimental because it violates the [directive](https://github.com/kinnala/scikit-fem/issues/579#issuecomment-799996754)

>  If yes, we should implement a property which fetches it from `TYPE_MESH_MAPPING`. 

Instead it *constructs* `skfem.io.meshio.TYPE_MESH_MAPPING` from whatever's in `module skfem.mesh` that has a `meshio_type`.

kinnala:
I suppose it's fine, but in a dataclass I somehow consider attributes something that can/should be changed instance by instance, and I think changing `meshio_type` doesn't lead to anything? I need to think about it further.

kinnala:
I mean, the automatically generated constructor has `meshio_type` and so does the automatically generated `__repr__` or `__hash__` if we start to use those at some point.

gdmcbain:
> in a dataclass I somehow consider attributes something that can/should be changed instance by instance

Ah.

So at the moment we have things like 

https://github.com/kinnala/scikit-fem/blob/4b2a8682b8a2b8cebd85149d5627deb4fd6a877f/skfem/mesh/mesh.py#L115

which is an instance method that is always going to return the same thing for every instance of each subclass of Mesh.  (In the same way as something&mdash;attribute, property, method&mdash;that would tell us what the `meshio_type` is.)

gdmcbain:
> I mean, the automatically generated constructor has meshio_type and so does the automatically generated __repr__ or __hash__ if we start to use those at some point.

Right, I see, yes.

So a property might be more appropriate.  Let's see&hellip;

gdmcbain:
I'd rather have had a property which is constant for a class depend on the class rather than the instance but that was #580 which didn't seem to work because there's no `@classproperty`, but that mightn't be too important.'

A (instance) property does mean that I can't ask it of the class which is annoying so I have to instantiate the class.

gdmcbain:
Ah, the order of MESH_TYPE_MAPPING was significant, wasn't it, so that a `MeshTri` with boundary elements wouldn't get mistaken for a `MeshLine`.

gdmcbain:
Well that worked, but it's a bit verbose and I just found that dataclasses do have [class variables](https://docs.python.org/3/library/dataclasses.html?highlight=dataclasses#class-variables) after all, so I'll try that.

gdmcbain:
Should something like

https://github.com/kinnala/scikit-fem/blob/4b2a8682b8a2b8cebd85149d5627deb4fd6a877f/skfem/mesh/mesh.py#L755

also be a `ClassVar`?  Every `MeshTri.elem` has to be the same; it shouldn't be changed instance by instance or be included in `__repr__` &c.

kinnala:
> also be a `ClassVar`? Every `MeshTri.elem` has to be the same; it shouldn't be changed instance by instance or be included in `__repr__` &c.

Fair point. Technically it's possible to change `elem` to something else in `MeshTri1`, have meaningful results and some of the methods would work just fine, but, e.g., `MeshTri1.refine` pretty much assumes `elem=ElementTriP1`.

I think this is more like a technical thing regarding the superclass `Mesh` which can now, in theory, be initialized using any `elem`. One could, for example, invent `ElementHexagonal` and use that to define a hexagonal mesh.

Can we set `elem` to `ClassVar` in the subclass (i.e. `MeshTri1`) only and still have it free in `Mesh`? Does it pass `mypy` check?

gdmcbain:
Right, sorry, `Mesh.elem` was a digression.  I don't think I realized that a `Mesh` had an `elem`; I knew that a `Basis` has an `elem` of course (alongside its `mesh`) but was probably confusing the `elem` *of* a `Mesh` with the meshio cell-type (which is a class constant).  (I'll look into what `Mesh.elem` is for later.)

I think, on reflexion, I'm pleased that I got this to work and pleased to have discovered `typing.ClassVar` but I don't know that #579 is serious enough to merit the changes developed here.  I propose to close this and close #579 and just cope with the missing `Mesh.meshio_type` in user-level scripts.

PR #585, opened 2021-03-18
Cache element_dofs property in basis

topikuu:
This improves performance in almost all cases.

Signed-off-by: Topi Kuutela <topi.kuutela@aalto.fi>

### DIFF
--- a/skfem/assembly/basis/basis.py
+++ b/skfem/assembly/basis/basis.py
@@ -1,5 +1,6 @@
 import warnings
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
+from functools import lru_cache
 
 import numpy as np
 from numpy import ndarray
@@ -9,6 +10,7 @@ from skfem.mapping import Mapping
 from skfem.mesh import Mesh
 from skfem.quadrature import get_quadrature
 from skfem.refdom import Refdom
+from skfem.generic_utils import HashableNdArray
 
 
 class Basis:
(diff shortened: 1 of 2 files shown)


topikuu:
element_dofs can't be turned into a functools.cached_property as the basis.tind is modified at some points, breaking its evaluation.

Where should the HashableNdArray be put? The roof folder is probably not the right location.

kinnala:
Where is `self.tind` modified?

kinnala:
Alright, it has something to do with the initialization order.

kinnala:
I think it could be possible to refactor the constructors so that `self.tind` is basically always a constant whenever `self.element_dofs` gets called.

kinnala:
New:
```python
In [2]: m = MeshTet().refined(5)
In [6]: basis = InteriorBasis(m, ElementTetP2())                                
In [7]: %timeit laplace.assemble(basis)                                         
2.83 s ± 25.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Old:
```python
In [7]: %timeit laplace.assemble(basis)                                         
3.35 s ± 36.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

skfem/assembly/basis/basis.py
```diff
@@ -97,8 +99,15 @@ def N(self):
     @property
     def element_dofs(self):
         if self.tind is None:
+            return self.get_element_dofs(None)
+        else:
```
kinnala, line 103: `else` is unnecessary?

skfem/assembly/basis/basis.py
```diff
@@ -97,8 +99,15 @@ def N(self):
...
+            return self.get_element_dofs(HashableNdArray(self.tind))
+
+    @lru_cache(maxsize=128)
```
kinnala, line 106: What is the rationale for `maxsize=128`?

kinnala:
> Where should the HashableNdArray be put? The roof folder is probably not the right location.

This location is fine for now.

kinnala:
I did similar testing of initializing `InteriorBasis` as was done in #586 but the timings here seem to be equal, and assembly is faster, so I think this is good to go.

PR #586, opened 2021-03-18
Cache kernels

topikuu:
I'm not sure if this is a good idea in the mainline. Feel free to deny this PR.

Caching _kernel evaluations is extremely beneficial in my use cases as I often use a composite element basis. Basically this allows skipping the evaluation of _kernel(..) when the basis consists of zeros. The basis can be zeros when CompositeElements are 

I found that in order to get reasonable performance on MacroElements (see: [WIP commit](https://github.com/topikuu/scikit-fem/commit/dd5619679f2a2e2b5bd91ffafc36c908a5614ef8)) really need either this or a rather extensive modification to dofs.py and bilin_assembly.
The modification would consist of:
- Changing dofs.element_dofs into a list of arrays of different lengths (instead of current ndarray) as all elements don't have all the basis functions.
- Finding the intersection of basis supports in bilin_assembly before _kernel evaluations. Bookkeeping the indexing gets non-trivial but is doable.
- A ton of small changes all over the place as a lot of things seem to rely on all elements having all the basis functions.

Is it justifiable to add this complexity to the mainstream to support, quite frankly, obscure and rarely used elements?

### DIFF
--- a/skfem/assembly/basis/basis.py
+++ b/skfem/assembly/basis/basis.py
@@ -1,5 +1,6 @@
 import warnings
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
+from functools import lru_cache
 
 import numpy as np
 from numpy import ndarray
@@ -9,6 +10,7 @@ from skfem.mapping import Mapping
 from skfem.mesh import Mesh
 from skfem.quadrature import get_quadrature
 from skfem.refdom import Refdom
+from skfem.generic_utils import HashableNdArray
 
 
 class Basis:
@@ -97,8 +99,14 @@ class Basis:
     @property
     def element_dofs(self):
         if self.tind is None:
+            return self.get_element_dofs(None)
+        return self.get_element_dofs(HashableNdArray(self.tind))
+
+    @lru_cache(maxsize=128)
+    def get_element_dofs(self, tind):
+        if tind is None:
             return self.dofs.element_dofs
-        return self.dofs.element_dofs[:, self.tind]
+        return self.dofs.element_dofs[:, tind]
 
     def complement_dofs(self, *D):
         if type(D[0]) is dict:
--- a/skfem/assembly/form/bilinear_form.py
+++ b/skfem/assembly/form/bilinear_form.py
@@ -1,9 +1,11 @@
 from typing import Optional, Any
+from functools import lru_cache
 
 import numpy as np
 
 from .form import Form, FormDict
 from ..basis import Basis
+from skfem.generic_utils import HashableNdArray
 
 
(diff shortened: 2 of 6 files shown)


kinnala:
I will try this out; reporting later if I notice anything performance wise. I think this is not complicating things too much.

kinnala:
Perhaps surprisingly, there seems to be a significant difference between `master` and this branch in the time taken to intialize `InteriorBasis` for larger tetrahedral meshes.

This branch:
```python
In [1]: %timeit basis = InteriorBasis(m, ElementTetP2())                        
1.4 s ± 223 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

master:
```python
In [1]: %timeit basis = InteriorBasis(m, ElementTetP2())                        
554 ms ± 32.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

The initialization code is as follows:
```python
from skfem import *
m = MeshTet().refined(5)
```

The assembly itself times equal.

kinnala:
I realized doing `%timeit` here is not very wise because the resulting objects use quite a bit of memory, but still running individual timings the difference seems to be twofold:
```python
In [1]: %time InteriorBasis(m, ElementTetP2())                                  
CPU times: user 1.75 s, sys: 214 ms, total: 1.97 s
Wall time: 1.96 s
Out[1]: <skfem.assembly.basis.interior_basis.InteriorBasis at 0x7f5522a78040>

In [2]: %time InteriorBasis(m, ElementTetP2())                                  
CPU times: user 792 ms, sys: 189 ms, total: 981 ms
Wall time: 980 ms
Out[2]: <skfem.assembly.basis.interior_basis.InteriorBasis at 0x7f5522b810d0>

In [3]: %time InteriorBasis(m, ElementTetP2())                                  
CPU times: user 807 ms, sys: 192 ms, total: 999 ms
Wall time: 998 ms
Out[3]: <skfem.assembly.basis.interior_basis.InteriorBasis at 0x7f5555b67190>

In [4]: %time InteriorBasis(m, ElementTetP2())                                  
CPU times: user 823 ms, sys: 182 ms, total: 1 s
Wall time: 1 s
Out[4]: <skfem.assembly.basis.interior_basis.InteriorBasis at 0x7f5522b81610>
```
whereas it's something like 500 ms using `master`.

topikuu:
Right, I had only measured this with element_dof caching vs. master without element_dof caching. Considering the performance drop is quite significant in the common use case, I'll close this PR.

The proper approach by computing the (sub)domain intersections would most probably be performant enough (with element_dof caching), but in my opinion not worth the effort and complexity at this point.

PR #596, opened 2021-03-26
COOData, utils.enforce, unsymmetric eigenproblems, visuals.matplotlib bugfix

### LINKED ISSUE #591: apply Dirichlet conditions by setting rows to I, rhs to value
For the first time, I'm looking at scikit-fem in a serious manner. Diving into it, I notice that the common way the "apply" Dirichlet conditions is to eliminate the respective entries from the matrix/rhs à la
```python
A, b = condense(A, b, I=mesh.interior_nodes())
```
I on the other hand would like to see those condition applied via setting the respective rows in `A` to diagonal `1` and `0` otherwise plus the respective rhs entry to the value. Something like
```python
apply_dirichlet(A, b, mesh.boundary_nodes(), vals)
```
perhaps. The resulting matrix has the same spectral properties as the condensed `A`, so the number of iterations with Krylov methods will be the same, too. Advantages of this approach include the fact that neither `A` nor `b` need to be rewritten in memory. Also, the solution will include the values of _all_ degrees of freedom, making postprocessing easier.

### LINKED ISSUE #608: saving solution as SVG results in huge file
When saving a solution of a 1D problem with 1000 dofs as an SVG, the file will be a whopping 116MB large. I suppose there's something wrong with `plot`?

MWE:
```python
from skfem import *
from skfem.models.poisson import laplace, unit_load

m = MeshLine().refined(10)

e = ElementLineP1()
basis = InteriorBasis(m, e)

A = asm(laplace, basis)
b = asm(unit_load, basis)

x = solve(*condense(A, b, I=m.interior_nodes()))

if __name__ == "__main__":
    from os.path import splitext
    from sys import argv
    from skfem.visuals.matplotlib import plot, savefig

    plot(m, x, shading='gouraud', colorbar=True)
    savefig("test_solution.svg")
```

kinnala:
- Fixes #608 
- Fixes #591


skfem/assembly/form/coo_data.py
```diff
@@ -0,0 +1,54 @@
...
+    def enforce(self, D):
+        """Enforce an essential BC by setting rows and diagonals to 0 and 1.
+
+        Parameters
+        ----------
+        D
+            A list of (Dirichlet) degrees-of-freedom to enforce.
+
+        """
+        rows_mapping = np.ones(self.shape[0])
```
gdmcbain, line 45: Should this method have an option for handling mass matrices?  Or &lsquo;pencils&rsquo; consisting of a stiffness matrix and a mass matrix? As in Mx' + Kx = 0, as in 

https://github.com/kinnala/scikit-fem/blob/a88739744505ace72f59021400a548f3e16e0ded/docs/examples/ex29.py#L122

Or more general cases like the second-order Mx'' + Cx' + Kx = f, e.g. for linear elastodynamics.  I think that M or M & C get zeros on the diagonal rather than ones, for constant Dirichlet data.

I've been meaning to put together a simple example to investigate #591 in generalized eigenproblems, e.g. a vibrating string fixed at each end.  (Not ex16 as that doesn't need boundary conditions because the termini are singular points of Legendre's equation.)

kinnala: So basically skipping the part where diagonal is set to zero?

kinnala: Check out the latest commit.

skfem/assembly/form/coo_data.py
```diff
@@ -0,0 +1,38 @@
...
+    def tocsr(self) -> csr_matrix:
+        """Return a sparse SciPy CSR matrix."""
+        return self._assemble_scipy_csr(
+            self.data,
+            self.rows,
+            self.cols,
+            self.shape,
+        )
+
+    def todense(self) -> ndarray:
```
gdmcbain, line 35: [`coo_matrix.todense`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.coo_matrix.todense.html#scipy.sparse.coo_matrix.todense) actually returns a [`numpy.matrix`](https://numpy.org/devdocs/reference/generated/numpy.matrix.html#numpy.matrix) and 

> It is no longer recommended to use this class, even for linear algebra. Instead use regular arrays. The class may be removed in the future.

It is a subclass of `ndarray` though so probably passes type-tests on that basis.

Maybe go with [`coo_matrix.toarray`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.spmatrix.toarray.html#scipy.sparse.spmatrix.toarray)?

skfem/utils.py
```diff
@@ -219,11 +221,87 @@ def _flatten_helper(S, key):
...
+    elif D is None and I is not None:
+        D = np.setdiff1d(np.arange(A.shape[0]), I)
+    else:
+        raise Exception("Give only I or only D!")
+
+    if x is None:
+        x = np.zeros(A.shape[0])
+    elif b is None:
+        b = np.zeros_like(x)
+
```
gdmcbain, line 275: Share the above with `condense`?

skfem/utils.py
```diff
@@ -219,11 +221,87 @@ def _flatten_helper(S, key):
...
+        The original system is both modified (for performance) and returned
+        (for compatibility with :func:`skfem.utils.solve`).
+
+    Parameters
+    ----------
+    A
+        The system matrix
+    b
+        Optionally, the right hand side vector.
```
gdmcbain, line 242: For `condense`, the second positional argument, `b` is taken to be the mass matrix of a generalized algebraic eigenvalue problem if it's an `spmatrix` rather than an `ndarray`.  Is it worth doing the same here?  It will be the same entries that are to be affected in both matrices.

gdmcbain: I think of the generalized algebraic eigenvalue problem as arising from the dynamical system Mu' = Ku by the separation of variables u(t) = exp(st) v as sMv=Kv.  The steady problem is Kv=0 and that's the one that takes the possibly non-zero Dirichlet values so it's K that gets ones on the diagonal. In rhe unsteady problem, the Dirichlet degrees of freedom don't change so the corresponding diagonal entries of M should be zero. In the generalized eigenvalue problem, each Dirichlet degree of freedom has a s=1/0 = &infin; generalized eigenvalue corresponding to a zero time-constant in exp st = exp (t/0). If the Dirichlet data were changing in time, the degrees of freedom would follow instantaneously, with no relaxation time. Obviating these additional eigenvalues is one motivation for condensation over enforcement.

gdmcbain: > I think for eigenvalue problems you cannot set inhomogeneous boundary values

No, that's right. The inhomogeneous data would be applied to the steady problem then the unsteady problem rewritten for the perturbation. I skiipper that step in the previous. So Mx' = Kx - b with steady problem Kx=b and perturbation u satisfying the homogeneous ordinary differential equation Mu' = Ku.

kinnala:
I'm on a leave since yesterday so it might take some days before this gets merged but I'll do the fixes as soon as I'm able to.

skfem/utils.py
kinnala: Can't get this to work, maybe I'm missing something.

kinnala:
```python
    if b is not None:
        if isinstance(b, spmatrix):
            # eigenvalue problem
            b = enforce(b, D=D, diag=0.)
        else:
            # set rhs to the given value
            b[D] = x[D]
        return A, b
```

kinnala: https://scicomp.stackexchange.com/questions/31140/dirichlet-boundary-conditions-in-generalized-eigenvalue-problem

kinnala: Seems I was accidentally using `eigsh` which is for symmetric cases,

gdmcbain: Sorry, I didn't get to put together the MWE before Easter. The string on a MeshLine would be fine or actually even simpler the thermal problem, the one-dimensional eigenvalue analogue of [ex19](https://github.com/kinnala/scikit-fem/blob/3369c040005c90e7edd925694620c8bfd1048286/docs/examples/ex19.py). 

In the scicomp SE question, I think the OP's idea that 1 (=condense) makes most sense for generalized algebraic eigenproblems is right while 2 (=enforce) for &lsquo;solving linear systems of equations&rsquo; is reasonable and matches #591 , although I'm wondering whether its real advantages will shine for unsteady and nonlinear problems. I don't like 3 (a kind of inverse enforce) as it's unphysical. The additional zeros eigenvalues obtained with `enforce` do have a physical interpretation; comparison with (the as yet unwritten) `penalize` should reveal this. SB's suggestion of further modifying the mass matrix to artificially preserve its properties of symmetry and positive definiteness is definitely something that i would rather avoid. It's true that there is a lot of that in the spectral methods and hydrodynamic stability literature but compare how simple [ex29](https://github.com/kinnala/scikit-fem/blob/3369c040005c90e7edd925694620c8bfd1048286/docs/examples/ex29.py) is, using `condense`.

The issue of hanging on to these properties might not be relevant if, as usual in Krylov methods for eigenvalue problems, one is employing a shift (`sigma`) since that modifies the pencil, but that would have to be checked.

Meanwhile, so as not to hold up `enforce` for the static case while the generalized eigenproblem is pondered, particularly since `condense` is possibly preferred there anyway, it might be reasonable to leave the `isinstance(b, spmatrix)` case as `NotImplemented` to begin with.

kinnala: I think I was able to get it to work. I changed `TestEx21` to run using `enforce` and the result was almost the same, maybe 0.1% difference due to unsymmetric matrix and using a different function of ARPACK.

Issue #598, opened 2021-03-30
default zero RHS


gdmcbain:
As the discussion of how to enforce essential conditions #591 turned to usability and syntax there arose:

> (I wonder whether it would be reasonable to let the RHS `b` default to zero in `condense` and its forthcoming siblings? `zeros_like` does appear fairly often in the second slot for condense.)

I'll try this out.

PR #599, opened 2021-03-30
b=zeros like x

gdmcbain:
For #598.


kinnala:
Looks fine.

I wonder if at some point we could maybe get rid of the possibility of calling `condense(A)` with no extra parameters? Is that useful anywhere?

gdmcbain:
But that is also a case that could well benefit from `skfem.utils.enforce` #591, in which case maybe yes, one could exclude `condense(A, I=I)`.

Issue #593, opened 2021-03-25
construct evaluation matrix


nschloe:
Given a domain and a number of `points` within this domain, I would like to construct the small and fat evaluation matrix `E` which, given coefficient vector `x` returns `E @ x`, the values of the corresponding finite element function at the `points`. In FEniCS, I could do
```python
  def _build_eval_matrix(V, points):
      """Build the sparse m-by-n matrix that maps a coefficient set for a
      function in V to the values of that function at m given points.
      """
      mesh = V.mesh()

      bbt = BoundingBoxTree()
      bbt.build(mesh)
      dofmap = V.dofmap()
      el = V.element()
      sdim = el.space_dimension()

      rows = []
      cols = []
      data = []
      for i, x in enumerate(points):
          cell_id = bbt.compute_first_entity_collision(Point(*x))
          cell = Cell(mesh, cell_id)
          coordinate_dofs = cell.get_vertex_coordinates()

          rows.append(numpy.full(sdim, i))
          cols.append(dofmap.cell_dofs(cell_id))

          v = el.evaluate_basis_all(x, coordinate_dofs, cell_id)
          data.append(v)

      rows = numpy.concatenate(rows)
      cols = numpy.concatenate(cols)
      data = numpy.concatenate(data)

      m = len(points)
      n = V.dim()
      matrix = sparse.csr_matrix((data, (rows, cols)), shape=(m, n))
      return matrix
```
The explicit loop doesn't hurt much since the number of `points` is usually small.

Any hints?

gdmcbain:
This is very closely related to the recent feature &lsquo;flip interpolator as probes&rsquo; #573.  That defined a `probes` method of an `InteriorBasis` that took an `ndarray` of points and returned a function from V to &reals;<sup><i>m</i></sup>.  Internally, the returned function does

https://github.com/kinnala/scikit-fem/blob/b29f231e81f9ab6027dd696cd3e682a64cd3ed46/skfem/assembly/basis/interior_basis.py#L147

where `phis` and `indices` are ndarrays known to the function (it's a closure).

Is this good enough?  It'd be easy enough to rejig `InteriorBasis.probes` to return `E` if required, I think.

kinnala:
> It'd be easy enough to rejig `InteriorBasis.probes` to return `E` if required, I think.

Isn't `E` essentially `phis` inside `probes` with different shape and some additional zeros?

Should we have a separate method for returning `E` or is this sufficient?

[...]

Issue #600, opened 2021-03-30
method of characteristics


gdmcbain:
> The method of characteristics is tailored to solve *unsteady advection equations*

(Ern & Guermond 2004, _Theory and Practice of Finite Elements_, &sect;6.3.3 &lsquo;The method of characteristics&rsquo;).

(This is a stub which I'll fill out shortly, but for now to refer to from the discussion of the evaluation matrix #593.  I expect it to be a heavy load on that or `probes` #572.)

kinnala:
Should it be solved like this: 1) find the points that map to the quadrature points through the advection map, 2) find out which of them are outside of the domain, 3) evluate rest of them through `probes`.

Then we could have a separate function which checks if a point is inside the domain or not.

gdmcbain:
Following the discussion in #646, in particular your [question](https://github.com/kinnala/scikit-fem/issues/646#issuecomment-886016345)

> Is the time step so large that the point can move over several elements?

which is called in the finite difference context the Courant number *u* &Delta; *t* / &Delta; *x* or CFL condition after Courant, Friedrich, & Lewy (1928, 1967), I'm thinking that (3) should use a modified version of probes that is already told which cell each point is in and so doesn't have to call `element_finder` as in

https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/skfem/assembly/basis/cell_basis.py#L146

To begin with, one might check using the local-to-canonical map whether the point is in the guessed cell. For those that aren't, try the adjacent cell. I haven't looked at the details of what the local-to-canonical map looks like for a point not in the cell to see whether that might guide which facet has been crossed and so which of the adjacent cells is to be taken as the next guess, but if there's something there, it be better to use it than to fall back on an exhaustive search of the neighbours.

The mesh's facet connectivity array already indicates which facets are on the boundary of the domain, via the sentinel value &minus;1.

https://github.com/kinnala/scikit-fem/blob/0855dae117728588f269d968c8da4dbe06446d1b/skfem/mesh/mesh.py#L670

   —

 * Courant, R., Friedrichs, K. & Lewy, H. (1967). On the Partial Difference Equations of Mathematical Physics. *IBM Journal*, 11, 215--234.

gdmcbain:
On the other hand, in examples like 
* exx 25, 28 (simple channels)
*  ex 27 (backwardfacing step)

[...]

PR #601, opened 2021-03-30
probes matrix

gdmcbain:
To flesh out the discussion of #593, to see what the matrix `E` would look like.

```python
from docs.examples.ex01 import basis, x
import numpy as np

probes_points = np.array([0, .5])[:, None] + np.outer([1, 0], np.linspace(0, 1, 9))
E = basis.probes_matrix(probes_points)
E @ x
np.linalg.norm(E @ x -basis.probes(probes_points)(x))
```

### DIFF
--- a/skfem/assembly/basis/interior_basis.py
+++ b/skfem/assembly/basis/interior_basis.py
@@ -131,7 +131,7 @@ class InteriorBasis(Basis):
 
         return M, w.flatten()
 
-    def _probes_data(self, x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
+    def _probes_data(self, x: ndarray) -> Tuple[ndarray, ndarray]:
         """Return a values of basis functions and relevant indices
         for interpolating a solution at the given points `x`."""
 
@@ -145,7 +145,7 @@ class InteriorBasis(Basis):
         indices = self.element_dofs[:, cells]
         return phis, indices
 
-    def probes_matrix(self, x: np.ndarray) -> coo_matrix:
+    def probes_matrix(self, x: ndarray) -> coo_matrix:
         """Return a coo_matrix which acts on a solution to probe its values
         at the given points `x`"""
         phis, indices = self._probes_data(x)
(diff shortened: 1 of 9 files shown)


skfem/assembly/basis/interior_basis.py
```diff
@@ -130,9 +131,9 @@ def refinterp(self,
 
         return M, w.flatten()
 
-    def probes(self, x: np.ndarray) -> Callable[[np.ndarray], np.ndarray]:
-        """Return a function handle, which can be used for finding
-        the values on points `x` of a given solution vector."""
+    def probes_data(self, x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
```
kinnala, line 134: Rename to `_probes_data`?

kinnala: Another comment: I've been importing `ndarray` explicitly from numpy. No any real reason but to make the type definitions shorter (they are sometimes pretty long). You can also here write them without `np` because `ndarray` is already imported.

gdmcbain: > Rename to `_probes_data`?

Yes, of course; done.

I was also wondering whether we wanted to rethink `InteriorBasis.probes`; whether we need or want both the function and matrix versions.  `.probes` is pretty new, indeed unreleased, so I think it could be switched to return the matrix instead of the function and the vestigial `InteriorBasis.interpolator` still reimplemented, as is, in terms of the matrix.  What do you think?  I might try that.  Would there be any efficiency difference between the matrix and the function?  I'm thinking it unlikely (but might well find out in #600).

gdmcbain: In 4e3afb4, I got rid of `_probes_data` anyway; should anyone want the old `phis` and `indices` (though I can't think why) they could be recovered from the `data` and `col` attributes of the `coo_matrix` returned by the revised `probes` method.

gdmcbain:
> `E = basis.probes_matrix(probes_points)`

With 4e3a4fb, this would be just

```python
E = basis.probes(probe_points)
```
and the old `.probes` method returning a function would no longer available for comparison.

Issue #603, opened 2021-04-01
ExteriorFacetBasis.with_element


gdmcbain:
Whoops, attempted something like
```python
from docs.examples.ex24 import element, inlet_basis
 
inlet_basis.with_element(element["u"].elem)
```
and got [`NotImplementedError`](https://github.com/kinnala/scikit-fem/blob/93844bb32d6823eb7386a58a77a40e4b3b7e2472/skfem/assembly/basis/basis.py#L285).

I think it should be like the implementation in `InteriorBasis` but with `facets=self.find` instead of 
https://github.com/kinnala/scikit-fem/blob/93844bb32d6823eb7386a58a77a40e4b3b7e2472/skfem/assembly/basis/interior_basis.py#L167

PR #604, opened 2021-04-01
ExteriorFacetBasis.with_element

gdmcbain:
For #603.


gdmcbain:
Much of the code could be shared with the `InteriorBasis` implementation but it might look a little tangled.

PR #605, opened 2021-04-01
basis_to is not Optional in utils.projection

gdmcbain:
It's needed here

https://github.com/kinnala/scikit-fem/blob/93844bb32d6823eb7386a58a77a40e4b3b7e2472/skfem/utils.py#L445

and `None` won't work.

### DIFF
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -397,7 +397,7 @@ def adaptive_theta(est, theta=0.5, max=None):
 
 
 def projection(fun,
-               basis_to: Optional[Basis] = None,
+               basis_to: Basis,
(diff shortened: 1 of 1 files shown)


gdmcbain:
Ah, this fails because the deprecated `skfem.utils.project` passes an `Optional[Basis]` `basis_to` to its successor, `projection`; that can't be made in turn not `Optional` because it follows `basis_from` in the argument list of `project` and that is `Optional`&hellip;  I guess this has to wait till `project` is removed.

kinnala:
The original order of arguments in `project` was obviously a mistake because it's so common to specify only the first argument and `basis_to`.

Issue #609, opened 2021-04-07
enforcing Dirichlet conditions in ex21 raises IndexError or TypeError


gdmcbain:
While following up the discussion of `enforce` #591, in particular in relation to generalized eigenvalue problems, I noticed that it had been used in 

https://github.com/kinnala/scikit-fem/blob/56fe1564d39c6ae0ea94c98c4ace2f864e4db106/docs/examples/ex21.py#L88

but then that running the example raises quite a lot of noise.
```
/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/numpy/core/_asarray.py:83: ComplexWarning: Casting complex values to real discards the imaginary part
  return array(a, dtype, copy=False, order=order)
Exception in Tkinter callback
Traceback (most recent call last):
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/tkinter/__init__.py", line 1885, in __call__
    return self.func(*args)
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/tkinter/__init__.py", line 806, in callit
    func(*args)
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/matplotlib/backends/_backend_tk.py", line 253, in idle_draw
    self.draw()
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/matplotlib/backends/backend_tkagg.py", line 9, in draw
    super(FigureCanvasTkAgg, self).draw()
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/matplotlib/backends/backend_agg.py", line 407, in draw
    self.figure.draw(self.renderer)
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/matplotlib/artist.py", line 41, in draw_wrapper
    return draw(artist, renderer, *args, **kwargs)
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/matplotlib/figure.py", line 1863, in draw
    mimage._draw_list_compositing_images(
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/matplotlib/image.py", line 131, in _draw_list_compositing_images
    a.draw(renderer)
  File "/home/gmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/matplotlib/artist.py", line 41, in draw_wrapper
[...]

PR #612, opened 2021-04-08
real-if-close when drawing a MeshTet 

gdmcbain:
This is a minimal alternative to #611 for fixing #609, using `real_if_close` rather than `real`.  There are other aspects of #609 discussed there and under #611 though.


gdmcbain:
I prefer this to #611 but think #613 is better than either for the moment.

Issue #75, opened 2018-10-18
Investigate ex21 and add new convergence test


kinnala:
Checked the eigenvalues in ex17 and they seem to be a bit high for my taste, so I'm not sure whether the equations are correct or not.

Found an analytical test case from https://cfwebprod.sandia.gov/cfdocs/CompResearch/docs/elast.pdf.

Todo: implement a convergence test.

kinnala:
Either this was renamed to ex21 at some point or there was a typo.

gdmcbain:
Picking up here from the duplicate #318.

I recall that I also noticed some anomalies in gdmcbain/fenics-tuto-in-skfem#3 in translating the FEniCS tutorial 06 elastostatic cantilever problem.  There it was FEniCS that was inaccurate but the anomaly went away when its mesh was refined.

I'm a bit worried by this issue.  I'll try to get to bottom of it.  I assume it's not `scipy.sparse.linalg.eigsh`, although there are different ways to use this; I'll check that by trying a direct solver on a big machine, on a smaller problem if required.  What else?  Reduced quadrature?  Basis functions or their gradients?  Definition of linear elastic forms?  Everything, I guess, and all of them could have further consequences!  I'll look into this, as a matter of urgency.

gdmcbain:
Ta for Day & Romera (2004).  Rather than exact solutions to a continuous problem though, I'm particularly interested in exact or at least accurate eigensolutions of the discrete finite element problem on a definite mesh with definite elements.

There's some on a very simple geometry and mesh in:

* Ghodge, V., Bhattu, A. & B.Patil, S. (2018). Vibration analysis of beams. _International Journal of Engineering Trends and Technology,_ 55, 81–86. [doi:10.14445/22315381/IJETT-V55P216](https://doi.org/10.14445%2F22315381%2FIJETT-V55P216) 
 
It's a 550 × 50 × 5 mm cuboidal cantilever, fixed at one of the small ends, meshed with 44×4×1 hexahedra.  I presume these are S2 #314.  The eigenfrequencies quoted for this very coarse mesh are similar to the analytical values from one-dimensional Euler–Bernoulli beam theory.  I've had these checked in ANSYS and it gives the same values for 'linear' elements which I presume are the same as `ElementHex1`.  I'll attach the scikit-fem implementation here once I tidy it up; however, in brief, it's giving the same sort of misbehaviour as ex21.  (Whereas, as noted in the duplicate #318, ex03 seems fine.)

[...]

PR #613, opened 2021-04-08
revert to condense from enforce, passing symmetric solver

gdmcbain:
This takes ex21 back to where it was before #591 and #596, thereby avoiding having to excise the spurious imaginary part introduced by the asymmetric eigensolver as in #611 or #612.

Eventually it might be possible to `enforce` rather than condense the Dirichlet conditions, but I don't think it's quite clear yet how to do this for generalized eigenvalue problems.

### DIFF
--- a/docs/examples/ex03.py
+++ b/docs/examples/ex03.py
@@ -3,7 +3,6 @@
 from skfem import *
 from skfem.helpers import dot
 from skfem.models.elasticity import linear_elasticity
-from scipy.sparse.linalg import eigsh
 import numpy as np
 
 m1 = MeshLine(np.linspace(0, 5, 50))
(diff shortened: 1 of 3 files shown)


kinnala:
I did not observe any spurious imaginary part, just a different dtype with zero imaginary part, but I think you are correct here in the sense that this is the proper way to solve a symmetric problem.

I changed the behavior of `solve` because I think it should deal with the most general case by default or otherwise the result can be surprising: wrong results without any warning.

gdmcbain:
> I did not observe any spurious imaginary part, just a different dtype with zero imaginary part, 

Ah, right, I didn't check that the spurious imaginary part was actually zero.

> but I think you are correct here in the sense that this is the proper way to solve a symmetric problem.

Yeah, there's still all the other problems of #75 and I haven't totally decided that `enforce` can't work for something like this if used correctly; we'll see whether its original proponent chimes in.  It will introduce zero eigenvalues for the Dirichlet degrees of freedom but that doesn't necessarily matter.  I hadn't at first realized that #596 was only replacing the rows and not the columns.  I think in elementary presentations (reference?) the column is also replaced so that symmetry is preserved but of course that's too inefficient for a `csr_matrix` so symmetry is sort-of abandoned but not really as described by E. & G. &sect;8.4.3, &c., &c.

So I think this PR is the right thing to do in the short term (get the example running correctly without warnings or errors or makeshifts), but it might be bettered.

> I changed the behavior of solve because I think it should deal with the most general case by default or otherwise the result can be surprising: wrong results without any warning.

Yes, this was the right thing to do, for sure.

Issue #531, opened 2020-12-23
skfem.ivp


gdmcbain:
From #529 at [2020-12-24](https://github.com/kinnala/scikit-fem/issues/529#issuecomment-750296567):

> Note that we don't have any helper functions for time integration yet. That would be a welcome improvement though, e.g., `skfem.ivp` or something.

kinnala:
It's fine if most of the functionality are wrappers to stuff like https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.RK45.html but I think having a consistent interface (easy to experiment with the methods and parameters) which is specifically designed for finite element calculations (e.g. use of mass matrices in Mu'(t) = Au(t) and especially sparse matrices) would be beneficial.

gdmcbain:
This is quite a big issue.  I think getting the right abstraction could be very helpful but I fear that it's hard; not many packages get this right.  Some of the big questions:

* The entire [scipy.integrate.solve_ivp](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html#scipy.integrate.solve_ivp) framework is useless here since it assumes the form *y* ' = *f* (*y, t*) which doesn't cover finite element systems with unlumped mass matrices.  

* Even then, I find it restrictive that it requires `t_span`, the &lsquo;interval of integration&rsquo;; very often one would like to integrate until some condition is met, e.g. a steady state is obtained.

* An initial value problem framework should work together with other frameworks, for related problems like
  * stationary nonlinear problems (cf. #119)
    * continuation
      * natural parameter
      * pseudotransient
  * eigenproblems
  * ...

All of the above should really be outside the scope of scikit-fem but although I've been looking for a couple of decades, I haven't really found anything suitable.  There's [PETSc](https://www.mcs.anl.gov/petsc/) but it's a bit heavy.  There was [GarlicSim](https://github.com/cool-RR/GarlicSim) but it had already been abandoned before I found it.

gdmcbain:
On the `t_span` question, I'm very much informed by the  [example](http://www.r6rs.org/final/html/r6rs/r6rs-Z-H-18.html#node_chap_D) at the end of the _Revised Report on the Algorithmic Language Scheme_, in particular that:

>  The value returned by `integrate-system` is an infinite stream of system states.

In Python this is naturally a generator, like `ex19.evolve` without the `while`-termination criterion; i.e. something like:
```python
def evolve(t: float,
[...]

Issue #529, opened 2020-12-22
How can I insert the Coordinate and Element arrays in a simple poisson problem in scikit fem? 


nikosmatsa:

nikosmatsa:
Want to implement a simple poisson problem $-\Delta u = 10$ in a L shape domain given two matrices with coordinates and elements as shown in the link below (my so far effort).The problem is how can I insert them in order to have these specific coordinates?  

https://github.com/nikosmatsa/Finite-Differences-Methods-for-PDE-/blob/main/Finite%20Element%20.ipynb

kinnala:
Try feeding those to the default constructor of `MeshTri` as follows:
```python
MeshTri(coordinates.T, elements.T - 1)
```
Note that indices in Python are zero-based.

nikosmatsa:
it worked.

but I don't know if in the simple parabolic problem $$(u_{t}-u_{xx}=10)$$, with the same coordinates are right for (Crank Nicolson Method) below the Poisson problem 



https://github.com/nikosmatsa/Finite-Differences-Methods-for-PDE-/blob/main/Finite%20Element%20.ipynb

gdmcbain:
No?  Looks O. K., I think?  It's pretty much the same as [ex19](https://scikit-fem.readthedocs.io/en/latest/listofexamples.html#example-19-heat-equation), isn't it?

nikosmatsa:
I plot(m,u) not plot(m,x) as in poisson.

gdmcbain:
Ah, O. K., yes, sorry.  I went with `u` in ex19 rather than `x` as in, e.g., ex01, to avoid confusion with the `x` coordinate as in, e.g., ex03 or ex14, though of course `x` is standard for the unknown in linear algebra; e.g. [`scipy.sparse.linalg.spsolve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.spsolve.html).  Naming things is hard&hellip;

nikosmatsa:
But again i need confirmation.as my heat equation ut-uxx=10 is correctly implemented given these coordinates
'

gdmcbain:
[...]

PR #622, opened 2021-04-12
Do not overwrite by default

### LINKED ISSUE #620: overwrite_a: bool = False
In investigating the behaviour of the new #596 `skfem.utils.enforce` function [at the end of last week](https://github.com/kinnala/scikit-fem/issues/591#issuecomment-816408235), I wasted a couple of hours tracking down discrepancies with condensing and penalizing Dirichlet conditions because of not realizing that in a call like 
```python
L0, M0 = enforce(L, M, D=basis.find_dofs())
```
`L` and `M` would also be mutated, though I do now read that

https://github.com/kinnala/scikit-fem/blob/13f5630bd0077a1484a6ee392c219b4570dcb591/skfem/utils.py#L290-L293

I had been assuming that as discussed in #591 on [2021-03-25](https://github.com/kinnala/scikit-fem/issues/591#issuecomment-806409380)

> it should have the same signature as `condense`

I understand that ([ibid](https://github.com/kinnala/scikit-fem/issues/591#issue-840495893))

> Advantages of this approach include the fact that neither `A` nor `b` need to be rewritten in memory.

but I don't think any of the NumPy and SciPy routines that scikit-fem is built on mutate their arguments, and other functions from those libraries that do, for example, the dense [`scipy.linalg.solve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html#scipy.linalg.solve) only does so if `overwrite_a` or `overwrite_b` is `True` and both are `False` by default.

I think it does happen fairly often that an assembled finite element matrix is reused for the different things.  Of course mutation can be avoided in such cases by copying first, but only if one realizes that there is a threat of mutation.  Would it be safer to introduce overwrite-flags that are `False` by default?

I had a quick look at the implementation of `skfem.utils.enforce` and it doesn't look as though this would be too intrusive.

https://github.com/kinnala/scikit-fem/blob/13f5630bd0077a1484a6ee392c219b4570dcb591/skfem/utils.py#L326-L331

I think the assembly of `idx` and `d` would be the same but one only modifies `A` if `overwrite_A` and otherwise returns a fresh matrix with the condition enforced.

kinnala:
Fixes #620

### DIFF
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -284,7 +284,8 @@ def enforce(A: spmatrix,
             x: Optional[ndarray] = None,
             I: Optional[DofsCollection] = None,
             D: Optional[DofsCollection] = None,
-            diag: float = 1.) -> EnforcedSystem:
+            diag: float = 1.,
+            overwrite: bool = False) -> EnforcedSystem:
     r"""Enforce degrees-of-freedom of a linear system.
 
     .. note::
@@ -307,6 +308,10 @@ def enforce(A: spmatrix,
     D
         Specify either this or ``I``: The set of degree-of-freedom indices to
         enforce (rows/diagonal set to zero/one).
+    overwrite
+        Optionally, the original system is both modified (for performance) and
+        returned (for compatibility with :func:`skfem.utils.solve`).  By
+        default, ``False``.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


kinnala:
It has only single parameter. Is there a need for two?

skfem/utils.py
```diff
@@ -307,39 +306,47 @@ def enforce(A: spmatrix,
...
-    A.setdiag(d)
+    Aout.setdiag(d)
 
-    if b is not None:
-        if isinstance(b, spmatrix):
+    if bout is not None:
+        if isinstance(bout, spmatrix):
             # eigenvalue problem
-            b = enforce(b, D=D, diag=0.)
+            bout = enforce(bout, D=D, diag=0.)
```
gdmcbain, line 343: Should `overwrite=overwrite` be passed on the recursive call?

gdmcbain (changes requested):
As in `overwrite_a` and `overwrite_b` like [`scipy.linalg.solve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html#scipy.linalg.solve)?  I don't think so; I can't think of a case where one would want to overwrite one but not the other but I'm probably not the best one to ask as I'm rather averse to mutation.

skfem/utils.py
gdmcbain: Actually no, there's something else.  On stepping through this with pencil and paper, for `b: spmatrix` I see:
* `overwrite=False`: `bout` is a fresh copy of `b` so actually could well be overwritten but the recursive call has `overwrite=False` and so a second copy of `b` is made in `Aout = A if overwrite else A.copy()`.  That doesn't seem right.  Well, it's not wrong but making a second copy seems wasteful.
* `overwrite=True`: `bout = None` so `bout is not None` fails and only `Aout` is returned.  That's wrong; we do want to return a mass matrix with the Dirichlet condition enforced, i.e. with the Dirichlet rows zeroed.

gdmcbain: *Aside:&mdash;* I wonder whether it would be clearer to separate the two uses of `b`, instead of having it stand for either right-hand side vector or mass matrix.  Perhaps adding the argument:
```python
M: Optional[spmatrix] = None
```
for the mass matrix, as in [`scipy.sparse.linalg.eigs`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.eigs.html#scipy.sparse.linalg.eigs).

This might be a bit of an overhaul though; I haven't thought it through.

Note that a case with `spmatrix` `b` (or `M`) is not necessarily a generalized eigenvalue problem, it might be an initial value problem #531 , like ex19, and in that case we might have both a mass matrix and a nonzero right-hand side vector.  I think that something like that was asked for in #529, like a heat equation with a source term.

Issue #628, opened 2021-04-21
MeshLine.element_finder


gdmcbain:
While working on the idea that a point-source vector #561 might be the transpose of the probes matrix #601, I tried out a simple one-dimensional example and found something anomalous:
```python
MeshLine(np.linspace(0, 1, 3)).element_finder()(np.linspace(0, 1, 4))
```
returns
> [1 1 0 0]

gdmcbain:
Since `p` is
> [[0.  0.5 1. ]]
and `t` is
```python
 [[0, 1],
  [1, 2]]
```
the expected result was `[0, 0, 1, 1]`.

Issue #629, opened 2021-04-21
point source


gdmcbain:
I commented on the old issue #561 

> Is the point-source vector just the transpose of the probe matrix #601?

but will open a new issue to investigate this.  It might be nice feature to add to the library, or at least demonstrate in an example.

The thinking was that interpolating a function *u* at a point **p** is like (*u*, &delta;(**x** &minus; **p**)) whereas a point-source right-hand side vector is like (&delta;(**x** &minus; **p**), *v*).

gdmcbain:
To experiment with this, I considered the minimal example &minus;*u*'' = &delta;(*a*) on 0 < *x* < 1 with *u* (0) = *u* (1) = 0 and say &frac12; < *a* < 1.  The exact solution is *u* (*x*) = min {(1 &minus; *a*)*x*, (1 &minus; *x*)*a*}, which is piecewise linear, vanishes at the termini, and takes maximum (1 &minus; *a*)*a* at the break-point *x* = *a*.

For 
```python
mesh = MeshLine(np.linspace(0, 1, 3))
basis = InteriorBasis(mesh, ElementLineP1())
```
The laplacian `asm(laplace, basis)` is
```python
[[ 2. -2.  0.]
 [-2.  4. -2.]
 [ 0. -2.  2.]]
```
and by hand the RHS should be `[0, 1, 1] / 2` but `basis.probes(np.array([[3/4]]).toarray()[0]` returns
> [-0.5, 1.5, 0]
which is nothing like it, but also a bit surprising.  A bug in `MeshLine.element_finder` is suspected #628 and if fixed as in #630, the probe array is indeed `[0, 0.5, 0.5]`.

gdmcbain:
The anomalous probe vector `[-0.5, 1.5, 0]` for probe-point _x_ = &frac34; does actually give the correct interpolation, but it's linearly extrapolating from the other element rather than interpolating in the element containing the probe-point.

gdmcbain:
I read in FEniCS's [Changelog](https://fenics.readthedocs.io/projects/dolfin/en/latest/ChangeLog.html?highlight=pointsource#id19) for version 0.9.11 (2011-05-16)

> Add support for point sources (`class PointSource`)

but again I refrain from looking at the GPL implementation.

PR #630, opened 2021-04-21
Meshline element finder

gdmcbain:
For #628.

[`numpy.digitize`](https://numpy.org/doc/stable/reference/generated/numpy.digitize.html?highlight=digitize#numpy.digitize) returns the right end-point, so compare with `self.t[1]` rather than `self.t[0]`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Unreleased
 
+- Fixed: `MeshLine1.element_finder` 
+
 ### [3.0.0] - 2021-04-19
 
 - Added: Completely rewritten `Mesh` base class which is "immutable" and uses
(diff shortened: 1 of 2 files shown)


kinnala:
I suppose there was no test to catch this.

gdmcbain:
Yeah&hellip;, it would be good to have (and to have had) a test in place; I can't think of one that's not tautologous though.  If it had have been tested via testing `InteriorBasis.interpolate`, e.g. in exx 15, 16, 23, 29, 34, that might not have caught this because of the unexpectedly correct extrapolation [noted](https://github.com/kinnala/scikit-fem/issues/629#issuecomment-823712997) in #629. 

Perhaps I'll take a look at 

https://github.com/kinnala/scikit-fem/blob/13b5d697d41d96417abb3288fa524bb4f02290b9/docs/examples/ex34.py#L9

That will also see whether interpolation works with `ElementLineHermite` which might shed light on the [nonlagrangian question](https://github.com/kinnala/scikit-fem/pull/632#discussion_r617512290) of #632.

Issue #561, opened 2021-02-14
Point source vector / old scikit-fem version


jwittke:
Hey,
I just want to start my rather older code which uses a point source function. It was developed with an older version of
scikit-fem version but with the actual version it doesn't run anymore.

Any tips on how to correct the function below, to make it work with the new version?

    def dc_point_sources(basis, amplitude, point):
        pt = np.vstack([np.array([point]), np.array([0.0])])
        tris = find_element(basis.mesh, point)
        y = np.zeros(basis.N)

        pts = basis.mapping.invF(pt[:, :, np.newaxis], tind=tris)

        for k in range(basis.Nbfun):
            phi = basis.elem.gbasis(basis.mapping, pts, k, tind=tris)
            y[basis.element_dofs[k, tris]] += phi[0].flatten()

        return y * amplitude / 2.

Tanks,
Jan

kinnala:
What kind of error you get? There are some things missing for me to run it myself.

One idea is that it could've stopped working after `elem.gbasis` started returning tuples in `1.0.0`. So maybe changing `phi[0].flatten()` to `phi[0][0].flatten()` works?

jwittke:
Running gave me the error:

    y[basis.element_dofs[k, tris]] += phi[0].flatten()
    AttributeError: 'DiscreteField' object has no attribute 'flatten'

Changing `phi[0].flatten() `to `phi[0][0].flatten()` solves the problem.

Thanks

gdmcbain:
Is the point-source vector just the transpose of the probe matrix #601?

Issue #417, opened 2020-07-08
MeshLine.element_finder


gdmcbain:
It's impossible to construct an `InteriorBasis.interpolator` when `.mesh` is `MeshLine` since the latter has no `.element_finder`.

Hardly urgent since the nodal values are usually good enough.  What I was actually trying to do just now was plot the `ElementLineHermite` shape functions by interpolating onto a fine `numpy.linspace` of points from a `MeshLine` of a single cell with a cardinal vector for the degrees of freedom; I'd used `.lbasis` for `ElementLineP1` and `ElementLineP2` but `ElementLineHermite` doesn't have an `.lbasis`.  There might be another way to do that but anyway I figure the `.element_finder` might come in handy for something else later.

gdmcbain:
So far only `MeshTri` and `MeshQuad` have an `.element_finder`; those for `Mesh3D` might come in handy too later.

gdmcbain:
Partly looking ahead to `MeshTet` (and also hoping for an implementation independent of dimension), I looked at [scipy.spatial.Delaunay.find_simplex](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.Delaunay.find_simplex.html#scipy.spatial.Delaunay.find_simplex) but `Delaunay` doesn' t work in one dimension; `Delaunay(skfem.MeshLine().p.T)` raises `*** ValueError: Need at least 2-D data`.

kinnala:
I think we could make an algorithm based on octrees that finds the correct element efficiently in 3D.

gdmcbain:
I have some particle-tracking applications for which that would be nice although thus far I really only have the flow-fields in two dimensions and they're already covered.

I wonder whether the octree would generalize to *n* dimensions. (Specifically here including *n*=1.)

kinnala:
Yes, I think 1, 2, and 3 dimensions should work. The question is that whether it's worth the trouble to build a vectorized octree implementation ourselves or is there an implementation available somewhere in the ecosystem that we can use?

gdmcbain:
Yes, I'd think that there ought to be one out there. I haven't found it yet but then I don't need it right away. It probably would be better to leave it to a heavily optimized and tested package if one were found as in some applications (mapping between meshes?) it would get hammered.

PR #632, opened 2021-04-21
InteriorBasis.point_source

### LINKED ISSUE #629: point source

gdmcbain:
Fix #629 and exemplify.

ex38

This won't work for `MeshLine` until #628 is resolved, but is fine for `MeshTri`, &c.

Should probably compare the finite element solution with Green's function, as discussed in the docstring.  For a test, I'd suggest one dimension, pending #628.


skfem/assembly/basis/interior_basis.py
```diff
@@ -154,6 +154,10 @@ def probes(self, x: ndarray) -> coo_matrix:
             shape=(x.shape[1], self.N),
         )
 
+    def point_source(self, x: ndarray) -> ndarray:
+        """Return right-hand side vector for unit source at `x`"""
```
kinnala, line 158: Is this docstring true for other than P1 Lagrange?

gdmcbain: Hmm, I think so, in the Galerkin sense. Initially I reread Jan's point-source function in #561 and was reminded of what we'd done for the probe for Nico. Then i thought that at the continous variational level, interpolation is the inner product of a finite element function with a Dirac delta and a point source is the inner product of a Dirac delta with a test function. Those are adjoints and on discretization they're transposes. That line of thoight doesn't make any reference to the txpe of basis.

Similarly for `skfem.models.poisson.unit_load` and the functional that integrates, replacing the Dirac delta with the constant 1.

gdmcbain: I don't have a reference for this. I couldn't find one in a quick look. I remember that for years users were asking about point sources on other finite element msiling lists. Somtimes the discussion suggested mollification: convolving the point source with a Gaussian kernel, but I didn't like that so much as it still didn't seem well suited to the quadrature.  Then FreeFEM added one recently; I refrained from looking at  their implementation as it's GPL.

gdmcbain: Possible hole in the argument: the argument shows that the transpose property holds but it only follows from that that the function returns the Galerkin point source vector if the probes function works correctly. I didn't look into whether that works for all kinds of bases. Does it? I assumed that it did.

gdmcbain: A couple of brief references to point sources in some of the older FEM books on my self:
* Segerlind, L. J. (1976). *Applied Finite Element Analysis.* New York: Wiley, &sect;8.6 &lsquo;Point sources&rsquo;
* Hughes, T. J. R. (2000). *The Finite Element Method.* Mineola, New York: Dover, exercise 3.11.2(*b*)

Unfortunately Hughes doesn't give the solution to the exercise (delta function loading on a one-dimensional, quadratic, three-node element), but I think it's looking the similar to my pair of [linear two-node elements](https://github.com/kinnala/scikit-fem/issues/629#issuecomment-823701775).

Segerlind notes:

> The integral of a function multiplied by an impulse function, however, is simply the function evaluated at <i>X</i><sub>0</sub> and <i>Y</i><sub>0</sub>.

which is elementary and the subsequent example is only for `ElementTriP1`.

I'm yet to turn up anything on nonlagrangian elements.

gdmcbain:
```python
from pathlib import Path

import numpy as np

import skfem
from skfem.models import laplace
from skfem.visuals.matplotlib import plot, savefig

mesh = skfem.MeshLine(np.linspace(0, 1, 2 ** 1 + 1))
basis = skfem.InteriorBasis(mesh, skfem.ElementLineHermite())

lap = skfem.asm(laplace, basis)
print(lap.toarray())

a = 3 / 4
rhs = basis.point_source(np.array([a]))
print("RHS:", rhs)
condensed = skfem.condense(lap, rhs, D=basis.find_dofs())
print(condensed[0].toarray())
print(condensed[1])

solution = skfem.solve(*condensed)
print(solution)

ax = plot(basis, solution, Nrefs=3)
ax.set_title(f"{type(basis.elem)}")
x = np.linspace(mesh.p[0].min(), mesh.p[0].max())
ax.plot(x, np.stack([(1 - a) * x, (1 - x) * a]).min(0), linestyle="--", label="exact")
ax.plot(
    mesh.p[0],
    solution[basis.nodal_dofs[0]],
    marker="o",
    linestyle="None",
    label="nodal DoFs",
)
ax.legend()
savefig(Path(__file__).with_suffix(".png"))
```

mwe

I think that that's doing the right thing.  The `plot` looks a bit funny though. I copied from

https://github.com/kinnala/scikit-fem/blob/13b5d697d41d96417abb3288fa524bb4f02290b9/docs/examples/ex34.py#L47

`ElementLineMini` looks O. K. too.

mwe

gdmcbain:
I mean Hermite elements are not a good choice for this problem since they're C<sup>1</sup> and the solution isn't, but they're trying:

mwe

Actually
```python
condensed = skfem.condense(lap, rhs, D=basis.find_dofs())
```
isn't the right boundary condition for Hermite either, that's for a built-in Euler&ndash;Bernoulli beam, isn't it.

gdmcbain:
Perhaps I'll look at the elastic plate with `ElementTriMorley` too, adapting ex18.
```python
from skfem import *
from skfem.helpers import ddot, dd

import numpy as np


mesh = MeshTri.init_circle(4)
element = ElementTriMorley()
mapping = MappingAffine(mesh)
ib = InteriorBasis(mesh, element, mapping, 2)


@BilinearForm
def biharmonic(u, v, w):
    return ddot(dd(u), dd(v))


stokes = asm(biharmonic, ib)
psi = solve(*condense(stokes, ib.point_source(np.array([0.2, 0.3])), D=ib.find_dofs()))


if __name__ == "__main__":
    from os.path import splitext
    from sys import argv
    from skfem.visuals.matplotlib import draw, show
    from matplotlib.tri import Triangulation

    M, Psi = ib.refinterp(psi, 3)

    ax = draw(mesh)
    ax.tricontour(Triangulation(*M.p, M.t.T), Psi)
    name = splitext(argv[0])[0]
    show()
```

ex18_point-source

kinnala:
Looks good. Ready for merge or still a draft?

kinnala:
I've actually used this kind of load in some of the articles on Kirchhoff plates and always hacked it together in nonreusable way. This would've really simplified that. Thanks!

gdmcbain:
> Looks good. Ready for merge or still a draft?

I should add a test.

I'll add a comparison with the Green's function to the example later.

I assume the Green's function for the Kirchhoff plate is known too.  Have you got a reference to hand?  Actually that might make a nicer example than the membrane, showing right away that it works for other than Lagrange elements.

kinnala:
> I assume the Green's function for the Kirchhoff plate is known too. Have you got a reference to hand?

I think there is but I have no reference at hand.

On the other hand, there is an infinite series solution available for a square plate and a point load. The series is not converging very quickly but it's something if you truncate it to a sum. See Section 5.1 (eq. 5.13) in this preprint: https://arxiv.org/pdf/1707.08396.pdf

gdmcbain:
Hang on, something goes wrong if I change from `MeshLine(np.linspace(0, 1, ...))` to `MeshLine().refined(...)`.
mwe

gdmcbain:
```
>>> m = fe.MeshLine().refined()
>>> fe.InteriorBasis(m, fe.ElementLineP1()).probes(np.array([[3/4]])).toarray()
array([[-0.5,  0. ,  1.5]])
>>> fe.InteriorBasis(fe.MeshLine(m.p[0]), fe.ElementLineP1()).probes(np.array([[3/4]])).toarray()
array([[0. , 0.5, 0.5]])
```

gdmcbain:
Is the second element in `MeshLine().refined()` everted?
```python
>>> m.t
array([[0, 2],
       [2, 1]])
```

gdmcbain:
It looks like `MeshLine.element_finder` #417 #628 again.
```
>>> m = MeshLine().refined()
>>> m.element_finder()(np.array([.7]))
array([0])
>>> MeshLine(m.p[0]).element_finder()(np.array([.7]))
array([1])
```

kinnala:
> It looks like `MeshLine.element_finder` #417 #628 again.
> 
> ```
> >>> m = MeshLine().refined()
> >>> m.element_finder()(np.array([.7]))
> array([0])
> >>> MeshLine(m.p[0]).element_finder()(np.array([.7]))
> array([1])
> ```

There is `Mesh.sort_t` flag which defaults to `False`. `MeshTri1` has `sort_t=True` which is explained in the docstring of `Mesh`.

gdmcbain:
Let's take that up in #633.

kinnala:
Let me add, `sort_t=False` is something I'd prefer to have everywhere but it wasn't easy to drop it from `MeshTri1` because other code assumes it. So maybe sorting in `element_finder` is sufficient?

gdmcbain:
It remains to include the example in `tests/test_examples` and before that to put something testable in the example, e.g. comparison of the solution with the analytic Green's function (once I track down a reference for that&mdash;basically it's the usual logarithmic source minus an image sink on the other side of the circumference).

gdmcbain:
I still haven't been able to find a better reference for the Green's function (Sadybekov et al just quote the result without reference or derivation), but this is it anyway.

gdmcbain:
The PNG [above](https://github.com/kinnala/scikit-fem/pull/632#issue-619872748) will do for the Gallery.

kinnala:
> The PNG [above](https://github.com/kinnala/scikit-fem/pull/632#issue-619872748) will do for the Gallery.

Sure! Can you add it to `docs/listofexamples.rst`?

PR #634, opened 2021-04-22
Do not assume any order in MeshLine1.element_finder

### LINKED ISSUE #633: MeshLine.refine
While working on the point source #632, trouble was encountered with `MeshLine().refined()` and this was traced to `element_finder`:
```python
>>> m = MeshLine().refined()
>>> m.element_finder()(np.array([.7]))
array([0])
>>> MeshLine(m.p[0]).element_finder()(np.array([.7]))
array([1])
```

I'm not sure which of these is right.  I mean the first is wrong since 0.7 is definitely in the second element, but I'm not sure about the second.  Are the points passed to the constructed assumed increasing?

kinnala:
Fixes #633

### DIFF
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -1316,12 +1316,14 @@ class MeshLine1(Mesh):
 
     def element_finder(self, mapping=None):
         ix = np.argsort(self.p)
+        maxt = self.t[np.argmin(self.p[:, self.t], axis=1)[0],
+                      np.arange(self.t.shape[1])]
 
         def finder(x):
             maxix = (x == np.max(self.p))
             x[maxix] = x[maxix] - 1e-10  # special case in np.digitize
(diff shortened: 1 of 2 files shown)


skfem/mesh/mesh.py
```diff
@@ -1316,12 +1316,14 @@ def param(self):
 
     def element_finder(self, mapping=None):
         ix = np.argsort(self.p)
+        mint = self.t[np.argmin(self.p[:, self.t], axis=1)[0],
```
gdmcbain, line 1319: We could replace `np.argmin(self.p[:, self.t], axis=1)[0]` with
```python
np.argmin(self.p[0, self.t], axis=1)
```
but this line might need changing anyway.

gdmcbain (changes requested):
I think that we do have to look for maxt rather than mint since `np.digitizer` returns the right end of the bin.  I know this didn't work last time #630 but I think I've got it going now.  PR soon.

PR #637, opened 2021-04-30
penalize

### LINKED ISSUE #621: penalizing Dirichlet conditions
Following the introduction of `skfem.utils.enforce` in #591 and #596 and subsequent investigation which involved the penalty method ([2021-04-09](https://github.com/kinnala/scikit-fem/issues/591#issuecomment-816408235)) for a very special example, I realized that an implementation of the `skfem.utils.penalize` function would be quite similar.  It would add a large positive number to the diagonal elements of the stiffness matrix corresponding to the Dirichlet degrees of freedom and in the static case multiply the corresponding entries of the right-hand side vector by the same factor.  The mass matrix would be left untouched in the unsteady or eigenvalue cases.

Possibly applications of this besides pedagogical might follow from the feature that it would act very like `enforce` but not destroy symmetry; e.g. as Ern & Guermond (2004, &sect;8.4.4 &lsquo;Penalty method&rsquo;) put it:

> This technique, which is very simple to implement, is useful when one wishes to utilize a direct solution method based on the symmetry of the matrix such as the Choleski or the LDL<sup>T</sup> factorization

Though I don't think we have access to the latter and haven't used the former since `sksparse.cholmod` was dumped #447.

I'll take a stab at this, probably with overwriting turned off by default. #620

gdmcbain:
To fix #621.

### DIFF
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -348,6 +348,66 @@ def enforce(A: spmatrix,
     return Aout
 
 
+def penalize(A: spmatrix,
+             b: Optional[Union[ndarray, spmatrix]] = None,
+             x: Optional[ndarray] = None,
+             I: Optional[DofsCollection] = None,
+             D: Optional[DofsCollection] = None,
+             epsilon: float = 1e-6,
+             overwrite: bool = False) -> LinearSystem:
+    r"""Penalize degrees-of-freedom of a linear system.
+
+    Parameters
+    ----------
+    A
+        The system matrix
+    b
+        Optionally, the right hand side vector.
+    x
+        The values of the penalized degrees-of-freedom. If not given, assumed
+        to be zero.
+    I
+        Specify either this or ``D``: The set of degree-of-freedom indices to
+        solve for.
+    D
+        Specify either this or ``I``: The set of degree-of-freedom indices to
+        enforce (rows/diagonal set to zero/one).
+    epsilon
+        Very small value, the reciprocal of which penalizes deviations from
+        the Dirichlet condition
+    overwrite
+        Optionally, the original system is both modified (for performance) and
+        returned (for compatibility with :func:`skfem.utils.solve`).  By
+        default, ``False``.
+
+    Returns
+    -------
+    LinearSystem
+        A linear system with the penalized diagonal and RHS entries set to
+        very large values, 1/epsilon and x/epsilon, respectively.
+
+    """
+    b, x, I, D = _init_bc(A, b, x, I, D)
+
+    Aout = A if overwrite else A.copy()
+
+    d = Aout.diagonal()
+    d[D] = 1/epsilon
+    Aout.setdiag(d)
+
+    if b is not None:
+        if isinstance(b, spmatrix):
+            # mass matrix (eigen- or initial value problem)
+            pass
+        else:
+            # set rhs to the given value
+            bout = b if overwrite else b.copy()
+            bout[D] = x[D] / epsilon
+        return Aout, bout
+
+    return Aout
+
+
 def condense(A: spmatrix,
              b: Optional[Union[ndarray, spmatrix]] = None,
              x: Optional[ndarray] = None,
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -354,7 +354,7 @@ def penalize(A: spmatrix,
              x: Optional[ndarray] = None,
              I: Optional[DofsCollection] = None,
              D: Optional[DofsCollection] = None,
-             epsilon: float = 1e-12,
+             epsilon: Optional[float] = None,
              overwrite: bool = False) -> LinearSystem:
     r"""Penalize degrees-of-freedom of a linear system.
 
@@ -392,7 +392,10 @@ def penalize(A: spmatrix,
 
     Aout = A if overwrite else A.copy()
 
+        
     d = Aout.diagonal()
+    if epsilon is None:
+        epsilon = 1e-10 / np.linalg.norm(d[D], np.inf)
     d[D] = 1/epsilon
     Aout.setdiag(d)
 
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -354,7 +354,7 @@ def penalize(A: spmatrix,
              x: Optional[ndarray] = None,
              I: Optional[DofsCollection] = None,
              D: Optional[DofsCollection] = None,
-             epsilon: float = 1e-6,
+             epsilon: float = 1e-12,
              overwrite: bool = False) -> LinearSystem:
     r"""Penalize degrees-of-freedom of a linear system.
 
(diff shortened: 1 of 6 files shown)


tests/test_manufactured.py
```diff
@@ -27,16 +55,17 @@ class Line1D(TestCase):
     Solution is u(x) = x.
 
     """
+
     e = ElementLineP1()
 
     def runTest(self):
-        m = MeshLine(np.linspace(0., 1.)).refined(2)
+        m = MeshLine(np.linspace(0.0, 1.0)).refined(2)
```
kinnala, line 62: I'm not sure, but it seems like you've run `pyblack` here. Do you think we should start using that project-wide?

gdmcbain (commented):
This does very nicely for an IVP #531 like ex19.  It avoids the repacking needed by `condense` and `enforce` is no good because `splu` wants a `csc_matrix` for which we need to transpose but that's not valid for `enforce` which leaves spurious offdiagonal entries on the columns of enforced degrees of freedom.

I haven't figured out an optimal default `epsilon` but 1e-12 allows the same test tolerance to be used as for `enforce` in a Poisson problem.

skfem/utils.py
```diff
@@ -348,6 +349,66 @@ def enforce(A: spmatrix,
     return Aout
 
 
+def penalize(A: spmatrix,
+             b: Optional[Union[ndarray, spmatrix]] = None,
+             x: Optional[ndarray] = None,
+             I: Optional[DofsCollection] = None,
+             D: Optional[DofsCollection] = None,
+             epsilon: float = 1e-6,
```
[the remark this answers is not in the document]
gdmcbain: I haven't figured out an optimal default epsilon but 1e-12 allows the same test tolerance to be used as for enforce in a Poisson problem.

skfem/utils.py
```diff
@@ -348,6 +349,63 @@ def enforce(A: spmatrix,
     return Aout
 
 
+def penalize(A: spmatrix,
+             b: Optional[Union[ndarray, spmatrix]] = None,
+             x: Optional[ndarray] = None,
+             I: Optional[DofsCollection] = None,
+             D: Optional[DofsCollection] = None,
+             epsilon: float = 1e-12,
```
kinnala, line 357: A bit worried about so small default value: does it not deteriorate conditioning?

kinnala: Let me add that "optimal" is something depending on the mesh parameter `h` but we should not worry about that here: a constant value is fine. Just wondering if 1e-12 causes surprising behavior when using iterative solvers.

gdmcbain: Well, I had thought so and hence had been using `skfem.utils.condense` consistently (and an equivalent in in-house codes for years before that), but I'm wondering whether the analysis of `enforce` doesn't carry across.

Conventionally anyway, I think penalization is more used with direct methods than iterative, so the conditioning should be less important.

> This technique, which is very simple to implement, is useful when one wishes to utilize a direct solution method based on the symmetry of the matrix such as the Choleski or the LDL<sup>T</sup> factorization (Ern & Guermond 2004, &sect;8.4.4 &lsquo;Penalty method&rsquo;)

I would like to have a more rational choice for `epsilon`.  I'm looking at Barrett & Elliott (1986) now.  I think they use a value that scales with `Mesh.param`; unfortunately the `Mesh` isn't available.  I suspect a rational choice can be inferred just from the matrix as the method can be conceived of as applying to the linear algebraic problem.

Of course it's only a default; the user can (and should?) vary it in applications.

Also I'm adding `penalize` mostly for completeness, pedagogy, and its usefulness in concise examples like ex19.  I wouldn't think it would be used much for real work.  Later perhaps we can continue #150 and add `constrain` for Lagrange multipliers, something for Nitsche (or is that too problem-dependent?).  Indeed, Ern & Guermond continue:

> If *&epsilon;*<sup>&minus;1</sup> is not large enough, the method suffers from a lack of consistency.  This problem can be avoided by adding extra boundary terms ensuring consistency as shown by Nitsche [Nit71](https://doi.org/10.1007%2fBF02995904)].

Note that they highlight a problem for when *&epsilon;* is not small enough, not when it is too small.

It definitely gives spurious eigenvalues corresponding to the Dirichlet degrees of freedom, but that's just the same as `enforce`; I would always use `condense` for eigenvalue problems, myself.

Have you got an idea for a value?  Another more stringent test, which would show whether the conditioning deteriorates?

*  Barrett, J. W. & Elliott, C. M. (1986). Finite element approximation of the Dirichlet problem using the boundary penalty method. *Numerische Mathematik,* 49, 343-366. [doi: 10.1007/BF01389536](https://doi.org/10.1007%2fBF01389536)

gdmcbain: (Sorry, I didn't see the second comment when replying to the first.)

Here we go, a more definite suggestion (Becker, Carey, & Oden 1981, &sect;3.8):

> In actual computations we need only choose the penalty number *&beta;*<sub>0</sub> [=diffusivity/*&epsilon;*] a few orders of magnitude larger than other terms in **K**.  It is important to knote that the results are insensitive to the actual choice of *&beta;*<sub>0</sub>.  In the coding given below for subroutine `DRCHLT` a value of *&beta;*<sub>0</sub> is chosen to be 10<sup>10</sup> times as large as the diagonal term of **K** to which *&beta;*<sub>0</sub> is added.

* Becker, E. B., Carey, G. F.,, Oden, J. T. (1981). *Finite Elements, An Introduction* (Vol. 1). Englewood Cliffs, New Jersey: Prentice Hall

gdmcbain: Running ex03 (eigensolvers are iterative) with `penalize`, `L = `

>  [0.00418289 0.1218215  0.2647111  0.69461021 1.92916584 2.36681119]

whereas with `condense`, 

> [0.00418289 0.12182152 0.26471113 0.69461028 1.92916598 2.36681145]

gdmcbain: Here's the three-dimensional iterative (PyAMG, AMGCL) ex09, first with `penalize`, run through `tail -n 12`:
```
cg converged to tol=default and atol=default
1.4170612924768282
1.441653529485872
1.4408692692575318
1.4406879483931736
1.440678669684838
cg converged to tol=default and atol=default
1.4320668891726949
1.4408270931593417
1.4406765881678443
1.4406791759864541
cg converged to tol=default and atol=default
```
and back with `condense`:
```
cg converged to tol=default and atol=default
1.4262254636438005
1.4421224866634006
1.4408563228909756
1.4406850811798213
1.4406784733853732
cg converged to tol=default and atol=default
1.4323602036346905
1.440799855252304
1.4406777821036938
1.4406791187336812
cg converged to tol=default and atol=default
```

gdmcbain: These aren't very fine meshes, but I think it's looking O. K.?  I do like the Becker&ndash;Carey&ndash;Oden suggestion better than a constant, and the diagonal entries of the matrix already scale with `Mesh.param`, so no need to re-inspect the mesh.

kinnala: Great!

PR #641, opened 2021-05-12
don't need ex10.I

gdmcbain:
I presume the use of `I` here is a vestige of pre-#228.


gdmcbain:
I was looking at ex10 to see whether it might be a good candidate to exemplify `enforce` #591, which doesn't currently appear in `docs`.  It works, simply replacing `condense` with `enforce`; indeed the output is identical to many decimal places, but I don't know whether it's worthwhile.  It doesn't simplify the user-level script as the use of `penalize` did ex19 in #637.

PR #650, opened 2021-06-04
doc updates

### LINKED ISSUE #648: instructiveness of first example
[The first example](https://github.com/kinnala/scikit-fem/blob/master/docs/examples/ex01.py) is not terribly instructive. Laplace is of course a good choice, anyone trying skfem for the first time will want to see trial and test functions, integration and such. `laplace` takes care (hides) all of this nicely, but I'd rather be explicit here and perhaps add a comment that this could be replaced with `laplace()`.

nschloe:
Fixes #648.


kinnala:
Thanks! This is a nice improvement.

PR #652, opened 2021-06-15
Rename InteriorBasis to CellBasis

### LINKED ISSUE #606: InteriorBasis?
When I first saw `skfem.InteriorBasis`, I got a little confused because I thought it doesn't take into account the dofs on the boundary. If I understand correctly, the name is just to set it apart from `FacetBasis` (which if I understand correctly only means the _boundary_ facets, right?). If this is true, I would suggest to rename `InteriorBasis`. How about `ElementBasis` or just `Basis`? `FacetBasis` could then be `SurfaceBasis` or `BoundaryBasis` or something like that.

kinnala:
Fixes #606

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Unreleased
 
+- Added: `CellBasis`, a new preferred alias for `InteriorBasis`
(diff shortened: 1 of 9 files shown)


kinnala:
1. Release this
2. Worry about docs and examples after

gdmcbain (approved):

PR #653, opened 2021-06-18
add ElementTetCCR and ElementTriCCR

bhaveshshrimali:
This PR adds two new elements `ElementTetCCR` and `ElementTriCCR` useful in approximating displacement/velocity in inompressible elasticity/stokes PDEs. These allow choosing discontinuous pressure formulations offered by `ElementTetDG` and `ElementTriDG` respectively. 

References: 

https://www.springer.com/gp/book/9783642870491: Implementation of Finite Element Methods for Navier-Stokes Equations, Thomasset, F. (Page-79, 3.2.3)

https://link.springer.com/book/10.1007/978-3-642-61623-5: Finite Element Methods for Navier-Stokes Equations, Vivette GiraultPierre-Arnaud Raviart (Page-139, 2.2)

Original paper: Conforming and nonconforming finite element methods for solving the stationary Stokes equations I, M. CROUZEIX, P.-A. RAVIART
http://archive.numdam.org/article/M2AN_1973__7_3_33_0.pdf

Expected convergence rates:
`ElementTetCCR`: Same as `ElementTriCCR` ~~(4.41-4.42 as given in the original paper)~~
	~~`rateL2`: 4~~
	~~`rateH1`: 3~~

`ElementTriCCR`: (4.34-4.35 as given in the original paper)
	`rateL2`: 3
	`rateH1`: 2

Note that the basis functions for `ElementTriCCR` are incorrect in the original paper. They are off by a sign. They should be: λᵢ⋅(2⋅λᵢ - 1) + 3⋅λ₁⋅λ₂⋅λ₃, whereas in the original paper they are given to be: λᵢ⋅(2⋅λᵢ - 1) - 3⋅λ₁⋅λ₂⋅λ₃ (if I notice it correctly). This is verified in the tests too (Partition of Unity), but here is a short snippet with the correct basis functions (and the ones with a typo)

```python3
In [1]: from sympy import symbols, simplify, init_printing

In [2]: init_printing(use_unicode=True)

In [3]: x, y = symbols("xi, eta", real=True)

In [4]: l1, l2, l3 = x, y, 1-x-y

In [5]: N_nodal = [eval(f"l{i}*(2*l{i}-1) + 3*l1*l2*l3") for i in range(1,4)]

In [6]: N_nodal
Out[6]: 
[3⋅η⋅ξ⋅(-η - ξ + 1) + ξ⋅(2⋅ξ - 1), 3⋅η⋅ξ⋅(-η - ξ + 1) + η⋅(2⋅η - 1), 3⋅η⋅ξ⋅(-η
 - ξ + 1) + (-2⋅η - 2⋅ξ + 1)⋅(-η - ξ + 1)]

In [7]: N_facet = [eval(f"4*l{i}*l{j} - 12*l1*l2*l3") for i in range(1, 4) for j in range(i+1,4)]

In [8]: N_facet
Out[8]: 
[-12⋅η⋅ξ⋅(-η - ξ + 1) + 4⋅η⋅ξ, -12⋅η⋅ξ⋅(-η - ξ + 1) + 4⋅ξ⋅(-η - ξ + 1), -12⋅η⋅
ξ⋅(-η - ξ + 1) + 4⋅η⋅(-η - ξ + 1)]

In [9]: N_bubble = 27*l1*l2*l3

In [10]: N_bubble
Out[10]: 27⋅η⋅ξ⋅(-η - ξ + 1)

In [11]: simplify(sum(N_facet) + sum(N_nodal) + N_bubble)
Out[11]: 1

In [12]: N_nodal = [eval(f"l{i}*(2*l{i}-1) - 3*l1*l2*l3") for i in range(1,4)]

In [13]: simplify(sum(N_facet) + sum(N_nodal) + N_bubble)
Out[13]: 
    2           2
18⋅η ⋅ξ + 18⋅η⋅ξ  - 18⋅η⋅ξ + 1

```


kinnala:
Can you add these new elements to the following lists:
- https://github.com/kinnala/scikit-fem/blob/master/tests/test_elements.py#L51
- https://github.com/kinnala/scikit-fem/blob/master/tests/test_elements.py#L183
- https://github.com/kinnala/scikit-fem/blob/master/tests/test_elements.py#L239

Also, can you add something like
```python
class ConvergenceTriCCR(ConvergenceTriP1):
    rateL2 = 2.0
    rateH1 = 1.0
    def create_basis(self, m):
        e = ElementTriCCR()
        return InteriorBasis(m, e)
 ```
 to `tests/test_convergence.py`.

File: skfem/element/__init__.py [hunk 1]
```diff
@@ -45,6 +45,7 @@
     ElementHex0: ElementQuad0,
     ElementHex1: ElementQuad1,
     ElementHex2: ElementQuad2,
+    ElementTetCCR: ElementTriCCR,
```
kinnala, line 48: If you want to include it here (for `InteriorBasis.trace` support), you should check that it works, e.g., by modifying https://github.com/kinnala/scikit-fem/blob/master/tests/test_basis.py#L193

If it doesn't work, then you should remove this line.

bhaveshshrimali: Ok. Will check. Thanks

skfem/element/element_tri/element_tri_ccr.py
```diff
@@ -0,0 +1,70 @@
...
+    refdom = RefTri
+
+    def lbasis(self, X, i):
+        x, y = X
+
+        if i == 0:  # (0., 0.)
+            phi = -(-2*x + y*(3*x - 2) + 1)*(x + y - 1)
+            dphi = np.array([-6*x*y + 4*x - 3*y**2 + 7*y - 3,
+                             -3*x**2 - 6*x*y + 7*x + 4*y - 3])
+        if i == 1:  # (1., 0.)
```
kinnala, line 31: should be `elif` ?

kinnala: maybe it doesn't matter though

kinnala: ahh, of course it does. this causes the index error

bhaveshshrimali: Ah yes, I think it got jumbled somewhere in copying and rearranging the order. Fixed.

skfem/element/__init__.py
bhaveshshrimali: Removing it for now. For the integration, should I add higher order quadrature points/weights?

skfem/element/element_tri/element_tri_ccr.py
bhaveshshrimali: And I think I messed up the convergence order of `ElementTriCCR` as well.. Hopefully should work now..

tests/test_convergence.py
```diff
@@ -219,6 +219,15 @@ def create_basis(self, m):
         e = ElementTriCR()
         return InteriorBasis(m, e)
 
+class ConvergenceTriCCR(ConvergenceTriP1):
+
+    rateL2 = 2.0
+    rateH1 = 1.0
```
kinnala: You should verify what L2 and H1 convergence rate to expect from the references. The test seems to suggest that `rateL2` should be 3.0. I can also check it later.

Btw, can you also add similar test `ConvergenceTetCCR`? See `ConvergenceTetCR` for an example.

bhaveshshrimali: Sure :) !!

bhaveshshrimali: I have added the convergence rates `rateL2` and `rateH1` for `ElementTetCCR` as well. Though, looking till 4.37 in the paper by Crouzeix and Raviart everything looks fine I will take a fresh look at the derivation tomorrow.

skfem/element/__init__.py
bhaveshshrimali: As for the higher order quadrature rules, I took these verbatim from [`ffc`](https://bitbucket.org/fenics-project/ffc/src/cb26c917c89bad1e5888836378cd3b395c077227/ffc/fiatinterface.py#lines-208), namely

```python3
from ffc.fiatinterface import create_quadrature
deg = 8 # 2, 3, 4 ...
points, weights = create_quadrature("tetrahedron", deg, "default")

```

kinnala:
Fantastic! Thanks for the contribution.

skfem/element/__init__.py
kinnala: I'll verify the original source.

kinnala: Original source for 6th and 7th order rule: https://web.archive.org/web/20190609142211/http://people.sc.fsu.edu:80/~jburkardt/datasets/quadrature_rules_tet/quadrature_rules_tet.html

kinnala: 8th order rule is something else, I'll try to verify.

kinnala: Seems to be collapsed tensor rule: https://github.com/FEniCS/fiat/blob/7d418fa0a372ac6f5e103533ab77ad6a9fac764c/FIAT/quadrature.py#L280

kinnala: I wonder if this 8th order Keast rule is better? https://web.archive.org/web/20120104033925/http://people.sc.fsu.edu/~jburkardt/datasets/quadrature_rules_tet/keast8_x.txt

kinnala: There seems to be much larger number of points in the rule used in this PR. Collapsed rules seem to be that way. Let us verify if it's better to use rules from https://web.archive.org/web/20120104033925/http://people.sc.fsu.edu/~jburkardt/datasets/quadrature_rules_tet/keast8_x.txt

bhaveshshrimali: > Collapsed rules seem to be that way. Let us verify if it's better to use rules from web.archive.org/web/20120104033925/http://people.sc.fsu.edu/~jburkardt/datasets/quadrature_rules_tet/keast8_x.txt

Should I take a crack at this or leave it for the future? Do all the convergence tests pass without the 8th order rules?

bhaveshshrimali:
> Fantastic! Thanks for the contribution.

Thanks for considering it. It was on my list of TODOs since almost last year but couldn't find time to put it together until now. Having this element in `scikit-fem` should help me in my research too.

skfem/element/__init__.py
bhaveshshrimali: I decided to use the 8th order keast rule and it seems to pass the tests for convergence (at least locally for the moment). 

```
======================================================================================================================= test session starts =============================================================================================================================================================
platform win32 -- Python 3.9.2, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: D:\Research\scikitFEMContrib\scikit-fem\tests
plugins: anyio-2.2.0, hypothesis-6.13.7, cov-2.11.1, dependency-0.5.1
collected 25 items
                                                                                                                                                                                                                                                              [100%]
test_convergence.py .........................
                               [100%]                                                                                                                                                                                          =====================================

================================================================================================================== 25 passed in 98.99s (0:01:38) ===================================================================================================================

```

kinnala: It seems that some of these quadrature rules have negative weights: https://github.com/kinnala/scikit-fem/pull/653/files#diff-a32c8dc529b4ef2327e933b5e85b5ec80e8a108709027ef19e83ec7df8625fbbR574

It's not often an issue, but sometimes if you are performing minor variational crimes, say that your material field changes discontinuously inside the element and you are not meshing exactly the discontinuous interface, then this could cause non-positive-definite stiffness matrix.

However, we can leave it for now and fix it later if we notice any issues or find better high-order rules for tetrahedra.

PR #661, opened 2021-06-27
Update some uses of deprecated InteriorBasis

kinnala:


gdmcbain (approved):
Is there a rationale for alternating betwen `CellBasis` and its alias `Basis` in `docs/examples`?  I wonder whether this might be confusing for newcomers?  (e.g., might they ask: Why do I see `Basis` in ex05 and `CellBasis` in ex06?  Is there some distinction implied that I don't understand yet?)

Anyway, following the definitively if not quite unanimously resolved discussion of what to rename the old `InteriorBasis` #606, I find I do like the simple alias `Basis`.  It seems quite natural in the framework of the finite element method.  If someone says &lsquo;basis&rsquo; without qualification, they almost certainly mean exactly this, whereas other bases (facets, various Lagrange multiplier spaces such as the global scalar in ex05) would naturally be qualified.  I don't know; naming things is hard.

kinnala:
Fair point. Let's use `Basis` only.

Issue #663, opened 2021-06-28
Mesh initializers don't accept lists


nschloe:
There is some code to support list inputs, but it doesn't work. MWE:
```python
import skfem


x = [0.0, 1.0]
c = [[]]
mesh = skfem.MeshLine(x, c)
```
```
  File "/path/to/skfem/mesh/mesh.py", line 1282, in __post_init__
    if len(self.doflocs.shape) == 1:
AttributeError: 'list' object has no attribute 'shape'
```

gdmcbain:
Yes, it would be nice from these to act more like NumPy's [array creation routines](https://numpy.org/devdocs/reference/routines.array-creation.html) and what's described in that documentation as `array_like`.  The [gloss](https://numpy.org/devdocs/glossary.html?highlight=array_like#term-array_like) of that isn't technical, but it seems to work well and many users will be used to it.  Now, how does `array_like` work, I haven't looked into it yet&hellip;  I suppose there's already something similar in initializing `meshio.Mesh` that should do the trick.

gdmcbain:
Oh, wow, [`numpy.empty_like`](https://github.com/numpy/numpy/blob/923996a75a75fc5789a7495fcb14e4970e675bc6/numpy/core/multiarray.pyi#L134-L136), for example, uses [`typing.overload`](https://docs.python.org/3/library/typing.html?highlight=typing#typing.overload)

```python
@overload
def empty_like(
    prototype: _ArrayType,
```
I don't know whether we want to go to that length.

gdmcbain:
What should be the interpretation of 
```python
c = [[]]
```
in the MWE?  I'm not sure that meshes with points not included in any cell are valid.

But here using 
```python
mesh = skfem.MeshLine(*map(np.asarray, [x, c]))
```
we get a single line-segment cell from 0 to 1.   That seems odd but I assume is a separate issue.

gdmcbain:
Anyway, [`meshio.Mesh`](https://github.com/nschloe/meshio/blob/6dd053a0d5b0f32a7a7d52a2ba47e951e026471c/src/meshio/_mesh.py#L31) uses [`numpy.asarray`](https://numpy.org/devdocs/reference/generated/numpy.asarray.html#numpy.asarray) for its `points`:
```python
        self.points = np.asarray(points)
```
which has been working pretty well for a while; let's go with that.

gdmcbain:
`MeshLine1.__post_init__` overrides the one inherited from `Mesh`.

[...]

PR #664, opened 2021-06-30
super post-init for MeshLine1

gdmcbain:
For #663.

### DIFF
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -349,13 +349,8 @@ class Mesh:
         if self.sort_t:
             self.t = np.sort(self.t, axis=0)
 
-        if not isinstance(self.doflocs, ndarray):
-            # for backwards compatibility: support standard lists
-            self.doflocs = np.array(self.doflocs, dtype=np.float64)
-
-        if not isinstance(self.t, ndarray):
-            # for backwards compatibility: support standard lists
-            self.t = np.array(self.t, dtype=np.int64)
+        self.doflocs = np.asarray(self.doflocs, dtype=np.float64)
+        self.t = np.asarray(self.t, dtype=np.int64)
 
         M = self.elem.refdom.nnodes
 
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -349,7 +349,7 @@ class Mesh:
         if self.sort_t:
             self.t = np.sort(self.t, axis=0)
 
-        self.doflocs = np.asarray(self.doflocs, dtype=np.float64)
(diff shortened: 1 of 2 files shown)


skfem/mesh/mesh.py
```diff
@@ -349,13 +349,8 @@ def __post_init__(self):
...
 
-        if not isinstance(self.doflocs, ndarray):
-            # for backwards compatibility: support standard lists
-            self.doflocs = np.array(self.doflocs, dtype=np.float64)
-
-        if not isinstance(self.t, ndarray):
-            # for backwards compatibility: support standard lists
-            self.t = np.array(self.t, dtype=np.int64)
+        self.doflocs = np.asarray(self.doflocs, dtype=np.float64)
+        self.t = np.asarray(self.t, dtype=np.int64)
```
kinnala, line 353: Will this always convert the memory layout to `C_CONTIGUOUS`? Thus, making the checks below redundant?

kinnala: I think there is a keyword argument `order='K'` to keep the memory layout of the original array if this was not intended.

gdmcbain: Hmm, sorry, I have no idea. I've many times wondered what the two warnings I often get about C-contiguity are but not yet gotten around to looking into what they mean or how to avoid them. So, no, I didn't intend to change any behaviour here. Sorry. What's the right thing to do here?

kinnala: In scikit-fem we do lots of slices like `mesh.p[0, :]` which is more cache-efficient if those numbers are next to each other in memory. So the conversion basically reorders those numbers so that `mesh.p[0, 0]` is just before `mesh.p[0, 1]` which is just before `mesh.p[0, 2]` and so on. However, because this has nonzero computational cost, I wanted to warn the user: sometimes this can be fixed outside scikit-fem.

kinnala: Now this PR does the conversion silently, without the warning. I'd like to preserve the warning so if you add `order='K'` no reordering will be done by `np.asarray`.

kinnala: Btw, I'm not sure if cache-efficiency is the correct term, but what I'm trying to say is that these CPU's will assume a specific access order and preload certain numbers to L1 cache.

gdmcbain: Right. That makes a lot of sense and sounds important.  I'll make the suggested change when I get to a terminal.

Could you elaborate on

> sometimes this can be fixed outside scikit-fem.

?

I have seen and ignored a lot of these warnings; have I been passing low-hanging fruit by?

PR #667, opened 2021-07-13
Improvements to Mesh.element_finder

### LINKED ISSUE #646: element_finder and external points
The behaviour of `MeshLine.element_finder` is anomalous when some of the points passed to the function returned are outside.  For example:

```python
MeshLine().element_finder()(np.array([[-0.3, 0.3, 0.9]]))
```
returns `array([[0, 0]])`.  The first point is external and silently dropped.  The two returned elements are those containing the remaining two points, but how is the caller to know?

```python
MeshLine().element_finder()(np.array([[0.3, 0.9, 1.5]]))
```
raises `IndexError: index 2 is out of bounds for axis 0 with size 2`.

More examples:

```python
MeshTri.init_circle().element_finder()(*[np.array([2])]*2)
```
returns `array([36])` while
```python
MeshTri.init_circle().element_finder()(*[np.array([-2])]*2)
```
returns `array([39])` but
```python
 MeshTri().element_finder()(*[np.array([2])]*2)
```
raises `IndexError: index 2 is out of bounds for axis 2 with size 2`.  Actually `MeshTri().element_finder()` seems to raise this for any point, even for one that should be internal.
```python
 MeshTri().element_finder()(*[np.array([0.5])]*2)
```

This arose while looking into the method of characteristics #600, for the feet of characteristics through points near an inlet.

What's a sensible thing to do here?  Return a sentinel value like &minus;1 for points for which a containing element can't be found?  Raise a specific exception?  Return the closest element?  (Is that what's happening for `MeshTri.init_circle()`?)

### LINKED ISSUE #666: Examples providing your own mesh and retriving the basis functions?
Are there any examples for supplying your own mesh (for example from scipy Delaunay) and extracting basis functions of some degree? 

For example you might evaluate the basis functions at a set of points for fitting against data as a kind of smoothing spline.

kinnala:
There is a heuristic optimization in `MeshTri1.element_finder` and `MeshTet1.element_finder` which tries to reduce the number of candidate elements where the query point may reside. The optimization logic looks for the query points inside elements whose midpoints are nearby. However, if the number of input points is large (in comparison to the number of elements), this "optimization" actually leads to a larger number of candidate elements and more memory being allocated.

- As in #668, take only unique candidate elements into consideration. Fixes #666.

Moreover, it was pointed out in #666 that the optimization logic does not work for some meshes with degenerate elements.

- Add a robustness test for randomized Delaunay `MeshTri1`and `MeshTet1`; compare against `scipy.spatial.Delaunay.find_simplex`
- Search all elements if the correct candidate is not found on the first try
- Increase the maximum number of candidate elements in `MeshTet1.element_finder`

While testing the new implementation it was found out that the optimization logic fails entirely for meshes that have less than 5 or 8 elements and raises error.

- Fix `MeshTri1.element_finder` and `MeshTet1.element_finder` for meshes that have a small number of elements.

Moreover, as pointed out in #646, points outside of the mesh are not detected correctly.

- Raise error if the point is outside of the mesh. Fixes #646.

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,8 +257,9 @@ with respect to documented and/or tested features.
 - Added: `ElementTriCCR` and `ElementTetCCR`, conforming Crouzeix-Raviart finite elements
 - Fixed: `Mesh.mirrored` returned a wrong mesh when a point other than the origin was used
 - Fixed: `MeshLine` constructor accepted only NumPy arrays and not plain Python lists
-- Fixed: `Mesh.element_finder` (and `CellBasis.probes` and `CellBasis.interpolator`) was not working properly for a small number of elements (<5) or a large number of input points (>1000)
-- Fixed: `MeshTet.element_finder` is now slightly more robust against degenerate tetrahedra
+- Fixed: `Mesh.element_finder` (and `CellBasis.probes`, `CellBasis.interpolator`) was not working properly for a small number of elements (<5) or a large number of input points (>1000)
+- Fixed: `MeshTet` and `MeshTri.element_finder` is are now more robust against degenerate elements
+- Fixed: `Mesh.element_finder` (and `CellBasis.probes`, `CellBasis.interpolator`) raises exception if the query point is outside of the domain
 
 ### [3.1.0] - 2021-06-18
 
--- a/README.md
+++ b/README.md
@@ -257,6 +257,7 @@ with respect to documented and/or tested features.
 - Added: `ElementTriCCR` and `ElementTetCCR`, conforming Crouzeix-Raviart finite elements
 - Fixed: `Mesh.mirrored` returned a wrong mesh when a point other than the origin was used
 - Fixed: `MeshLine` constructor accepted only NumPy arrays and not plain Python lists
+- Fixed: `Mesh.element_finder` (and `CellBasis.probes` and `CellBasis.interpolator`) was not working properly for a small number of elements (<5) or a large number of input points (>1000).
 
 ### [3.1.0] - 2021-06-18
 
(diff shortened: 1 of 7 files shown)


kinnala:
Already found out several cases where this is not working at all. Clearly requires more testing in addition to the optimizations.

skfem/mesh/mesh_tet_1.py
```diff
@@ -34,18 +35,36 @@ def element_finder(self, mapping=None):
...
-            )
-            return np.array([ix[np.argmax(inside, axis=0)]]).flatten()
+            inside = ((X[0] >= 0) *
+                      (X[1] >= 0) *
+                      (X[2] >= 0) *
+                      (1 - X[0] - X[1] - X[2] >= 0))
+            elems = np.argmax(inside, axis=0)
+
+            if ix is not None and not inside[elems].any(axis=0).all():
+                warnings.warn("Falling back to exhaustive search because the "
```
kinnala, line 62: I verified that this is activated only in `test_finder_simplex` where some of the elements have a nasty shape.

docs/examples/ex28.py
```diff
@@ -111,11 +103,7 @@ def make_mesh(halfheight: float,  # mm
               length: float,
               thickness: float) -> MeshTri:
     with geometrycontext as g:
-        if version.parse(pygmsh.__version__) < version.parse('7.0.0'):
-            geom = pygmsh.built_in.Geometry()
-            geom.add_curve_loop = geom.add_line_loop
-        else:
-            geom = g
+        geom = g
```
kinnala, line 106: Remove.

README.md
```diff
@@ -257,6 +257,8 @@ with respect to documented and/or tested features.
 - Added: `ElementTriCCR` and `ElementTetCCR`, conforming Crouzeix-Raviart finite elements
 - Fixed: `Mesh.mirrored` returned a wrong mesh when a point other than the origin was used
 - Fixed: `MeshLine` constructor accepted only NumPy arrays and not plain Python lists
+- Fixed: `Mesh.element_finder` (and `CellBasis.probes` and `CellBasis.interpolator`) was not working properly for a small number of elements (<5) or a large number of input points (>1000)
+- Fixed: `Mesh.element_finder` is now more robust against degenerate triangles and tetrahedra
```
kinnala, line 261: Clarify.

PR #668, opened 2021-07-18
In mesh tri 1, remove duplicate indices but keep order.

cottrell:
Putting this here to try to run the tests again. I think they were failing for unrelated reasons. 

Might simply be absorbable into #667 rather than clash with it.


kinnala:
Part of #667

PR #673, opened 2021-07-26
DOC: correct analytical eigenmodes

gdmcbain:
Looking at ex19 &lsquo;Heat equation&rsquo; again, in preparation for examples on the method of characteristics #600, ex19 being the sole unsteady example to date, I noticed that the analytical formula for the eigenmodes was incorrect.


gdmcbain:
The test-failures are related to pygmsh 7 in ex28, which is puzzling; I haven't used pygmsh 7 myself.  I don't think it's related to the present doc-fix.

gdmcbain:
That looks like #669.

gdmcbain:
Which is included in #667.  This can wait till that's merged; this isn't urgent.

kinnala:
Can you merge master so we see that the tests run?

PR #674, opened 2021-07-26
an unsteady one-dimensional example

gdmcbain:
Is it worth including an unsteady one-dimensional example?  If not, this can be closed and can just be referred to here.

Is ex19 (&lsquol;Heat equation&rsquo;) the only unsteady example so far?  This is a reduction of it to one dimension.  The finite element and analytical aspects are straightforward, but the plotting is different; one dimension is more different to two and three than the latter are to each other&mdash;we don't normally colour along a line for temperature in one dimension.

### DIFF
--- a/docs/examples/ex39.py
+++ b/docs/examples/ex39.py
@@ -70,22 +70,20 @@ if __name__ == "__main__":
     from matplotlib.animation import FuncAnimation
     import matplotlib.pyplot as plt
 
-    from skfem.visuals.matplotlib import plot
-
     parser = ArgumentParser(description="heat equation in a slab")
     parser.add_argument(
         "-g", "--gif", action="store_true", help="write animated GIF",
     )
     args = parser.parse_args()
 
-    ax = plot(mesh, u_init[basis.nodal_dofs.flatten()])
+    sorting = np.argsort(basis.doflocs)[0]
+    fig, ax = plt.subplots()
     ax.set_xlabel(r"$x/w$")
     ax.set_ylabel(r"reduced temperature, $u(x, t)/u(0, 0)$")
     ax.set_xlim(-halfwidth, halfwidth)
     ax.set_ylim(0, 1)
     title = ax.set_title("t = 0.00")
-    field = ax.get_children()[0]  # vertex-based temperature-colour
-    fig = ax.get_figure()
+    line = ax.plot(basis.doflocs[0, sorting], u_init[sorting], marker="o")[0]
 
     probe = basis.probes(np.zeros((mesh.dim(), 1)))
 
@@ -103,11 +101,7 @@ if __name__ == "__main__":
         )
 
         title.set_text(f"$t$ = {t:.2f}")
-
-        # skfem.visuals.matplotlib.plot_meshline plots a segment for
-        # each cell, separating them with a "None".
-
-        field.set_ydata(mesh.segment_data(u))
+        line.set_ydata(u[sorting])
 
     animation = FuncAnimation(fig, update, evolve(0.0, u_init), repeat=False)
     if args.gif:
--- a/docs/examples/ex39.py
+++ b/docs/examples/ex39.py
@@ -22,7 +22,6 @@ discretizing time using the generalized ('theta method') trapezoidal
 rule.
 
 """
-from skfem.mesh import MeshLine
 from typing import Iterator, Tuple
 
 import numpy as np
@@ -38,7 +37,7 @@ diffusivity = 5.0
 
 mesh = MeshLine(np.linspace(-1, 1, 2 * ncells) * halfwidth)
 
-element = ElementLineP1()
+element = ElementLineP2()  # or ElementLineP1
 basis = Basis(mesh, element)
 
 L = diffusivity * asm(laplace, basis)
@@ -53,7 +52,7 @@ B = M0 - (1 - theta) * L0 * dt
 
 backsolve = splu(A.T).solve  # .T as splu prefers CSC
 
-u_init = np.cos(np.pi * mesh.p[0, :] / 2 / halfwidth)
+u_init = np.cos(np.pi * basis.doflocs[0] / 2 / halfwidth)
 
 
 def evolve(t: float, u: np.ndarray) -> Iterator[Tuple[float, np.ndarray]]:
@@ -79,7 +78,7 @@ if __name__ == "__main__":
     )
     args = parser.parse_args()
 
-    ax = plot(mesh, u_init)
+    ax = plot(mesh, u_init[basis.nodal_dofs.flatten()])
     ax.set_xlabel(r"$x/w$")
     ax.set_ylabel(r"reduced temperature, $u(x, t)/u(0, 0)$")
     ax.set_xlim(-halfwidth, halfwidth)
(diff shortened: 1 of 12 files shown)


gdmcbain:
The main trick is that the plotting of the snapshot is changed from

https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/docs/examples/ex19.py#L117

to

https://github.com/kinnala/scikit-fem/blob/5b788e79ff5280b56a0b787e07565e604b03fd3b/docs/examples/ex39.py#L108-L113

This relies on the internal details of

https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/skfem/visuals/matplotlib.py#L89-L101

which it might be clearer to abstract, but I haven't attempted that here.  (I have though replaced the for-loop with a NumPy equivalent in 66144ed.)

skfem/visuals/matplotlib.py
```diff
@@ -137,22 +137,14 @@ def plot_meshline(m: MeshLine, z: ndarray, **kwargs):
...
-        xs.append(None)
-        ys.append(y1)
-        ys.append(y2)
-        ys.append(None)
-
-    ax.plot(xs, ys, color)
+    ax.plot(
+        *[
+            np.vstack([a[m.t], np.full(m.t.shape[1:], None)]).flatten("F")
+            for a in [m.p[0], z]
```
gdmcbain, line 144: Need to check generality of this. The previous uses only first two rows of `m.t` so should it be `m.t[:2]`? Need to check whether the first two points are the termini.

gdmcbain: Or is this only for two-point line-elements?

gdmcbain: So it's invoked for `MeshLine`

https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/skfem/visuals/matplotlib.py#L130-L131

which is [`MeshLine1`](https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/skfem/mesh/mesh_line_1.py#L13)

https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/skfem/mesh/__init__.py#L38

and that does only cover two-point line-segments defined by their termini.

However, it remains to consider the array of ordinates `z`, which might be defined on a basis other than `ElementLineP1`.

gdmcbain: Right, no, it fails for
```python
element = ElementLineP2()
```
raising `ValueError: dimension mismatch`.

gdmcbain: &hellip;but that was only because there was a bunch of other stuff that was only going to work with P1; e.g. 

https://github.com/kinnala/scikit-fem/blob/5b788e79ff5280b56a0b787e07565e604b03fd3b/docs/examples/ex39.py#L56

https://github.com/kinnala/scikit-fem/blob/5b788e79ff5280b56a0b787e07565e604b03fd3b/docs/examples/ex39.py#L82

I've fixed all that now; the element can be changed from `ElementLineP1` or `ElementLineP2` without changing anything else.

skfem/visuals/matplotlib.py
```diff
@@ -137,22 +137,8 @@ def plot_meshline(m: MeshLine, z: ndarray, **kwargs):
     else:
         ax = kwargs["ax"]
 
-    xs = []
-    ys = []
-
     color = kwargs["color"] if "color" in kwargs else 'ko-'
-    for y1, y2, s, t in zip(z[m.t[0]],
```
kinnala, line 144: I don't know if this is needed though? Isn't ax.plot(m.p[0], z) enough? This could be just some copypasta from draw_mesh2d where appending None gives faster performance.

kinnala: Oh I see, this is because of the arbitrary order of m.p[0]. I think np.argsort(m.p[0]) is needed before plotting.

gdmcbain: Oh, when i read the old `for`-loop implementation, I thought it was deliberately drawing one line segment per cell, possibly so that this would work with 
* unsorted `.p` or
* discontinuous Galerkin?

kinnala:
Unsorted p must be supported but dG is done through plot_basis.

gdmcbain:
I see that we don't have `ElementLineDG` yet. #675

skfem/visuals/matplotlib.py
gdmcbain: I opened #676 to discuss whether `plot_meshline` needs revision.

gdmcbain: > I don't know if this is needed though? Isn't `ax.plot(m.p[0], z)` enough?

It is indeed, or even using `basis.doflocs` and plotting all the ordinates, which is richer if using `ElementLineP2`.  I've gone with that in bdc010b and reverted the changes to `plot(MeshLine1)`.

That might be revisited in #676, or perhaps the resolution to that will end of being again, &lsquo;isn't `ax.plot(m.p[0], z)` enough?&rsquo;

gdmcbain:
ex39

*Figure:&mdash;* Evolution of the fundamental cooling mode of a uniform slab with cold walls, rendered directly in matplotlib (rather than the wrapper in `skfem.visuals`)

gdmcbain:
I'm thinking of following this with (besides the method of characteristics #600 for pure advection or advection&ndash;diffusion) something on a second-order IVP #531, e.g. a wave equation or possibly transverse vibration of an Euler&ndash;Bernoulli beam (ex34, 2a297e03a188290d6d215033e50b33fb09e45170).  I don't think the second time derivative will pose any difficulties (it didn't in my in-house code years ago and scikit-fem is only better).  A partial motivation for this (besides these two being of intrinsic interest) is the idea of treating strong advection with a wave-like reformulation.

That's not being held up by the present PR but will probably use a similar idiom for time-stepping and animation.

*  Gresho, P. M., Sani, R. L. (1998). *Incompressible Flow and the Finite Element Method.  Volume One: Advection&ndash;Diffusion*. Wiley. &sect;2.7.7*f* &lsquo;A wave equation method&rsquo;
*  Glowinski, R. & Pan, T.-W. (2019). Two decades of wave-like equation for the numerical simulation of incompressible viscous flow: A review.  In B. N. Chetverushkin, W. Fitzgibbon, Y. Kuznetsov, P. Neittaanmäki, J. Periaux & O. Pironneau (eds.), *Contributions to Partial Differential Equations and Applications* (pp. 221--250). Springer. [doi: 10.1007/978-3-319-78325-3_13](https://doi.org/10.1007/978-3-319-78325-3_13)

docs/examples/ex39.py
```diff
@@ -0,0 +1,110 @@
...
+L0, M0 = penalize(L, M, D=basis.find_dofs())
+A = M0 + theta * L0 * dt
+B = M0 - (1 - theta) * L0 * dt
+
+backsolve = splu(A.T).solve  # .T as splu prefers CSC
+
+u_init = np.cos(np.pi * basis.doflocs[0] / 2 / halfwidth)
+
+
+def evolve(t: float, u: np.ndarray) -> Iterator[Tuple[float, np.ndarray]]:
```
kinnala, line 58: Can you think about out how this could be run in the test suite? In particular, `tests/test_examples.py` should ideally run all examples so that they won't accidentally break. Maybe add some call to `evolve` to `tests/test_examples.py`?

gdmcbain: I've launched #678 for this.

gdmcbain:
Subsumed by #679.

PR #679, opened 2021-08-03
test unsteady ex19

### LINKED ISSUE #678: test unsteady examples
@kinnala, in reviewing #674 [yesterday](https://github.com/kinnala/scikit-fem/pull/674#pullrequestreview-719862159):

> Can you figure out how this could be run in the test suite? In particular, `tests/test_examples.py` should ideally run all examples so that they won't so easily break in the future. Maybe add some call to `evolve` to `tests/test_examples.py`?

Yes, increasingly important as more initial value problems #531 are tackled.

Currently, we have

https://github.com/kinnala/scikit-fem/blob/590960123fd73a616bec6ddde75e242882846e3f/tests/test_examples.py#L147-L149

which will catch some things, but I think it'll be easy enough to test evolution, as suggested.

gdmcbain:
Fixes #678.

### DIFF
--- a/docs/examples/ex17.py
+++ b/docs/examples/ex17.py
@@ -67,21 +67,32 @@ f = joule_heating * asm(unit_load, core_basis)
 
 temperature = solve(L + H, f)
 
+T0 = {
+    "skfem": (basis.probes(np.zeros((2, 1))) @ temperature)[0],
+    "exact": (
+        joule_heating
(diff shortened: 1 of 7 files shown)


gdmcbain:
I thought it might be as well to plot the exact solution too in the one-dimensional example #674, seeing as that's easy.

ex39

kinnala (approved):

Issue #474, opened 2020-09-15
Examples/snippets demonstrating the integrations


kinnala:
I'd like to keep a track of the libraries that can be integrated to scikit-fem with ease. How about having a single example which simply demonstrates the possible integrations? Or should it be a collection of snippets instead? If the latter, how is it tested in pytest?

gdmcbain:
I think a collection; the integrations serve diverse ends.

For pytest, I've seen this idiom:

```python
netCDF4 = pytest.importorskip("netCDF4")
```

https://github.com/nschloe/meshio/blob/3a380fc1100827ed38b8f38960087a066c4aab26/test/test_exodus.py#L6

kinnala:
So basically in practice we could have have something like `docs/integrations/pygmsh.py`, `docs/integrations/dmsh.py`, and so on, make sure that these are collected by pytest, and then maybe have `docs/snippets.rst` which includes all these in the Sphinx documentation?

gdmcbain:
Yes, I think so. That directory structure should make it easy to find the case of interest.

Is there a catch with trying to `import pygmsh` from inside a file.called `pygmsh.py` or from another file in a directory with that file?

kinnala:
Yeah, let's call them `mesh_pygmsh.py`, `mesh_dmsh.py`, `solver_pacopy.py`, and so on, i.e. try to have a 'category' which describes the type of integration?

kinnala:
Here are some integrations that come to mind:
- [ ] pygmsh
- [ ] pacopy
- [ ] dmsh
- [ ] pyamgcl
- [ ] sksparse
- [ ] optimesh
- [ ] pyfvm (moving mesh to and from, e.g., for iterative coupling)
- [ ] vedo (actually I should continue on #361 instead)

I'm sure there are others. Feel free to add to the list.

kinnala:
Some additional urls for consideration:
- https://github.com/andrenarchy/krypy
- https://pymesh.readthedocs.io/en/latest/api_sparse_solver.html
- https://github.com/nschloe/pygalmesh
- https://github.com/nschloe/meshzoo
- https://github.com/pyvista/pyvista
- https://pypi.org/project/petsc4py/

gdmcbain:
There's a start on PETSc in #236 .

After that recent question about pressure boundary conditions for the Stokes equation #466 , which referred to a FEniCS example using MUMPS, I started looking at [PyMumps](https://github.com/pymumps/pymumps).

[...]

PR #680, opened 2021-08-03
Improvements to skfem.io.meshio

### LINKED ISSUE #158: serialization of meshes and solutions

### LINKED ISSUE #261: Mesh.save subdomains and boundaries

### LINKED ISSUE #271: load auxiliary data with meshes

### LINKED ISSUE #683: internal boundaries in with_boundaries
Internal boundaries were queried in #336 and support added in #340.  I wanted to check that these were preserved in serialization #158 #261 under #680, but it appeared whereas they were supported by the old [`Mesh.define_boundary`](https://github.com/kinnala/scikit-fem/blob/0855dae117728588f269d968c8da4dbe06446d1b/skfem/mesh/mesh.py#L142-L144), this has been deprecated in favour of the (functional rather than mutating) [`Mesh.with_boundaries`](https://github.com/kinnala/scikit-fem/blob/0855dae117728588f269d968c8da4dbe06446d1b/skfem/mesh/mesh.py#L164-L165) which doesn't support them.

The problem is in 

https://github.com/kinnala/scikit-fem/blob/0855dae117728588f269d968c8da4dbe06446d1b/skfem/mesh/mesh.py#L180

which hardcodes `boundaries_only=True`.

Minimal example:
```python
m = MeshLine([0, 1, 2]).with_boundaries({"mid0": lambda x: x[0] == 1.0})
m.define_boundary("mid1", lambda x: x[0] == 1.0, boundaries_only=False)
```
Then `m.boundaries` is
> {'mid0': array([], dtype=int64), 'mid1': array([1])}

kinnala:
- Write/read boundaries/subdomains to/from external file formats
- Optionally, read mesh data from external formats

Fixes #158 
Fixes #261
Fixes #271 
Fixes #683 

Example 1: Export boundaries to external formats.
```python
m = MeshHex().with_boundaries({'left': lambda x: x[0] == 0, 'bottom': lambda x: x[2] == 0})
m.save('test.msh', file_format='gmsh22')
```
Kuvakaappaus - 2021-08-09 11-53-08

Example 2: Load the exported boundaries
```python
In [1]: from skfem import *                          
In [2]: m = MeshHex().with_boundaries({'left': lambda x: x[0] == 0, 'bottom': lambda x: x[2] == 0})       
In [3]: m.save('test.vtk')                           
In [4]: M = MeshHex.load('test.vtk')                 
In [5]: M.boundaries                                 
Out[5]: {'left': array([0]), 'bottom': array([2])}
In [6]: m.boundaries                                 
Out[6]: {'left': array([0]), 'bottom': array([2])}
```

Example 3: Load raw data from meshio
```python
In [1]: from skfem import *                          
In [2]: out = ['cell_data']                          
In [3]: m = Mesh.load('docs/examples/meshes/beams.msh', out=out)                                         
In [4]: out[0].keys()                               
Out[4]: dict_keys(['gmsh:physical', 'gmsh:geometrical'])
```


gdmcbain:
Good morning.  Serialization #158 is back on the agenda, excellent: I believe this is an important long-term goal.  Would this fix
#261 &lsquo;`Mesh.save` boundaries and subdomains&rsquo;  ?

There's quite a history of efforts on that here and even more in meshio.  I'll have to refresh myself on that to see what the status is.  Prior to that, my impressions are that:
* `meshio.Mesh.cell_sets` and equivalents (!) have not been implemented consistently with respect to different external file formats
* it is not possible to write a nontrivial mesh, i.e. with subdomains or boundaries, in MSH 4.1 using meshio

I do however routinely save meshes with subdomains and boundaries.  I can outline my current practice, in case that's helpful, or share code&mdash;basically I write MSH 2.2 using meshio using `cell_data` rather than `cell_sets`.  Are you trying to solve a specific problem?  Or rather to generally extend the capabilities of `skfem.io`?  I can also share my current thoughts on strategy for #261, if the current PR doesn't cover it.

While there may not be documentation for `meshio.Mesh.cell_sets`, &lsquo;cell_sets&rsquo; is mentioned in [29 issues](https://github.com/nschloe/meshio/search?q=cell_sets&type=issues) !  A lot of discussion !

gdmcbain:
> it is not possible to write a nontrivial mesh, i.e. with subdomains or boundaries, in MSH 4.1 using meshio

What I mean by this is that if just before

https://github.com/kinnala/scikit-fem/blob/80e39e282b07d5924918ee1cf81ac429025151e0/docs/examples/ex28.py#L145

I put
```python
        mesh = geom.generate_mesh(dim=2)
        mesh.write(Path(__file__).with_suffix(".msh"), binary=False)
```

and then in the script
```python
make_mesh(length, halfheight, thickness)
```
I get 
```
  File "/home/gdmcbain/src/scikit-fem/docs/examples/ex28_make_mesh.py", line 149, in <module>
    make_mesh(length, halfheight, thickness)
  File "/home/gdmcbain/src/scikit-fem/docs/examples/ex28_make_mesh.py", line 146, in make_mesh
    mesh.write(Path(__file__).with_suffix(".msh"), binary=False)
  File "/home/gdmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/meshio/_mesh.py", line 219, in write
    write(path_or_buf, self, file_format, **kwargs)
  File "/home/gdmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/meshio/_helpers.py", line 141, in write
    return writer(filename, mesh, **kwargs)
  File "/home/gdmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/meshio/gmsh/main.py", line 111, in <lambda>
    "gmsh": lambda f, m, **kwargs: write(f, m, "4.1", **kwargs),
  File "/home/gdmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/meshio/gmsh/main.py", line 102, in write
    writer.write(filename, mesh, binary=binary, float_fmt=float_fmt)
  File "/home/gdmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/meshio/gmsh/_gmsh41.py", line 358, in write
    _write_nodes(fh, mesh.points, mesh.cells, mesh.point_data, float_fmt, binary)
  File "/home/gdmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/meshio/gmsh/_gmsh41.py", line 607, in _write_nodes
    raise WriteError(
meshio._exceptions.WriteError: Specify entity information (gmsh:dim_tags in point_data) to deal with more than one cell type. 
```
I'm pretty sure that this is a meshio issue.

However, if I insert
```python
file_format="gmsh22", 
```
into the call to `meshio.write`, all goes well : [ex28_make_mesh.msh.txt](https://github.com/kinnala/scikit-fem/files/6928407/ex28_make_mesh.msh.txt)

kinnala:
So I'm planning to write `meshio.Mesh.cell_sets` inside `skfem.io.meshio.to_meshio` in order to call `meshio.Mesh.sets_to_int_data` and then use the resulting `meshio.Mesh.cell_data` for further purposes. But I think I'm not properly understanding the correct format of `cell_sets`.

kinnala:
I might have now got it correctly. Pushing some more commits soon.

gdmcbain:
O. K.

The best way to see what `meshio.Mesh.cell_sets` should look like is to `meshio.read` a genuine MSH 4.1 file (genuine as in generated by Gmsh itself, e.g. from GEO code); meshio reads MSH 4.1 O. K., it's the writing that's tricky. I'm not sure it's even conceptually possible to write 4.1 for an arbitrary mesh that is acceptable to Gmsh.

tests/test_mesh.py
```diff
@@ -290,26 +291,98 @@ def test_meshio_cycle(m):
...
+def test_saveload_cycle_tags(m):
+
+    m = (m
+         .refined()
+         .with_subdomains(_test_lambda)
+         .with_boundaries({'test': lambda x: x[0] == 0}))
+    from tempfile import NamedTemporaryFile
+    with NamedTemporaryFile() as f:
+        m.save(f.name + ".inp", write_boundaries=True)
+        m2 = Mesh.load(f.name + ".inp")
```
kinnala, line 378: This is funny that Abaqus inp format works properly for write/read cycle but almost nothing else does without sorcery.

kinnala:
This allows now doing stuff like this:
```python
In [1]: from skfem import *                          

In [2]: m = MeshTri().with_boundaries({'left': lambda x: x[0] == 0})                                      

In [3]: m.save('testing.msh', write_boundaries=True, sets_to_int_data=True, file_format='gmsh22')         
WARNING:root:msh2 requires 3D points, but 2D points given. Appending 0 third component.
WARNING:root:Binary Gmsh needs 32-bit integers (got int64). Converting.
WARNING:root:Binary Gmsh needs 32-bit integers (got int64). Converting.
WARNING:root:Appending zeros to replace the missing physical tag data.
WARNING:root:Appending zeros to replace the missing geometrical tag data.

In [4]: M = MeshTri.load('testing.msh', int_data_to_sets=True)                                            

In [5]: M.boundaries                                 
Out[5]: 
{'set-1.0': array([0, 3, 4]),
 'set0.0': array([1]),
 'gmsh:physical': array([0, 1, 3, 4]),
 'gmsh:geometrical': array([0, 1, 3, 4])}

In [6]: m.boundaries                                 
Out[6]: {'left': array([1])}
```
As you can see, the tag names are lost due to the `sets_to_int_data`/`int_data_to_sets` cycle, but the correct set is there in the end.

gdmcbain:
Fantastic!

> As you can see, the tag names are lost due to the sets_to_int_data/int_data_to_sets cycle, but the correct set is there in the end.

I wonder whether the names can be stashed in `field_data` or similar. (Or in the names of the cell data?)

That's weird and unfortunate about Abaqus .inp. Is that a viable format? Can one do anything with it without access to the Abaqus program? It might just be that better work has been done on that submodule of meshio.

MSH 2.2 is not ideal, it's very slow. In my first industrial application of scikit-fem, reading MSH 2.2 was actually the bottleneck.

gdmcbain (commented):
Oh, no, I see that we've gone from having one named boundary
```python
{'left': array([1])}
```
to four, all misnamed, and three spurious
```python
{'set-1.0': array([0, 3, 4]), 'set0.0': array([1]), 'gmsh:physical': array([0, 1, 3, 4]), 'gmsh:geometrical': array([0, 1, 3, 4])}
```
Yeah, no, that's not so good.  Hmm&hellip;Some further thoughts in #261.

skfem/io/meshio.py
```diff
@@ -70,21 +95,17 @@ def from_meshio(m, force_mesh_type=None):
...
+        bnd_type = BOUNDARY_TYPE_MAPPING[meshio_type]
 
         def find_tagname(tag):
             for key in m.field_data:
                 if m.field_data[key][0] == tag:
                     return key
             return None
 
+        if int_data_to_sets:
+            m.int_data_to_sets()
```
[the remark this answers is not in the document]
kinnala: AFAIC losing of the names is a bug of meshio and should be fixed there. I think the logic in `int_data_to_sets` which tries to match the tag names from the combined `cell_data` tag (`left`, `right` gets transformed to `left-right` in `sets_to_int_data` and the latter is again parsed by `int_data_to_sets`) is not correct.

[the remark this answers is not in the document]
kinnala: Also, the additional tags with `gmsh` prefix come from the `gmsh22` reader. I think these readers can pretty much arbitrarily add additional stuff.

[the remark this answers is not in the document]
gdmcbain: I know, and mysterious new ones keep appearing, like "gmsh:dim_tags" in nschloe/meshio#922.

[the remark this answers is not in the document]
kinnala: Tagging is now fixed in nschloe/meshio main branch.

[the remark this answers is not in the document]
gdmcbain: Good one. Does that fix nschloe/meshio#736?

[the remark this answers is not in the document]
kinnala: I think partially. It's not reading field_data though. Maybe the docstring should read cell_data instead.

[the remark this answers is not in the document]
gdmcbain: The [docstring](https://github.com/nschloe/meshio/blob/b3630c2d8d7649b14d26665ca75679258b6fc483/src/meshio/_mesh.py#L342) seems to have stopped referring to `field_data` a long time ago (I failed to bisect this over the move to the `src` layout nschloe/meshio#1090, but no matter).

I'm not sure what you mean by

> Tagging is now fixed 

When I looked again at nschloe/meshio#736 this morning, I found that the keys to `cell_sets` after `int_data_to_sets` in the MWE there looked like `"set1", "set2", ...` which is different from yesterday, but still a little inscrutable.

One could use something like
```python
{k: f"set{v[0]}" for k, v in mio.field_data.items()}
```
to decode the otherwise opaque names, connecting them back to the ones meaningful in the application?  But I'm not quite clear on the kind of interoperation that is planned in the motivating application.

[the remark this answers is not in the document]
kinnala: It won't parse `gmsh22` tags from `field_data` because that's not the purpose of `int_data_to_sets`. But the issue here is not related because `gmsh22` writer neither does write the tags to `field_data`. See https://github.com/nschloe/meshio/issues/1160 https://github.com/nschloe/meshio/issues/1164 https://github.com/nschloe/meshio/issues/1165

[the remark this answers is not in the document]
kinnala: I think tags being in `field_data` is specific to `gmsh22` whereas `int_data_to_sets` tries to be generic w.r.t the format.

tests/test_mesh.py
```diff
@@ -290,26 +291,98 @@ def test_meshio_cycle(m):
...
+def test_saveload_cycle_tags(m):
+
+    m = (m
+         .refined()
+         .with_subdomains(_test_lambda)
+         .with_boundaries({'test': lambda x: x[0] == 0}))
+    from tempfile import NamedTemporaryFile
+    with NamedTemporaryFile() as f:
+        m.save(f.name + ".msh", write_boundaries=True, sets_to_int_data=True, file_format='gmsh22')
+        m2 = Mesh.load(f.name + ".msh", int_data_to_sets=True)
```
kinnala, line 378: What I mean by "tagging is fixed" that this now works properly using nschloe/meshio main branch.

skfem/io/meshio.py
```diff
@@ -118,8 +118,9 @@ def from_meshio(m,
                                    if f in v])
                       for k, v in facets.items()}
 
-    # fall back to MSH 2.1 tag parsing
-    if not m.cell_sets:
+
+    # MSH 2.1 tag parsing
```
gdmcbain, line 122: 2.2 or 4.1?

kinnala: Sorry, it's 2.2 I think.

kinnala:
MSH 4.1 will likely require some workaround, e.g., writing `gmsh:dim_tags`. I need to look into that further.

kinnala:
If I remove `WriteError` line from meshio, then everything works just fine. However, I cannot figure out what is this `gmsh:dim_tags`.

gdmcbain:
> MSH 4.1 will likely require some workaround, 

Personally i wouldn't see writing MSH 4.1 as a requirement. Is that needed for interoperation with Elmer?

It might (for my purposes #261) be worth working out a format that meshio writes and reads reasonably quickly. It does read MSH 4.1 much faster than MSH 2.2; this is an intrinsic benefit of the redesign from the Gmsh tesm, avoiding mixed topologies.

Unfortunately, the meshio speed tests don't test mixed topologies, and so while XDMF looks fast, that's only true if only one kind of cell is written. The current meshio XDMF writer, if given domain and boundary cells, will revert to a mixed topology which i fear will be slow to reread&mdash;each cell will have to be processed individually  to see whst kind it is. This is not intrinsic to XDMF but a shortcoming of meshio; a fix for that is proposed in nschloe/meshio#622.

This was part of the motivation for writing boundary facets as such in #681 rather than as lower-dimensional cells (besides the main motivation of that being a more direct representation). But neither of those are overriding considerations ; good interoperation is.

I'm thinking that VTK will probably be a reasonable default for using this PR to fix #261 ; I.e. write and reread by skfem.

If Elmer doesn't use meshio to read MSH 2.2 but rather has a C++ reader, it might not be slow, so that could be a workable procedure there.

kinnala:
I think I'll be able to fix MSH 4.1; it seemed to work just fine when I simply commented out the dim_tags exception from meshio.

One possible goal is to get Elmer support to meshio. For now I've been using something that I hacked together in kinnala/elmer-runner but I'd like it to not depend on scikit-fem (for distribution) while still being able to easily use meshes from scikit-fem (via meshio). However, I might also be able to use ElmerGrid importers because it already supports quite a few mesh formats.

kinnala:
Seems that MSH 4.1 write support in meshio is quite broken, nothing I can do about that here. I'll simply make `.msh` write using `gmsh22` always.

kinnala:
I have one more idea which encodes boundaries to point_data using bit encoding and then applies columnwise bit_and on the array point_data[m.facets] during decoding. This could be then used optionally by disabling the boundary cell writing via a flag. It should support pretty much any format in existence, right?

gdmcbain:
> Seems that MSH 4.1 write support in meshio is quite broken, nothing I can do about that here. 

I agree. It never worked originally for more thsn the trivial subset of MSH 4.1 for which `$Entities` can be omitted. I didn't understand the last substantive change which introduced `"gmsh:dim_tags"`&c. Bug reports accumulate but I don't think anyone knows what to do about it. I gave up a long time ago, deciding to treat MSH 4.1 as read-only.

> I'll simply make .msh write using gmsh22 always.

Yes, I have `file_format="gmsh22"` everywhere in my code so may as well make it the default. meshio used to fall back on 2.2 when it admitted that it couldn't write real 4.1 but that was dropped with the dim-tags change. 

It might be worth one more bug report on meshio writing MSH 4.1 if you've got the patience to record your unhappy findings; the situation is poor but should be plainly and openly documented. We do remain reliant on meshio.

gdmcbain:
> I have one more idea which encodes boundaries to `point_data` using bit encoding and then applies columnwise bit_and on the array `point_data[m.facets]` during decoding. This could be then used optionally by disabling the boundary cell writing via a flag. It should support pretty much any format in existence, right?

Yes, that's right; having only one type of cell does simplify things. It also makes meshio writing & reading XDMF much faster, pending 622. 

FEniCS currently uses an awful system of two XDMF files, one for domain cells, one for (*n* – 1)-dimensional cells representing boundary-facets. Each with a copy of the points.

There are some mesh formats in meshio which don't seem to write point or cell data—some might not support it (StL?), others I presume remain incompletely implemented (Exodus?)—but by and large yes, a homogeneous mesh with point and cell data (rather thsn explicit subsets) is a safe bet, very broadly supported and portable. 

Is the advantage of encoding boundaries in `point_data` rather than recording the facets as (domain) cell data as in #681 that ParaView &c. then do a better job of rendering the boundaries?

kinnala:
I think I don't fully understand how you distinguish different facets of the cells in #681. I need to look into it more carefully. I was thinking more like encoding all boundaries into a single point_data entry and all subdomains into a single cell_data entry. This way there is less data to write/read.

kinnala:
Now should work with any format supporting point_data and cell_data.

skfem/mesh/mesh.py
```diff
@@ -84,6 +84,68 @@ def subdomains(self):
     def boundaries(self):
         return self._boundaries
 
+    def encode_point_data(self):
```
kinnala, line 87: Here is the relevant part.

gdmcbain: Will the point-data work for &lsquo;boundaries&rsquo; which aren't on the boundary? Currently the facet-based approach does support this (as does Gmsh in exporting (*n* –1)-dimensional cells representing facets).

gdmcbain:
> I was thinking more like encoding all boundaries into a single point_data entry and all subdomains into a single cell_data entry. This way there is less data to write/read.

O. K., right, that makes sense. In #681 there's one cell data for each subdomain or boundary, it being thought that multiple arrays wouldn't really be very expensive. The original idea was to use strict indicator functions, i.e. with values in {0, 1} or `bool`, as a very direct representation of the subsets that are stored on the `skfem.Mesh`attributes.

https://github.com/gdmcbain/scikit-fem/blob/4c35c4f927d448cff7b321c7db12da17fba48a0b/docs/examples/ex_261.py#L12-L17

```python 
    subdomains = {
        f"skfem:subdomain:{name}": [
            np.isin(np.arange(mesh.t.shape[1]), subdomain).astype(int)
        ]
        for name, subdomain in mesh.subdomains.items()
    }
```

> I think I don't fully understand how you distinguish different facets of the cells in #681. 

The above works for subdomains but for boundaries we need to specify which facet of the cell, so it needs to be uint rather than bool.  Then I realized some boundaries have more than one facet of a cell, hence the bit encoding.

There wasn't any attempt to encode multiple subdomains or boundaries into single arrays but that's probably a straightforward extension, if required or desired.  I suppose there's some generalization of the bit-encoding that would work in composition with the bit-encoded uint-valued boundary facets? But I won't pursue it as the present approach seems fine.

Is the idea that some other FE packages (Elmer?) will understand the point-data? I think there might be codes that do accept specification of boundaries like that.

skfem/mesh/mesh.py
gdmcbain: Old issue on internal boundaries: #336

gdmcbain: A preliminary investigation (based on the MWE from #683)
```python
m = MeshLine([0, 1, 2])
m.define_boundary("mid1", lambda x: x[0] == 1.0, boundaries_only=False)
print("Before:", m.boundaries)
print("After:", from_meshio(to_meshio(m)).boundaries)

```
suggests no: the internal boundary `"mid1"`of `m` is emptied during the round-trip `to_meshio`, `from_meshio`:
> Before: {'mid1': array([1])}
> After: {'mid1': array([], dtype=int64)}

gdmcbain: Of course that is an improvement on the master branch, in which `to_meshio` doesn't attempt to serialize the boundaries at all.

> Before: {'mid1': array([1])}
> /home/gdmcbain/src/scikit-fem/skfem/io/meshio.py:134: UserWarning: Unable to load tagged boundaries/subdomains.
  warnings.warn("Unable to load tagged boundaries/subdomains.")
> 'gmsh:physical'
> After: None

gdmcbain: I think the culprint is `_decode_point_data`:

https://github.com/kinnala/scikit-fem/blob/d0febca6118745fd41c4bb04679efec40fb3f167/skfem/mesh/mesh.py#L109-L110

kinnala: The approach in this PR cannot support "interfaces". I also think we should not call them "boundaries" anymore. However, Mesh.interfaces will require a different approach. I propose that we do that in a different issue/PR, I'm running out of time on this one.

kinnala: I mean, suppose you have a tetrahedral element with all vertices on the boundary. If `point_data` contains the value 1 on all of its vertices, then is the interior facet part of the named boundary or not? In order to serialize internal facets, we should write to `cell_data`, some matching identified to both elements sharing the facet.

kinnala:
> Is the idea that some other FE packages (Elmer?) will understand the point-data? I think there might be codes that do accept specification of boundaries like that.

I'm not sure if that's the case. I think Elmer requires explicit boundary facets. However, I think we shouldn't write those out just for the serialization purpose since the mesh format support will be reduced. There could be an additional flag `write_facets` though.

kinnala:
> The above works for subdomains but for boundaries we need to specify which facet of the cell, so it needs to be uint rather than bool. Then I realized some boundaries have more than one facet of a cell, hence the bit encoding.

Does it rely on the mesh import/export not perturbing the order of the nodes, i.e., the facets in `t2f` will always be in the same order? I ran into some issues regarding this and `MeshHex`, probably because all the tools have a different idea of what is the correct node ordering.

skfem/mesh/mesh.py
gdmcbain: >  I propose that we do that in a different issue/PR, I'm running out of time on this one.

Sure, sorry.  Press ahead and read this later if at all.

> If point_data contains the value 1 on all of its vertices, then is the interior facet part of the named boundary or not?

It could be that this indicates a limitation of the idea of representing boundaries or interfaces by points when really they're facets of domain-cells.

> In order to serialize internal facets, we should write to cell_data a matching identifier for both elements sharing the interior facet.

Actually the approach taken in #681 seems to cope O. K. with "interfaces".  It doesn't distinguish between interfaces and boundaries; the `.boundaries` attribute of `skfem.Mesh` is just an array of indices of `.facets` which belong to the named subset, as previously.  Because it works directly with the facets of the internal representation, this just works.

> The approach in this PR cannot support "interfaces". I also think we should not call them "boundaries" anymore. 

Are they indeed distinct?  Up till now it seems to have been O. K. to just treat them as named subsets of facets.

>  In order to serialize internal facets, we should write to cell_data a matching identifier for both elements sharing the interior facet.

That's right, that is what was done automatically in #681, because such a facet showed up as `True` in two columns of the `isin` array:  

https://github.com/kinnala/scikit-fem/blob/cfb38f7212eb1cde93b7686480b30f5921adfc63/docs/examples/ex_261.py#L18-L23

But you're running out of time and I'm quibbling, push ahead and yes this can be sorted out later if required.  I can use #681 in-house indefinitely.

gdmcbain:
Yes, it is completely reliant on `.p`, `.t` and the other derived connectivity arrays being identical and not permuted.  

I think `meshio` is quite careful not to permute anything though, isn't it?  This has been holding up a PR on handling OpenFOAM's polyMesh format nschloe/meshio#1105.  That seems to be difficult (although I've only been following casually) as polyMesh nowhere specifies the cell&ndash;point connectivity, only cell&ndash;facet and facet&ndash;point.

Isn't the point&ndash;cell ordering handled for us by consistently reordering e.g. `MeshHex` (I think that's the only one at the moment) to and from meshio?  And trusting meshio to do the same when it writes and reads any other format?

kinnala:
Yeah, it seems that meshio is indeed preserving `t` which should also imply that `t2f` is preserved. I wouldn't want to write a separate reader for this format though because it really relies on how `t2f` is constructed in `scikit-fem`. :-)

kinnala:
Alright, I stole your approach from #681 and fixed it to work with `MeshHex`.

gdmcbain:
> because it really relies on how t2f is constructed in scikit-fem. 

Yeah, it has been disappointing and frustrating not to find an established finite element mesh file format with a notion of facet. I keep hoping to turn one up but it's been a while now.

kinnala:
Now user can pass the optional parameter `out` to `Mesh.load`. If it is a dictionary, it gets populated with `cell_data`, `point_data` and `field_data`.

kinnala:
I noticed that in `ex28.py`, the lines depending on `pygmsh` are not being run at all. So I removed the unused lines and, thus, reverted the license to BSD-3-Clause. I suppose that's fine by you @gdmcbain?

kinnala:
Note that I'm still planning to do #474 at some point, which will hopefully include stuff like how to use `pygmsh`.

gdmcbain:
Yes, I noticed that too in ex28 and was puzzled. Thanks, it's better now.

gdmcbain:
I have just fetched this branch and taken it for a spin.  It's awesome.  I wish we had done this years ago!  So far everything just works, very easily, no surprises; very nice to use.

One slight disappointment is that the new `docs/examples/ex28.msh` is 156K whereas the old JSON was only 52K.  A bit of a shame since the latter is not compressed at all and was hardly designed for efficiency.  VTK is a bit better at 128K and parity is restored with XDMF at 48+4=52K, with the additional requirement of `h5py` (not difficult but it is an additional requirement).  I suppose all our simple boolean arrays of indicator functions for subdomains and uints for boundaries are getting stored as big floats (float64?).  I don't think this matters too much; in particular I don't think it is too much to pay for the ease of being able to immediately open the VTK or XDMF in ParaView and immediately see each subdomain and boundary, along with e.g. `point_data={"temperature": temperature[basis["heat"].nodal_dofs.flatten()]}`.  Conclusion: the indicator functions are verbose but portable, which is a fair compromise and probably the right one to make here.

The pinked visualization of the boundaries is a bit zany, but it kind of makes sense, I think; I wonder what others will think of it.

Is this ready for a review?  Anyway, I'm going to begin using it for everything immediately.

kinnala:
Yes, I think it's ready.

kinnala:
I updated the first message in the PR with some examples.

gdmcbain (approved):

PR #684, opened 2021-08-09
reallow internal boundaries

gdmcbain:
With new `with_boundaries` which replaces deprecated mutating `define_boundary`.


kinnala:
This is now part of #680 and the import/export is also tested there.

Issue #686, opened 2021-08-12
trace regression


gdmcbain:
MWE:
```python
mesh = MeshQuad().with_boundaries({"left": lambda x: x[0] == 0.0})
basis = CellBasis(mesh, ElementQuad1())
left = BoundaryFacetBasis(mesh, basis.elem, facets=mesh.boundaries["left"])
trace = left.trace(basis.doflocs[1], lambda x: x[1])
print(trace[1])
```

This works with `pip install scikit-fem<3.2`, printing `[0.5]`, but fails in `pip install 3.2` or current master (e7196f61b59c61706ef105de35b3cc0453fb8c72 ), raising an `IndexError`:
```

Traceback (most recent call last):
  File "C:\Users\gmcbain\mwe.py", line 6, in <module>
    trace = left.trace(basis.doflocs[1], lambda x: x[1])
  File "C:\Users\gmcbain\Miniconda3\envs\skfem\lib\site-packages\skfem\assembly\basis\boundary_facet_basis.py", line 179, in trace
    CellBasis(target_meshcls(projection(p), t), elemcls()),
  File "<string>", line 11, in __init__
  File "C:\Users\gmcbain\Miniconda3\envs\skfem\lib\site-packages\skfem\mesh\mesh_line_1.py", line 23, in __post_init__
    super().__post_init__()
  File "C:\Users\gmcbain\Miniconda3\envs\skfem\lib\site-packages\skfem\mesh\mesh.py", line 417, in __post_init__
    if self.doflocs.shape[1] > 1000:
IndexError: tuple index out of range
```

I think it's something to do with whether the points defining a `MeshLine1` have `ndim` 1, or 2 with `shape[0] == 1`; accepting the former is a special case still accepted by the main `MeshLine1` constructor.

The example does run (in 3.1, 3.2, and current) if the `projection` passed to `left.trace` is changed to `lambda x: x[1:]`, so a makeshift is to change old code.

gdmcbain:
Uh-oh, `git bisect` says:
```
26af4c44e05e61c2b948b7b33bb39a7abbf552b5 is the first bad commit
commit 26af4c44e05e61c2b948b7b33bb39a7abbf552b5
Author: G. D. McBain <gdmcbain@protonmail.com>
Date:   Wed Jun 30 10:48:57 2021 +1000

    super post-init; fixes #663

 skfem/mesh/mesh_line_1.py | 3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)
```

PR #687, opened 2021-08-12
atleast_2d before super-post-init

gdmcbain:
Coercion of &lsquo;scalar&rsquo; one-dimensional coordinates to 1-vectors needs to happen before `Mesh.__post_init__`; otherwise the trace of a two-dimensional basis with a &lsquo;scalar&rsquo; one-dimensional `projection` hits #686.

### DIFF
--- a/skfem/mesh/mesh_line_1.py
+++ b/skfem/mesh/mesh_line_1.py
@@ -20,11 +20,8 @@ class MeshLine1(Mesh):
 
     def __post_init__(self):
 
+        self.doflocs = np.atleast_2d(self.doflocs)
         super().__post_init__()
-
-        if self.doflocs.ndim == 1:
-            # support flat arrays
-            self.doflocs = np.array([self.doflocs])
 
         if self.t.shape[1] != self.doflocs.shape[1] - 1:
             # fill self.t assuming ascending self.doflocs if not provided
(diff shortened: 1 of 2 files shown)


kinnala:
Why no test caught this?

gdmcbain:
Indeed. Let me fix that.

gdmcbain:
The only test always passes an (*n* &minus; 1)-dimensional projection.

https://github.com/kinnala/scikit-fem/blob/e7196f61b59c61706ef105de35b3cc0453fb8c72/tests/test_basis.py#L250

(It did cross my mind yesterday that maybe the points passed to construct a one-dimensional mesh should have `ndim==2` and `shape[0]==1` but that is a bit too strict, isn't it.  People do want to be able to pass just `ndim==1` arrays of just the _x_-coordinate, or even lists #663, and it would seem to be pythonic to accept them.)

gdmcbain:
This works (fails on master, passes here).  

The `pytest.mark.parametrize` is getting verbose; maybe it could be shortened with some generation or itertools, but I've left that for now.

gdmcbain:
Maybe a full factorial test against #686 isn't required, just a single test with flat doflocs; I doubt it interacts with the other factors.

Issue #670, opened 2021-07-19
speed benchmark


halbux:
Hi,

This is not an issue, just a question.
I am developing my fem library too, in c++, but before that I worked in Matlab and I remember being able to get decent speeds with fully vectorized code.
May I ask for a timing for my personal knowledge about what optimized python code can provide:

How long does it take to assemble a grad*grad (i.e. weak formulation of Laplace equation) on 1million quads at order 1 interpolation (and 2 and 3 as well if possible) on a structured quad square (no bc., just assembly time needed). Integration order should be exact for order 2 (so that's 4 gauss points if I am not mistaken).
Time should be from assembly start to having the matrix A (of Ax=b) fully ready to be used.
Corei7 4cores laptop would be best for the timing.

Thanks for your help!

kinnala:
There is something in the README using triangles. For quads I have nothing premade but can try it out when I get back from the holidays.

For quads, there is the need of using a more complicated reference mapping if you want to be general w.r.t the shape of the quad. That will certainly sacrifice some cycles which could be neglected by using a less general mapping which just moves and scales the reference quad. We also don't have a very good support for parallel assembly at the moment and, thus, those numbers in the README are single-threaded assembly.

halbux:
Ok, thank you. Yes assembly for general quads is good.

gdmcbain:
Just to give you a very quick preliminary idea, I've put together something rudimentary: [ex_670_speed_benchmark.py](https://github.com/gdmcbain/scikit-fem/blob/50166f6b31fbf0540bcf464550c44bf06cb6bf61/docs/examples/ex_670_speed_benchmark.py#L1-L12).

On my System 76 laptop running Pop!_OS 21.04 and just timing with
```shell
time python docs/examples/ex_670_speed_benchmark.py
```
(hence including meshing and building the basis), I get:
| mapping | *t*/s |
| ----------- | ------ | 
| affine | 1.8 | 
| isoparametric | 5.3 |

halbux:
Thanks!

What do you mean with affine?
For the isoparametric it can mean any order since you can use higher order elements for mesh curvature: are you talking about the usual bilinear shape function (4dofs) quad?

kinnala:
I've actually never tried if MappingAffine works for a quad mesh like is done above. This is not done by default, and the behaviour can be undefined since its not tested at all.

halbux:
That seems faster than the best I could ever obtain in my former matlab code some time ago!

[...]

PR #692, opened 2021-08-18
Mesh improvements

### LINKED ISSUE #639: Bug: Error on save MeshHex2()
Hi,

`mesh.save("out.vtk")` throws an Error for the following lines of code:

```python
skfem.MeshTet2().save('out.vtk')
```

and

```python
skfem.MeshHex2().save('out.vtk')
```

For the case of `MeshHex2` it seems that it is not implemented in the type matching dict in `skfem/io/meshio.py` whereas for the  `MeshTet2` case the element connectivity is of shape `(n, 4)` whereas meshio is expecting `(n, 10)` (n = number of elements).

### LINKED ISSUE #676: revise plot(MeshLine1)
Some questions about `plot(MeshLine)`, viz.

https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/skfem/visuals/matplotlib.py#L130-L131

were raised in working on the first unsteady one-dimensional example. #674 

Because efficiently animating evolving solutions in matplotlib involves something like [`matplotlib.lines.Line2D.set_data`](https://matplotlib.org/stable/api/_as_gen/matplotlib.lines.Line2D.html?highlight=set_data#matplotlib.lines.Line2D.set_data), as in the existing two-dimensional unsteady example

https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/docs/examples/ex19.py#L117

one needs to know the structure of the data used in the original plot of the initial condition.  Whereas this is just the solution's nodal DoFs in two dimensions

https://github.com/kinnala/scikit-fem/blob/9608b5f1bd0216dcc6c5c75b6faebb698bde339f/docs/examples/ex19.py#L99

https://github.com/kinnala/scikit-fem/blob/9608b5f1bd0216dcc6c5c75b6faebb698bde339f/docs/examples/ex19.py#L117

in one dimension, the initial plot is delegated to `plot_meshline` which has some rather involved preprocessing

https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/skfem/visuals/matplotlib.py#L144-L153

which needs to be equally applied at each step.

There was a proposal to extract this as a method `MeshLine1.segment_data` in  ad43d00 but this is where [questions were raised](https://github.com/kinnala/scikit-fem/pull/674#pullrequestreview-715485508):

> I don't know if this is needed though? Isn't `ax.plot(m.p[0], z)` enough? This could be just some copypasta from draw_mesh2d where appending None gives faster performance.

The main work of the first unsteady one-dimensional example #674 is to demonstrate how post-processing works in one dimension because one dimension is different to two and three, so if one-dimensional plotting should generally be done differently, that should be sorted out before writing the example.

By 
```python
    ax.plot(m.p[0], z)
```
or a variation with `np.argsort`, do you mean dispensing with `visuals.matplotlib.plot_meshline`?  I guess that that works well enough in

https://github.com/kinnala/scikit-fem/blob/d08af2a3f4607e28023656b1b6ad025b1a530aa1/docs/examples/ex16.py#L49

### LINKED ISSUE #688: periodic boundary conditions
Any ideas on periodic boundary conditions?  There's a brief mention in the JOSS paper.

I'm getting increasingly interested in problems of propagation, for which these often appear in textbook examples, so I'll soon look into throwing something together.

kinnala:
Fixes #688. Fixes #676. Fixes #639. See changelog for other changes.


tests/test_assembly.py
```diff
@@ -464,5 +465,67 @@ def threaded_nonsym(u, v, w):
...
+        ),
+    ]
+)
+def test_periodic_mesh_assembly(m, mdgtype, etype):
+    m = m.reordered(m.nodes_satisfying(lambda x: x[0] == 1))
+    m = m.with_boundaries({
+        'left': lambda x: x[0] == 0,
+        'right': lambda x: x[0] == 1,
+    })
+    mp = mdgtype.from_mesh(m, m.periodic_connectivity('right', 'left'))
```
kinnala, line 516: I want to somehow combine these three steps into one, but don't know yet what is the correct approach.

tests/test_assembly.py
```diff
@@ -464,5 +465,67 @@ def threaded_nonsym(u, v, w):
...
+            MeshQuad1DG,
+            ElementQuad1,
+        ),
+        (
+            MeshQuad().refined(2),
+            MeshQuad1DG,
+            ElementQuad2,
+        ),
+        (
+            MeshTri().refined(2),
```
kinnala, line 504: There is a caveat that no element can touch both of the periodic interfaces. I must figure out a way to detect that. Maybe by checking for duplicate node indices over each element?

README.md
```diff
@@ -254,6 +254,21 @@ with respect to documented and/or tested features.
...
+  to additional caching
+- Changed: `Mesh` string representation is now more informative
+- Changed: `Form.assemble` allows now only `DiscreteField` or 1d/2d `ndarray`
+  objects, and will automatically pass 1d `ndarray` objects to
+  `Basis.interpolate` for convenience
+- Changed: `FacetBasis` is now only a shorthand for `BoundaryFacetBasis` and no
+  more initializes `InteriorFacetBasis` or `MortarFacetBasis` if the keyword
+  argument `side` is passed to the constructor
+- Removed: the deprecated `Mesh.define_boundary` method
+- Removed: the unused `Mesh.validate` attribute
```
kinnala, line 271: See here for a summary of the changes. This basically paves the path to 4.0.0 because I also did some cleaning up. This was triggered by the change in `Mesh.__repr__` and since that's technically a backwards-incompatible change I decided to clean up the backlog a bit. I don't expect to release 4.0.0 yet but by the end of the year for sure.

skfem/assembly/__init__.py
```diff
@@ -7,7 +7,10 @@
 >>> m = fem.MeshTri()
 >>> e = fem.ElementTriP1()
 >>> m
-Triangular mesh with 4 vertices and 2 elements.
+<skfem MeshTri1 object>
+  Number of elements: 2
+  Number of vertices: 4
+  Number of nodes: 4
```
kinnala, line 13: This now better highlights the difference (if any) in the number of vertices (of the underlying topology) and the number of nodes in the finite element field describing the geometry.

kinnala: It also shows `boundaries` and `subdomains`. You can try it out in REPL.

skfem/assembly/form/form.py
```diff
@@ -48,29 +47,17 @@ def assemble(self,
...
-                w[k] = DiscreteField(np.array([z.value for z in w[k]]),
-                                     np.array([z.grad for z in w[k]]))
-            elif isinstance(w[k], tuple):
-                warnings.warn("In future, any additional kwargs to "
-                              "asm() must be of type DiscreteField. "
-                              "In most cases this deprecation is "
-                              "fixed replacing asm(..., w=w) "
-                              "by asm(..., w=DiscreteField(*w)).",
-                              DeprecationWarning)
-                w[k] = DiscreteField(*w[k])
```
kinnala, line 71: These were mostly unused and deprecated since forever.

skfem/assembly/basis/__init__.py
```diff
@@ -10,16 +8,4 @@
...
-    """alias of :class:`~skfem.assembly.BoundaryFacetBasis`"""
-    if side is None:
-        return BoundaryFacetBasis(*args, **kwargs)
-    warnings.warn("Initializing FacetBasis using the keyword argument side "
-                  "is deprecated. Use InteriorFacetBasis or MortarFacetBasis "
-                  "instead.", DeprecationWarning)
-    if 'mapping' in kwargs:
-        if hasattr(kwargs['mapping'], 'helper_to_orig'):
-            return MortarFacetBasis(*args, side=side, **kwargs)
-    return InteriorFacetBasis(*args, side=side, **kwargs)
```
kinnala, line 25: This is something I had before to be super compatible. I think it can be removed now because these features are not used too often.

docs/examples/ex13.py
```diff
@@ -63,7 +63,7 @@ def port_flux(w):
 current = {}
 for port, boundary in mesh.boundaries.items():
     fbasis = FacetBasis(mesh, elements, facets=boundary)
-    current[port] = asm(port_flux, fbasis, u=fbasis.interpolate(u))
+    current[port] = asm(port_flux, fbasis, u=u)
```
kinnala, line 66: This demonstrates that you don't need to call `Basis.interpolate` anymore if `u` is based on the same `Basis` object.

skfem/mesh/mesh.py
```diff
@@ -22,8 +22,7 @@ class Mesh:
     _boundaries: Optional[Dict[str, ndarray]] = None
     _subdomains: Optional[Dict[str, ndarray]] = None
     elem: Type[Element] = Element
-    affine: bool = False
-    validate: bool = False  # unused; for backwards compatibility
```
kinnala, line 26: `validate` attribute is not doing anything. However, some tests tried to set it so I had it here during 3.0.0 release.

skfem/mesh/mesh.py
```diff
@@ -757,3 +731,60 @@ def element_finder(self, mapping=None):
...
+        ix2 = remap[ix0]
+        reremap[ix1] = ix2
+
+        periodic_mesh = cls.from_mesh(reordered_mesh, reremap[t])
+
+        # store reordered mesh and reverse mapping
+        periodic_mesh._unperiodic = reordered_mesh
+        periodic_mesh._nodes = reremap
+
+        return periodic_mesh
```
kinnala, line 790: Should this not be part of `Mesh` but instead some `MeshDG` which is inherited by all of the new DG mesh types?

gdmcbain (commented):
By this, I mean that both tests should use either the explicit number or the analytical formula.  There are advantages to both, the number noticing even the smallest changes (although this can be too sensitive, e.g. across operating systems, BLAS-implementations, hardware, I don't know), the formula reassuring the engineer that the finite element model is doing the right thing.  Having the two different bases for the tests is confusing.

tests/test_examples.py
```diff
@@ -100,6 +100,8 @@ def runTest(self):
         import docs.examples.ex13 as ex
         u = ex.u
         A = ex.A
+        current = ex.current
+        self.assertAlmostEqual(current['ground'], -0.441415756775655)
```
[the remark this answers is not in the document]
gdmcbain: By this, I mean that both tests should use either the explicit number or the analytical formula. There are advantages to both, the number noticing even the smallest changes (although this can be too sensitive, e.g. across operating systems, BLAS-implementations, hardware, I don't know), the formula reassuring the engineer that the finite element model is doing the right thing. Having the two different bases for the tests is confusing.

skfem/assembly/__init__.py
gdmcbain: Nice, very nice; very useful.  I will use this a lot, a simple `print(mesh)` will be a standard in my logging, I think.  Ta.

docs/examples/ex13.py
gdmcbain: Oh, is this new?  Very nice.  This will make user-level scripts simpler and tidier.  So there will be a few other examples to follow ex13 in this simplification; I'll manage that as an exercise in the new idiom.  Ta.

skfem/mesh/mesh.py
gdmcbain: Right, yes, ...  I'm no OO-theorist, but I have idly wondered (since before ) whether `skfem.Mesh` might benefit from multiple inheritance; e.g. whether `MeshTri1` might derive from both `Mesh2D` and `MeshSimplicial` (and to what extent this might indeed define it, as it kind of does mathematically, I think).  I don't know whether that makes sense in practice or offers any practical advantage, but it's something that I daydreamed about.

And so yes, would that make sense here?  It makes sense to me, I think, if it works and doesn't incur too much overhead.

What are the other ways of implementing that kind of thing?  `functools.singledispatch` as in #287, [`abc`](https://docs.python.org/3/library/abc.html#module-abc) as suggested as an alternative [at the time](https://github.com/kinnala/scikit-fem/pull/287#pullrequestreview-336936688)?  

Yeah, looking at `mesh.py` now, it doesn't look too good having the `.periodic` method there when it only applies to the discontinuous subclasses and then they probably do still want to derive from their dimensional superclasses via their cell-type superclasses, e.g. `MeshTri1DG` <- `MeshTri1` <- `Mesh2D` <- `Mesh`.  But then nor would one wish to have to duplicate the implementation of periodicity in each discontinuous subclass.

Is there any drawback to multiple inheritance here?  I might look to see where this is tested and try it out.

tests/test_assembly.py
gdmcbain: It does seem unlikely in practice, having a single element spanning the entire domain in one direction, but it is good practice to cover these degenerate cases where possible.

I notice that `test_periodic_mesh_assembly` doesn't test `MeshLine1DG`, but then that would be because it hasn't been written yet.  Actually that's the one I wanted first.  I'll try adding that as an exercise.

gdmcbain: Well, that particular test won't work in one dimension because the Poisson&ndash;Dirichlet problem is undefined, but anyway.  A simple test might be the eigenvalues of the laplacian on a line-segment, with the modes being the sines and cosines?

More generally, I'm a little uncomfortable with `test_periodic_mesh_assembly` as if the implementation of periodicity didn't actually work at all but just left the nominally periodic boundaries as default homogeneous Neumann, these tests would pass&mdash;the solutions are all going to be independent of the periodic coordinate, aren't they?  Because the forcing is

https://github.com/kinnala/scikit-fem/blob/4a295b519510ff7db7ab7df6aa4ab875116b4406/tests/test_assembly.py#L557

I'm not saying that I doubt the implementation, just that the test doesn't test it and if the implementation gets broken later, we might not be notified.

Also do we want an example as well as a suite of tests?  I might try that, if I get `MeshLine1DG` up.

Besides the advection&ndash;diffusion example that I posted earlier, which isn't a great test, I wonder maybe what about a periodic forcing?  That might be less trivial: I mean the derivatives at the periodic boundary should be nonzero.

Say u (x) = sin (2 &pi; _x_), u' = 2 &pi; cos 2 &pi; _x_, u'' = -4 &pi;&sup2; sin 2 &pi; _x_, so force with sin 2&pi;_x_ on 0 < *x* < 1?

gdmcbain: No, I'm afraid I can't get `MeshLine1DG` going; I find that I don't understand the implementation.

I assumed
```python
class ElementLineP1DG(ElementLineP1):

    nodal_dofs = 0
    interior_dofs = 2
    dofnames = ['u'] * 2
...
class MeshLine1DG(MeshLine1):
    """A lineal mesh with a discontinuous topology."""
    elem: Type[Element] = ElementLineP1DG
    affine: bool = False
    sort_t: bool = False
```
but then
```python
m = MeshLine([0, 1, 2])
mp = MeshLine1DG.periodic(m, [-1], [0])
b = Basis(mp, ElementLineP1())
```
fails with 
```
/home/gdmcbain/src/scikit-fem/skfem/assembly/basis/abstract_basis.py:62: UserWarning: Unable to calculate DOF locations.
  warnings.warn("Unable to calculate DOF locations.")
  File "/home/gdmcbain/src/scikit-fem/mwe.py", line 9, in <module>
    b = Basis(mp, elem)
  File "/home/gdmcbain/src/scikit-fem/skfem/assembly/basis/cell_basis.py", line 68, in __init__
    self.basis = [self.elem.gbasis(self.mapping, self.X, j, tind=elements)
  File "/home/gdmcbain/src/scikit-fem/skfem/assembly/basis/cell_basis.py", line 68, in <listcomp>
    self.basis = [self.elem.gbasis(self.mapping, self.X, j, tind=elements)
  File "/home/gdmcbain/src/scikit-fem/skfem/element/element_h1.py", line 13, in gbasis
    invDF = mapping.invDF(X, tind)
  File "/home/gdmcbain/src/scikit-fem/skfem/mapping/mapping_isoparametric.py", line 175, in invDF
    J = [[self.J(i, j, X, tind=tind) for j in range(self.dim)]
  File "/home/gdmcbain/src/scikit-fem/skfem/mapping/mapping_isoparametric.py", line 175, in <listcomp>
    J = [[self.J(i, j, X, tind=tind) for j in range(self.dim)]
  File "/home/gdmcbain/src/scikit-fem/skfem/mapping/mapping_isoparametric.py", line 175, in <listcomp>
    J = [[self.J(i, j, X, tind=tind) for j in range(self.dim)]
  File "/home/gdmcbain/src/scikit-fem/skfem/mapping/mapping_isoparametric.py", line 70, in J
    out += p[i, t[itr, :]][:, None] * dphi[j]
IndexError: index 4 is out of bounds for axis 1 with size 4
```

Indeed, `mp.doflocs.shape` is `(1, 4)` but `mp.dofs.element_dofs` is
```python
array([[0, 2, 4],
       [1, 3, 5]])
```
so that the indices of the last column are `> mp.doflocs.shape[1]`.  Why does `mp.t` have three elements instead of two like `m.p`?  When I use `MeshTri1DG`, the shape of `.t` doesn't change from `m` to `mp`.

kinnala: Try to refine `MeshLine` once. Now you have two equal elements and I think some of the indexing routines cannot deal with that.

kinnala: Somethings not right, I need to look into this.

gdmcbain: > Try to refine MeshLine once. Now you have two equal elements

How do you mean?  `MeshLine().refined()` has p `[0, 1, 0.5]` and t `[[0, 2], [2, 1]]`, which seems correct?  The two elements are equal in length, but that doesn't seem like it should cause a problem.  What do you mean by &lsquo;equal&rsquo;?

kinnala: This was in your example: `m = MeshLine([0, 1, 2])`. If you periodify that, then you have elements `[0 1]` and `[1 0]` which does not work for the same reason that this thread was started because there is no check yet. In theory it should work but I think some of the indexing routines will consider those elements equal.

kinnala: Oh, you found a bug which was caused by having [0] in one of the sets.

kinnala: Another bug lurking somewhere.

kinnala: `MeshLine1.__post_init__` looks suspicious.

kinnala: Alright, that if-clause won't work here.

kinnala: This now looks better:
```python
from skfem import *
from typing import Type

class ElementLineP1DG(ElementLineP1):

    nodal_dofs = 0
    interior_dofs = 2
    dofnames = ['u'] * 2

class MeshLine1DG(MeshLine):
    """A lineal mesh with a discontinuous topology."""
    elem: Type[Element] = ElementLineP1DG
    affine: bool = False
    sort_t: bool = False

    def __post_init__(self):
        pass

m = MeshLine().refined(2)
mp = MeshLine1DG.periodic(m, [0], [1])
```

kinnala: > `MeshLine1.__post_init__` looks suspicious.

Can we move this initialization with 1D list to another constructor? `MeshLine1.init_linear([0, 1, 2])` or something like that.

kinnala: I have another idea: make `MeshLine` not a shorthand for `MeshLine1` but a function which will support these fancy alternative constructor parameters.

gdmcbain: That's a good idea.  Yes, this is #663 again.  I so often turn to the one-dimensional case thinking that it will be easier and find special obstacles.

gdmcbain: The next hurdle is that `MappingIsoparametric` isn't implemented in one dimension.

I copied `affine = False` from `MeshTriP1DG` without knowing its rationale.

```
  File "/home/gdmcbain/src/scikit-fem/mwe.py", line 22, in <module>
    b = Basis(mp, elem)
  File "/home/gdmcbain/src/scikit-fem/skfem/assembly/basis/cell_basis.py", line 68, in __init__
    self.basis = [self.elem.gbasis(self.mapping, self.X, j, tind=elements)
  File "/home/gdmcbain/src/scikit-fem/skfem/assembly/basis/cell_basis.py", line 68, in <listcomp>
    self.basis = [self.elem.gbasis(self.mapping, self.X, j, tind=elements)
  File "/home/gdmcbain/src/scikit-fem/skfem/element/element_h1.py", line 13, in gbasis
    invDF = mapping.invDF(X, tind)
  File "/home/gdmcbain/src/scikit-fem/skfem/mapping/mapping_isoparametric.py", line 177, in invDF
    detDF = self.detDF(X, tind, J=J)
  File "/home/gdmcbain/src/scikit-fem/skfem/mapping/mapping_isoparametric.py", line 165, in detDF
    raise Exception("Not implemented for the given dimension.")
Exception: Not implemented for the given dimension.
```

gdmcbain: I hit the lack of rhe isoparamereic line rhe other day in discussing the speed benchmark #670 and the erroneous laplacian for the affine rectangular quad.  I had thought that isooarametric would reduce to affine for a rectangular quad and tried to investigate rhat by looking at the assumed simpler but in fact unfortunately anomalous line.

gdmcbain: > I have another idea: make MeshLine not a shorthand for MeshLine1 but a function which will support these fancy alternative constructor parameters.

The more i think about this, rhe better an idea it seems. The use of `dataclasses` is fouled by coercive constructors, but that all goes away if we use a constructor function. Then the instances can always be instantiated with their attribures without having to modify them. The `MeshLine` doesn't have to be a class, it can just be a function, like `numpy.array` which constructs an `ndarray`.

kinnala: > I copied `affine = False` from `MeshTriP1DG` without knowing its rationale.

That's correct. The thinking is that if something else than 1 node per vertex is used in the geometry representation, then `MappingIsoparametric` should be used. I'll try fixing `MappingIsoparametric` for 1D.

skfem/mesh/mesh.py
kinnala: I read about dataclasses and multiple inheritance. It seems to be a messy subject because of `__post_init__` but I'll try to make a proposition and update this PR.

kinnala:
> There is [`numpy.full`](https://numpy.org/devdocs/reference/generated/numpy.full.html#numpy.full) for this.
> 
> Also, is −16 a safe sentinel? If `mesh.nvertices >= 16`, it's a valid index and won't raise an `IndexError`.
> 
> What about
> 
> ```
>         remaps = np.full(mesh.nvertices, mesh.nvertices, dtype=np.int64)
> ```
> 
> Or, if this mightn't so clearly stand out as a sentinel to someone reading the source code
> 
> ```
>         sentinel = mesh.nvertices
>         remap = np.full(mesh.nvertices, sentinel, dtype=np.int64)
> ```
> 
> and then later one can test like `remap == sentinel`?

Yes this is bit of a hack right now. I'll fix it after all tests are in place so that I don't fail some special case.

PR #693, opened 2021-08-23
take care reading MSH 4.1 with meshio

### LINKED ISSUE #691: from_meshio fails on old ex27
I was trying to regenerate a mesh for ex27 as part of the discussion with @bhaveshshrimali on the speed of pypardiso relative to `scipy.sparse` in #690 so that the number of elements could be varied but the old `make_mesh(make_geom(...))` code seems not to work with the new (#680 ?) `skfem.io.mesh.from_meshio`.

The code:
```python
from skfem import *
from skfem.io.meshio import from_meshio

from itertools import cycle, islice

import numpy as np

def make_geom(length: float, lcar: float) -> Geometry:
    # Barkley et al (2002, figure 3 a - c)
    geom = Geometry()

    points = []
    for point in [
        [0, -1, 0],
        [length, -1, 0],
        [length, 1, 0],
        [-1, 1, 0],
        [-1, 0, 0],
        [0, 0, 0],
    ]:
        points.append(geom.add_point(point, lcar))

    lines = []
    for termini in zip(points, islice(cycle(points), 1, None)):
        lines.append(geom.add_line(*termini))

    for k, label in [
        ([1], "outlet"),
        ([2], "ceiling"),
        ([3], "inlet"),
        ([0, 4, 5], "floor"),
    ]:
        geom.add_physical(list(np.array(lines)[k]), label)

    geom.add_physical(geom.add_plane_surface(geom.add_line_loop(lines)), "domain")

    return geom


if __name__ == "__main__":
    from pathlib import Path

    geom = make_geom(35.0, 0.2)

    with open(Path(__file__).with_suffix(".geo"), "w") as fgeo:
        fgeo.write(geom.get_code())

    mio = generate_mesh(geom, dim=2)
    print(mio)

    mesh = from_meshio(mio)
    print(mesh)

    mesh.save(Path(__file__).with_suffix(".vtk"))
```

When I run this with pygmsh==6.1.1 (before the big API change in pygmsh 7), I believe that this generates a valid `*.geo` file, which is accepted by Gmsh (4.8.4 here) and Gmsh can write this as MSH 4.1 file which is understood by meshio, but something goes wrong in `skfem.io.meshio.from_meshio`:
```
  File "/home/gdmcbain/src/scikit-fem/docs/examples/ex27_mesh.py", line 55, in <module>
    mesh = from_meshio(mio)
  File "/home/gdmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/skfem/io/meshio.py", line 93, in from_meshio
    facets = {k: [tuple(f) for f in
  File "/home/gdmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/skfem/io/meshio.py", line 94, in <dictcomp>
    np.sort(m.cells_dict[bnd_type][v[bnd_type]])]
IndexError: index 381 is out of bounds for axis 0 with size 380
```

I don't mean to suggest that we should rely upon `pygmsh<7` but I believe the GEO and MSH 4.1 files are valid, as is the `meshio.Mesh`, but I'll check.

The intermediate files are generated and interrogated with
```shell
python docs/examples/ex27_mesh.py  # say, depending on where the above is saved
gmsh -2 docs/examples/ex27_mesh.geo
meshio info docs/examples/ex27_mesh.msh
```

gdmcbain:
Fixes #691.


gdmcbain:
The VTU added here to `docs/examples/meshes` for exx 24, 27 was generated with
```shell
gmsh -2 -clscale 0.2  backward-facing_step.geo
```
and then (partly to recode the boundaries more rationally #680, but mostly because it's half the size)
```python
Mesh.load("backward-facing_step.msh").save("backward-facing_step.vtu")
```

Here the original GEO: [backward-facing_step.zip](https://github.com/kinnala/scikit-fem/files/7029107/backward-facing_step.zip)

gdmcbain:
Actually this expanded a bit beyond its original intention, I might withdraw it and split it  up.  The important part is ff83c84; we need gloves on when using `meshio.read` for MSH 4.1 and the latter remains a crucial format for `Mesh.load`.

Then dac551a which replaces the JSON with MSH 4.1 for the backward-facing step of exx 24, 27 is kind of good because it does test the reading of a real MSH 4.1 generated by Gmsh (4.8.4), but it is a rather big file to have in the repository, so maybe not such a good idea.  This can be deferred.

Then the question of whether there is a should be a preferred format for static files for meshes (the old JSON, MSH 4.1, VTU?). That probably deserves a new issue for discussion as there are a few competing considerations (size, speed, portability, dependence on external libraries, &c., &c., as discussed in #158 #680, &c.).

PR #738, opened 2021-09-12
Preparation for the next release

### LINKED ISSUE #412: mixed meshes
I wonder what are good ways to treat mixed meshes; e.g. a combination of triangles and quadrilaterals. I recently received one with wedges #411 and tetrahedra, but I would envisage treating the tri&ndash;quad case first. One of these arose by accident in one of the examples when Gmsh failed to combine all the triangles into quadrilaterals. #288 

I wonder whether this could be done with the library as is, using multiple meshes with the same  points, a bit like subdomains (but not literal `Mesh.subdomains`); a `meshio.Mesh` can support blocks of cells of different types on the one array of points.

### LINKED ISSUE #672: Assemble inverse
Elements like `ElementTriDG(ElementTriP2())` lead to block diagonal matrices whose inverse is another block diagonal matrix. It should be possible to directly assemble the inverse matrix by inverting the blocks before assembly. In this task I plan to look into adding a flag `Form.assemble(inverse=True)` which does just that.

### LINKED ISSUE #733: Add TrilinearForm
In #732 the following snippet was devised to assemble trilinear forms:
```python
from typing import Optional, Tuple, Any

import numpy as np
from numpy import ndarray

from skfem import *
from skfem.assembly.form.form import Form
from skfem.assembly.form.form import FormExtraParams


class TrilinearForm(Form):

    def _assemble(self,
                  ubasis: Basis,
                  vbasis: Optional[Basis] = None,
                  wbasis: Optional[Basis] = None,
                  **kwargs) -> Tuple[ndarray,
                                     ndarray,
                                     ndarray,
                                     ndarray,
                                     Tuple[int, int, int]]:

        if vbasis is None:
            vbasis = ubasis
        if wbasis is None:
            wbasis = ubasis

        nt = ubasis.nelems
        dx = ubasis.dx
        wdict = FormExtraParams({
            **ubasis.default_parameters(),
            **self.dictify(kwargs, ubasis),
        })

        # initialize COO data structures
        sz = (ubasis.Nbfun, vbasis.Nbfun, wbasis.Nbfun, nt)
        data = np.zeros(sz, dtype=self.dtype)
        rows = np.zeros(sz, dtype=np.int64)
        cols = np.zeros(sz, dtype=np.int64)
        mats = np.zeros(sz, dtype=np.int64)

        # loop over the indices of local stiffness matrix
        for k in range(ubasis.Nbfun):
            for j in range(vbasis.Nbfun):
                for i in range(wbasis.Nbfun):
                    mats[k, j, i] = wbasis.element_dofs[i]
                    rows[k, j, i] = vbasis.element_dofs[j]
                    cols[k, j, i] = ubasis.element_dofs[k]
                    data[k, j, i] = self._kernel(
                        ubasis.basis[k],
                        vbasis.basis[j],
                        wbasis.basis[i],
                        wdict,
                        dx,
                    )

        return data, mats, rows, cols, (wbasis.N, vbasis.N, ubasis.N)

    def assemble(self, *args, **kwargs) -> Any:
        return self._assemble(*args, **kwargs)

    def _kernel(self, u, v, w, params, dx):
        return np.sum(self.form(*u, *v, *w, params) * dx, axis=1)



def asm():
    m = MeshTri().refined(7)
    basis = Basis(m, ElementTriP1())
    from skfem.helpers import dot, grad
    out = TrilinearForm(lambda u, v, w, p: dot(w * grad(u), grad(v))).assemble(basis)
    import sparse
    arr = sparse.COO(np.array([out[1].flatten(), out[2].flatten(), out[3].flatten()]), out[0].flatten(), out[4], has_duplicates=True)
    k_p = np.random.rand(100, basis.N)
    res = sparse.tensordot(arr, k_p.T, axes=1, return_type=sparse.COO)

    A = []
    for mat in res.T:
        A.append(mat.to_scipy_sparse().tocsr())
    return A
```
Adding this to scikit-fem is feasible but I think the output type above (variable `out`) is not optimal. It should be directly compatible with `sparse.COO` constructor and other sparse tensor libraries (`tensorly`, `pytorch.sparse`, `tf.SparseTensor` and so on). This means that the indices should be returned as a single _ndims x ndata_ array.

Some reasonable tests must be figured out. Above the idea is to assemble 100 forms with different field. So therefore we assemble only 1 trilinear form and perform tensor dot products using `sparse` package.

### LINKED ISSUE #737: ElementTriDG.dofnames is incorrect
In the constructor of `ElementTriDG`, the attribute `dofnames` is not repeated correctly. This possibly affects other DG-element constructors as well. I noticed this when trying to do
`ElementTriDG(ElementTriP1()) * ElementTriP0()` and it raised an exception.

kinnala:
- Fixes #737 
- Add `ElementTriSkeletonP0` and `ElementTriSkeletonP1`
- Add HDG example 40
- Add mixed mesh example 41
- Fixes #733 
- Fixes #412 (only linear elements, remaining work: high order elements)
- Support a list of bases in `asm()`
- Add `ElementHexDG`
- Fixes #672 
- Add periodic mesh example 42


kinnala:
Doing some fancy hybrid DG on a skeleton mesh:
Figure_2

kinnala:
ex40_skeleton

kinnala:
ex41_solution

kinnala:
Again a quite large PR (sorry for that) but @gdmcbain if you have any general suggestions I'm happy to hear those.

kinnala:
Oh, one thing that's missing is a test for #733.

gdmcbain:
On demonstrating and testing the `TrilinearForm` #733, I wondered when I first heard about it whether it would be applicable to the Newton-linearized linear and bilinear forms arising from the trilinear advective term in the Navier&dash;Stokes equations.

https://github.com/kinnala/scikit-fem/blob/5c4c2b1b837d2e0b314318c1d4cc8290a3e496bb/docs/examples/ex27.py#L66-L67

https://github.com/kinnala/scikit-fem/blob/5c4c2b1b837d2e0b314318c1d4cc8290a3e496bb/docs/examples/ex27.py#L80

https://github.com/kinnala/scikit-fem/blob/5c4c2b1b837d2e0b314318c1d4cc8290a3e496bb/docs/examples/ex27.py#L83-L84

https://github.com/kinnala/scikit-fem/blob/5c4c2b1b837d2e0b314318c1d4cc8290a3e496bb/docs/examples/ex27.py#L97-L98

but I haven't looked into it.

gdmcbain:
One thing that might be good soon is an example demonstrating spatial periodicity. #688  Though there is already [`tests.test_manufactured.test_periodic_mesh_assembly`](https://github.com/kinnala/scikit-fem/blob/e3ad1f0b51604962d591e47b36bdbf3de0edc2d7/tests/test_manufactured.py#L416).

docs/examples/ex22.py
```diff
@@ -50,14 +50,14 @@ def edge_jump(w):
...
-        return h * ((dw1[0] - dw2[0]) * n[0] +\
+        return h * ((dw1[0] - dw2[0]) * n[0] +
                     (dw1[1] - dw2[1]) * n[1]) ** 2
 
     eta_E = edge_jump.elemental(fbasis[0], **w)
     
     tmp = np.zeros(m.facets.shape[1])
     np.add.at(tmp, fbasis[0].find, eta_E)
-    eta_E = np.sum(0.5*tmp[m.t2f], axis=0)
+    eta_E = np.sum(.5 * tmp[m.t2f], axis=0)
```
gdmcbain, line 60: This little change reminds me that this biggish collection of changes leading to the next release might be a reasonable opportunity to adopt black, should that be something that's wanted eventually.  As discussed previously, I don't have a strong opinion on it; but it seems to be popular and does have some advantages, such as new contributors not having to learn the house style.

docs/examples/ex41.py
```diff
@@ -0,0 +1,47 @@
...
+from skfem.models import laplace, unit_load
+
+fname = 'docs/examples/meshes/mixedtriquad.msh'
+out = ['cell_sets_dict']  # read boundary nodes from meshio
+m = [
+    Mesh.load(fname, force_meshio_type='triangle'),
+    Mesh.load(fname, force_meshio_type='quad', out=out),
+]
+e = [ElementTriP1(), ElementQuad1()]
+basis = [Basis(*me) for me in zip(m, e)]
```
gdmcbain, line 32: This could be written slightly more concisely as
```python
basis = list(map(Basis, m, e))
```
but it hardly makes a difference.  Maybe yours is more pythonic anyway.

What is fantastic though is how concise clear and pythonic (and &lsquo;skfemic&rsquo;!) this whole example is!  A great advance on the long-standing #412.

skfem/assembly/form/bilinear_form.py
```diff
@@ -18,9 +18,9 @@ class BilinearForm(Form):
     trial function ``u``, test function ``v``, and a dictionary of additional
     parameters ``w``.
 
-    >>> from skfem import BilinearForm, InteriorBasis, MeshTri, ElementTriP1
+    >>> from skfem import BilinearForm, Basis, MeshTri, ElementTriP1
     >>> form = BilinearForm(lambda u, v, _: u * v)
-    >>> form.assemble(InteriorBasis(MeshTri(), ElementTriP1())).todense()
+    >>> form.assemble(Basis(MeshTri(), ElementTriP1())).todense()
```
gdmcbain, line 23: Maybe `toarray`?  [`.todense`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.spmatrix.todense.html#scipy.sparse.spmatrix.todense) returns a [`numpy.matrix`](https://numpy.org/devdocs/reference/generated/numpy.matrix.html#) and

> It is no longer recommended to use this class, even for linear algebra. Instead use regular arrays. The class may be removed in the future.

gdmcbain (approved):

docs/examples/ex40.py
```diff
@@ -0,0 +1,58 @@
...
+then the continuity is recovered with the help of Lagrange multipliers defined
+on the mesh skeleton (i.e. a "skeleton mesh" consisting only of the edges of
+the original mesh).
+
+As usual for these so-called "hybridizable" methods, the resulting system can
+be condensed to the skeleton mesh only using Schur's complement.  However, this
+is not done here as the example is meant to simply demonstrate the use of
+finite elements defined on the mesh skeleton.
+
+"""
```
gdmcbain, line 14: This is super interesting.  I'd only heard of HDG before and didn't know what it was compared to other discontinuous Galerkin techniques.  Would it be helpful to add a reference for this?  Or perhaps the term &ldquo;hybridizable&rdquo; and its usage is standard & stable enough that whatever I or future readers find in zbMath or arXiv or wherever is going to be sufficient.  Something like (unread)

* Giacomini, M., Sevilla, R. & Huerta, A. (2021). [HDGlab: An Open-Source Implementation of the Hybridisable Discontinuous Galerkin Method in MATLAB.](https://doi.org/10.1007/s11831-020-09502-5) _Archives of Computational Methods in Engineering,_ 28, 1941&ndash;1986.  Open access

?

gdmcbain (approved):

docs/examples/ex22.py
kinnala: I'm using it in some of the smaller projects and automatically checking black-conformity in CI. I'll consider using it here aswell.

docs/examples/ex40.py
kinnala: Sure, I'm not an expert but usually it refers to the use of Lagrange multipliers on the mesh skeleton. This technically allows you to do a static condensation of the internal DOFs and, thus, reduce the overall number of DOFs. We cannot yet do static condensation easily in scikit-fem but I hope to support it in the future.

docs/examples/ex41.py
kinnala: Decided to use this, I didn't know that `map` works like this for non-functions. Cool!

docs/examples/ex22.py
kinnala: There are some choices in black that I don't like and I'd probably want to fine tune the config quite a bit so I'll postpone it. Let's open another issue for that I think.

kinnala:
> On demonstrating and testing the `TrilinearForm` #733, I wondered when I first heard about it whether it would be applicable to the Newton-linearized linear and bilinear forms arising from the trilinear advective term in the Navier‐Stokes equations.
> 
> https://github.com/kinnala/scikit-fem/blob/5c4c2b1b837d2e0b314318c1d4cc8290a3e496bb/docs/examples/ex27.py#L66-L67
> 
> https://github.com/kinnala/scikit-fem/blob/5c4c2b1b837d2e0b314318c1d4cc8290a3e496bb/docs/examples/ex27.py#L80
> 
> https://github.com/kinnala/scikit-fem/blob/5c4c2b1b837d2e0b314318c1d4cc8290a3e496bb/docs/examples/ex27.py#L83-L84
> 
> https://github.com/kinnala/scikit-fem/blob/5c4c2b1b837d2e0b314318c1d4cc8290a3e496bb/docs/examples/ex27.py#L97-L98
> 
> but I haven't looked into it.

Probably yes but it's a bit unclear to me when and how the sparse 3-tensor contraction can be done efficiently. See #733 and the linked discussion for more info.

docs/examples/ex40.py
gdmcbain: Thanks. Yes, it's clear enough what the script is doing—it's all very legible, as usual—I meant more about situating this approach within the literature.  But perhaps the name &lsquo;hybridizable discontinuous Galerkin&rsquo; suffices.

kinnala:
ex42_solution
@gdmcbain I used your example almost verbatim. I hope you don't mind.

docs/examples/ex40.py
kinnala: I've read a few papers but those were not very good because I had to, for example, guess which direction the normal vector should point to when integrating over the interior facets. Turns out they used two different meanings for **n** depending on how they write the integrals. If the integral is over element boundary, i.e. \partial T for the element T, they talk about outward normal. If the integral is explicitly over the edge, then they talk about some specific orientation of the normal which is fixed beforehands.

kinnala: I'll try to look if there is something that I can approve.

gdmcbain:
Sure. It's simple and gives an immediate visual display of periodicity.

PR #742, opened 2021-09-19
Add two triangle Lagrangian elements to scikit-fem

xida2020:
P3 cubic element and P4 quartic element


kinnala:
Added some tests and fixed doflocs in #744. Continuing the work there. Thanks for your contribution!

PR #746, opened 2021-09-20
Add ElementTri15ParameterPlate

xida2020:
15-parameter Triangle Plate Element, an H2-nonconforming element, can be used to solve four order biharmonic problems.


kinnala:
Continuing in #747

PR #748, opened 2021-09-21
Make Mesh.draw a simple wrapper to skfem.visuals.vedo

### LINKED ISSUE #361: vedo, formerly known as vtkplotter
As noted in #44, matplotlib distorts three-dimensional views uncontrollably.  Some very nice plots were added to #357 using vtkplotter.  It might be a worthwhile addition to `skfem.visuals` #286.  (I haven't looked at the licence or portability.)

kinnala:
Fixes #361.

This is quite cool I think. First `pip install vedo`. Then try
```python
from skfem import *
m = MeshHex().refined(3)
m.draw(point_data={'y': m.p[0]}).show()
```
Then pressing `4` and `l` on the keyboard gives this fancy picture:

screenshot

Interfacing with vedo is done through a VTK file because the API of vedo is quite complex and this seems to do most of the things that we'd like to do.

### DIFF
--- a/skfem/io/meshio.py
+++ b/skfem/io/meshio.py
@@ -191,7 +191,9 @@ def from_file(filename, out, **kwargs):
 
 def to_meshio(mesh,
               point_data=None,
-              cell_data=None):
+              cell_data=None,
+              encode_cell_data=True,
+              encode_point_data=False):
 
     t = mesh.dofs.element_dofs.copy()
     if isinstance(mesh, skfem.MeshHex2):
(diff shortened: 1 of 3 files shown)


kinnala:
For some reason I can't get this to work in `ipython`. It detects that I'm in Jupyter Notebook even though I'm not.

kinnala:
If there are `boundaries`, then this interactive mode shows only the first one. I wonder what is the easiest way to toggle between those.

gdmcbain:
I confirm that this works under MS-Windows too, after a `pip install vedo`.  Very nice.

kinnala:
Trying to do something about `boundaries`, waiting for https://github.com/marcomusy/vedo/issues/464 to get resolved.

kinnala:
I mean, it should simply work after the above issue gets resolved, so I guess this can be merged?

PR #774, opened 2021-11-04
Preliminary work on static condensation and related optimizations

kinnala:
Update on the changes:

- Added `skfem.helpers.jump`.
- `asm()` now accepts and uses the keyword argument `asm(..., to=list)` which, instead of summing the resulting `COOData` objects, will apply the function given via `to` to the resulting `map` object. E.g., `asm(..., to=list)` will return a list of `COOData` objects, each representing different blocks.
- Now empty `DiscreteField()` will represent a "zero field" which is taken into account in most of `skfem.helpers` and allows for optimizing vectorial forms: no need to calculate multiplications with huge dense arrays filled with zeros.
- `ElementComposite` will now pass `DiscreteField()` to the unused component instead of huge zero arrays. [This is the scariest part, after changes no tests are broken but I think it may in some untested edge cases break code that worked in the previous versions.]
- The support for zero `DiscreteField()` was added to `AbstractBasis.interpolate`.
- `COOData.tolocal()` and `COOData.fromlocal()` for getting a 3-tensor of local stiffness matrices and transforming that back to `COOData` internal format for creating, e.g., CSR matrices. This is for manipulating and possibly inverting local stiffness matrices using `np.linalg.inv`.
- `Element.condensed` will return a pair of new `Element` objects: one with all the interior DOFs and one with all the remaining DOFs.

All of this was done to support for static condensation of interior DOFs. Here is an example how to condense the bubble DOFs for MINI element in Stokes problem:
```python
import numpy as np
from skfem import *


m = MeshTri().refined(4)
el = ElementVector(ElementTriMini()) * ElementTriP1()
elems = el.condensed()
bases = [Basis(m, e) for e in elems]


@BilinearForm
def stokes(u, p, v, q, w):
    from skfem.helpers import ddot, grad, div
    return ddot(grad(u), grad(v)) - div(u) * q - div(v) * p - 1e-2 * p * q


K = asm(stokes, bases, bases, to=list)

# static condensation
A = K[3].tocsr()
B = K[2].tocsr()
invC = K[0].inverse().tocsr()
S = A - B.T @ invC @ B


D = bases[1].get_dofs()
Dup = bases[1].get_dofs(lambda x: np.isclose(x[1], 1))
x = np.zeros(S.shape[0])
x[Dup.all('u^1^1')] = 1
x = solve(*condense(S, x=x, D=D.all(['u^1^1', 'u^2^1'])))

(u, ub), (p, pb) = bases[1].split(x)

from skfem.visuals.matplotlib import plot, show
plot(pb, p)
show()
```
The syntax is not yet too cool but now it's technically possible.

### DIFF
--- a/docs/examples/ex04.py
+++ b/docs/examples/ex04.py
@@ -54,7 +54,7 @@ This is a nonlinear problem since we do not know a priori which subset
 from skfem import *
 from skfem.models.elasticity import (linear_elasticity, lame_parameters,
                                      linear_stress)
-from skfem.helpers import dot, ddot, prod, sym_grad
+from skfem.helpers import dot, ddot, prod, sym_grad, jump
 import numpy as np
 from skfem.io import from_meshio
 from skfem.io.json import from_file, to_file
@@ -110,8 +110,7 @@ limit = 0.3
 @BilinearForm
 def bilin_mortar(u, v, w):
     # jumps
-    ju = (-1.) ** w.idx[0] * dot(u, w.n)
-    jv = (-1.) ** w.idx[1] * dot(v, w.n)
+    ju, jv = jump(w, dot(u, w.n), dot(v, w.n))
     nxn = prod(w.n, w.n)
     mu = .5 * ddot(nxn, C(sym_grad(u)))
     mv = .5 * ddot(nxn, C(sym_grad(v)))
@@ -125,7 +124,7 @@ def gap(x):
 
 @LinearForm
 def lin_mortar(v, w):
-    jv = (-1.) ** w.idx[0] * dot(v, w.n)
+    jv = jump(w, dot(v, w.n))
     mv = .5 * ddot(prod(w.n, w.n), C(sym_grad(v)))
     return ((1. / (alpha * w.h) * gap(w.x) * jv - gap(w.x) * mv)
             * (np.abs(w.x[1]) <= limit))
--- a/docs/examples/ex07.py
+++ b/docs/examples/ex07.py
@@ -1,7 +1,7 @@
 """Discontinuous Galerkin method."""
 
 from skfem import *
-from skfem.helpers import grad, dot
+from skfem.helpers import grad, dot, jump
 from skfem.models.poisson import laplace, unit_load
 
 m = MeshTri.init_sqsymmetric().refined()
@@ -14,8 +14,7 @@ fb = [InteriorFacetBasis(m, e, side=i) for i in [0, 1]]
 
 @BilinearForm
 def dgform(u, v, p):
-    ju = (-1.) ** p.idx[0] * u
-    jv = (-1.) ** p.idx[1] * v
+    ju, jv = jump(p, u, v)
     h = p.h
     n = p.n
     return ju * jv / (alpha * h) - dot(grad(u), n) * jv - dot(grad(v), n) * ju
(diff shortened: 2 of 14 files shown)


gdmcbain:
Overall: great. Very nice.

I had to look at `unwrap` a while to work out what it's doing. The LHS made it clear in the end, but I'm still not sure how the first two arguments on the RHS are interpreted without looking at the source. I think this kind of function might be called `split` in other codes; maybe `unpack` is a clearer name, i'm not sure. There are other things called split in Python (for `str`, `ndarray`). But beyond the name, which is fine, i wonder whether the calling syntax can be clarified. I'll rhink about this after i look inside.

Tangential : is the P0 for pressure intentional ? I believe it's conventionally P1 with Mini velocity (Ern & Guermond 2004, &sect;4.2.4).

kinnala:
> The LHS made it clear in the end, but I'm still not sure how the first two arguments on the RHS are interpreted without looking at the source.

The second argument to unpack is essentially len(basis) which is currently not passed to the forms in asm(). The call to asm() must be interpreted as four distinct calls to Form.assemble() and finally a block matrix such as
[ A B' ]
[ B 0 ]
is created.

The helper unpack will put the actual contents of u (and v) either to u or p (and v or q) depending on which block is being assembled. The unused variables are zero and I changed the helpers to accommodate this.

Actually now that I think of it, there is an issue in the initialization of the bases. The second basis should be offset by the number of Dofs in the first basis to get a matrix with correct shape. There was a recent change in #761 which allows doing this.

> Tangential : is the P0 for pressure intentional ? I believe it's conventionally P1 with Mini velocity (Ern & Guermond 2004, §4.2.4).

That's my mistake.

gdmcbain:
> The helper unpack will put the actual contents of u (and v) either to u or p (and v or q) depending on which block is being assembled. The unused variables are zero and I changed the helpers to accommodate this.

Yes, I was still working through this, trying to see whether the `2` in this example couldn't be inferred in some way, e.g. here as `k = len(args)`, which would work here, but I haven't thought about all the other ways in which `unpack` might be expected to be called.

kinnala:
The Dofs offsetting is used already in master to make this ex04 more readable: dofs2 = Dofs(M, E, offset=dofs1.N)

kinnala:
> Yes, I was still working through this, trying to see whether the `2` in this example couldn't be inferred in some way, e.g. here as `k = len(args)`, which would work here.

So consider some kind of BilinearForm with where u and v are 3-tuples, e.g., Reissner-Mindlin plate or weak symmetry mixed elasticity, or some methods for Timoshenko beam problems. We could change asm() to pass the length of len(basis) to w and then its possible to drop the second argument to unpack. unpack is already specific to calling through asm() because w.idx doesn't exist otherwise. I'll add assertion which explains this.

kinnala:
It would be also possible to do Dofs offsetting in asm() without changing those Basis objects. The benefit would be that it's then compatible with plot and others requiring Basis. This would then require additional flag in asm(..., block_matrix=True).

kinnala:
Unexpected floating point instability appears.

kinnala:
I updated the first message to match the current reality. I'm almost abandoning the idea of introducing `unpack` because I was able to modify `ElementComposite` to do the same optimizations.

PR #775, opened 2021-11-12
Improve get_dofs

kinnala:
#764

PR #776, opened 2021-11-13
{find->get}_dofs in examples

gdmcbain:
Following merger of #775 as discussed in #764.


gdmcbain:
Comments on style welcome.  Setting out a nice idiom is an important part of `docs/examples`; I'm trying to do the right thing here rather than push my own way of doing things.

gdmcbain:
What's the right way to get all the boundary dofs as in 

https://github.com/kinnala/scikit-fem/blob/dea787125d9a71edcca64b65bdf70847c405dd7b/docs/examples/ex13.py#L43-L47

Simple replacement of find with get looks wrong.

I do get the correct answer with
```python
boundary_dofs = np.concatenate([basis.get_dofs(b) for b in mesh.boundaries])
```
but is that required?

kinnala:
> What's the right way to get all the boundary dofs as in
> 
> https://github.com/kinnala/scikit-fem/blob/dea787125d9a71edcca64b65bdf70847c405dd7b/docs/examples/ex13.py#L43-L47
> 
> Simple replacement of find with get looks wrong.
> 
> I do get the correct answer with
> 
> ```python
> boundary_dofs = np.concatenate([basis.get_dofs(b) for b in mesh.boundaries])
> ```
> 
> but is that required?

I interpret "all DOFs" to mean the set of all DOFs on the boundary of the domain which is `basis.get_dofs()`. However, what you want to do is to get "all DOFs on all named boundaries". Could something explicit like
```python
boundary_dofs = basis.get_dofs("ground") | basis.get_dofs("positive")
```
work? Another option would be to add support for 
```python
boundary_dofs = basis.get_dofs(["ground", "positive"])
```
as suggested in #764.

docs/examples/ex02.py
```diff
@@ -69,7 +69,17 @@
...
 
-m = MeshTri.init_symmetric().refined(3)
+m = (
+    MeshTri.init_symmetric()
+    .refined(3)
+    .with_boundaries(
+        {
+            "left": lambda x: x[0] == 0,
+            "right": lambda x: x[0] == 1,
+            "top": lambda x: x[1] == 1,
```
kinnala, line 79: I'm now using `np.isclose(...)` because I ran into some issues when the right hand side was non-integer float.

gdmcbain: Ouch. That's another reason for generally preferring to define boundaries at the CAD geometry stage rather than from a mesh.

I have had trouble using things like is-close when the extent of the geometry is physically very small (&micro;m, nm) but the coordinates have been measured in inappropriate units like SI metres. 

I try to only use float-equality when the float is being compared with a definite value that it may or may not have been assigned. I believe that that's reliable. (Whereas is-close would not be.) An exampke is the use of milestones in `pacopy.natural`.

In our functions that mesh segments, rectangles, and cuboids, aren't the terminal values of the coordinates preserved exactly?  ('Exactly' in the sense that floating-point equality with the same value is `True`, not exactly in the sense of the reals. )  So that `==` 0.0 or 1.0 should work for

https://github.com/kinnala/scikit-fem/blob/eb5a56562b19c3d525388f4a1298f78443ca35d1/docs/examples/ex02.py#L73

kinnala:
Added support for `get_dofs(('left', 'right'))` in #778. Maybe that could be useful in some of these?

gdmcbain:
Right, i do mean union rather than concatenation.  Either proposed syntax is good (union of get-dofs of str or get-dof of collection of str). I'll try that.

gdmcbain:
O. K., thanks, yes, #778 is nice. i'll use that. Ta.

kinnala:
I'll fix the bugs in #778 asap.

kinnala:
Now it should be fixed but I'll wait for the tests to pass.

docs/examples/ex02.py
```diff
@@ -91,18 +101,12 @@ def C(T):
...
-    dofs['left'].facet['u_n'],
-    dofs['right'].nodal['u'],
-    dofs['top'].nodal['u'],
-))
+D = np.concatenate(
+    [
+        ib.get_dofs("left").facet["u_n"],
+        *[ib.get_dofs(s).nodal["u"] for s in ["left", "right", "top"]],
+    ]
+)
```
kinnala, line 109: Something like `np.hstack([ib.get_dofs('left'), ib.get_dofs(('right', 'top')).all('u')])` after #778 ?

kinnala: I don't suggest using `dofs1 | dofs2` because I noticed it can lead to unexpected behavior. I suppose it should be deprecated.

docs/examples/ex13.py
```diff
@@ -40,10 +40,10 @@
 basis = Basis(mesh, elements)
 A = asm(laplace, basis)
 
-boundary_dofs = basis.find_dofs()
+boundary_dofs = np.concatenate([basis.get_dofs(b) for b in mesh.boundaries])
 
 u = basis.zeros()
-u[boundary_dofs['positive'].all()] = 1.
+u[basis.get_dofs("positive").all()] = 1.
```
kinnala, line 46: `basis.get_dofs(("positive", "ground")).all()`

docs/examples/ex02.py
gatling-nrl: I would argue that `==` in reals are either outright bugs or eggs that will turn into bugs when some other part of the code changes. In the case of `facets_satsifying` etc I think you're comparing to the midpoint of the facet. Even if `==` works for meshes generated by `skfem` it might fail meshes imported from other tools. Since you in general want people to copy the examples as starting points for their own problems, using `==` is probably putting bug-eggs in their code.

docs/examples/ex02.py
gatling-nrl: Is `hstack` going to be the recommended way to concatenate `DofsViews`?

kinnala: Yes, I think so. Using those `|` and `+` operators you may get unintuitive result if `skip` was used in `get_dofs`.

docs/examples/ex02.py
gdmcbain: would argue that == in reals are either outright bugs or eggs that will turn into bugs when some other part of the code changes

gdmcbain: > that == in reals are either outright bugs or eggs that will turn into bugs when some other part of the code changes

For my strenuous and ultimately successful argument against this widely accepted maxim (which should indeed be upheld almost everywhere), see nschloe/pacopy#6. Basically if x, a qnd b are float and one defines either x = a in one path or b in another then x == a (i.e. testing agzinst exactly the value possibly used to define the variable) is a meaningful test while a fuzzy comparison isn't since the nearness of b to a is unknown.

One way around this is what's in FreeFEM for their function that returns a meshed square.  I believe that such a function is an important special case and potentially worth treating specially. What they do is return the square with the four sides tagged; here we'd prepopulate the `.boundaries ` attribute with "left", "right", "floor", "ceiling". I think we might know the indices a priori without having to check the coordinates. However we often follow with `.refine` and i don't know whether rhat preserves boundaries (which would itself be a nice enhancement).

kinnala: refine (uniform) used to preserve tagged boundaries for MeshTri in the old version. Should be easy enough to copy paste that code. For other meshes requires some work.

gatling-nrl: Your argument makes sense. And in the narrow case of your example, where I have assigned a or b to x, and later want to test for a or b, I agree it will work... however I think using floats for sentinel values like this is eventually going to introduce a bug.

In the case of the `skfem` examples though, it is not even obvious to me you fit this paradigm. When I create a mesh (for example a unit square mesh) in `gmsh`, read it into python with `meshio`, use `skfem.io.from_meshio` on it, and then ask for facets where `x[0]` is _exactly_ zero or one, I feel uneasy. I didn't explicitly assign zero to those points in the mesh.

But again, as you rightly point out, `isclose` is not a magic bullet. There's just no single right answer about how close you have to be. I think what you and I are both really doing in this thread is making the very strong case for labelling points during mesh creation instead of finding them with `facets_satsfying`.

gdmcbain: > I think what you and I are both really doing in this thread is making the very strong case for labelling points during mesh creation instead of finding them with facets_satsfying.

Yes, that's it. 100% Let's start working on that for the built-ins. I'll launch a discussion if you don't beat me to it.

> When I create a mesh (for example a unit square mesh) in gmsh, read it into python with meshio, use skfem.io.from_meshio on it, and then ask for facets where x[0] is exactly zero or one

Of this is really bad practice and to be discouraged, as you say. If one generates the mesh in Gmsh, the boundaries should be identified there. Considerable work has gone into having meshio and scikit-fem read and respect those boundaries from MSH 2.2 and 4.1 formats. 

I think the simole built-in canonical meshes do serve a role, avoiding recourse to external meshers, e.g., but not only, pedagogical examples, so let's get this right. Thanks!

docs/examples/ex13.py
gdmcbain: No, I don't think so?  In this example the Dirichlet boundary is the disjoint union of `"positive"` and `"ground"` but the imposed values are unity on one and zero on the other.  There is also a homogeneous Neumann boundary.  I think of the three in physical terms as source, sink, and insulated, respectively.  Here I just want to assign ones to `"positive"`, leaving zeros on `"ground"`.  Later when the all the Dirichlet degrees of freedom are constrained, it will be something like the suggestion.

Ah, I see, this suggestion should apply to 

https://github.com/kinnala/scikit-fem/blob/eb5a56562b19c3d525388f4a1298f78443ca35d1/docs/examples/ex13.py#L43

Ta.

gdmcbain: Adopted in 334bec5.

docs/examples/ex02.py
gdmcbain: Thanks, yes, that's better.

docs/examples/ex02.py
gdmcbain: > In the case of the skfem examples though, it is not even obvious to me you fit this paradigm. When I create a mesh (for example a unit square mesh) in gmsh, read it into python with meshio, use skfem.io.from_meshio on it, and then ask for facets where x[0] is exactly zero or one, I feel uneasy. I didn't explicitly assign zero to those points in the mesh.

Note that this is not at all what is done in any of the skfem examples.  In exx 02, 03, 04 (`M`), 05, 11, 13, 25, 34, 42, which I believe are all of the examples that define boundaries using floating-point comparisons on coordinates, the meshes are not imported, they are generated using the internal functions and I believe that these functions do fit this paradigm (define `x = a`, then test `x == a`).  The examples that do import externally generated meshes, exx 04 (`m`), 17,  21, 24, 26 &ndash; 28, 35, 41, all already have boundaries or subdomains predefined in the externally generated mesh, as we're now recommending.

gdmcbain: > I'm now using `np.isclose(...)` because I ran into some issues when the right hand side was non-integer float.

@kinnala, could you share an MWE?  I can't reproduce this; e.g. `np.pi/2` works fine.

https://github.com/kinnala/scikit-fem/blob/76b27f8901eb710b8c6c2af4c5876d1fe880291b/docs/examples/ex13.py#L29-L35

gdmcbain:
This just leaves ex36.

https://github.com/kinnala/scikit-fem/blob/9ee89bac51620fad4b3dffdfa4eda78dda9acc39/docs/examples/ex36.py#L163-L180

docs/examples/ex02.py
kinnala: I can try to reproduce at some point. I was creating a mesh and then applied translate((0.1, 0)) and was trying to locate the right boundary after translate.

gdmcbain: Ah yes, I'd define boundaries before translating, as in 

https://github.com/kinnala/scikit-fem/blob/76b27f8901eb710b8c6c2af4c5876d1fe880291b/docs/examples/ex13.py#L29-L37

(for the reasons outlined above, viz. `x=a` then `x==a`, &c.)

Also on translation, I was wondering about why it was required in `ex04.M`

https://github.com/kinnala/scikit-fem/blob/57adf735347e4a05ea1541e883c91031041c1a73/docs/examples/ex04.py#L67-L71

instead of 

https://github.com/kinnala/scikit-fem/blob/76b27f8901eb710b8c6c2af4c5876d1fe880291b/docs/examples/ex04.py#L67-L70

(which I've taken the liberty of substituting in here).

kinnala: No reason whatsoever. :-)

kinnala:
Alright, the last one is slightly complicated, let's see if I can do something about it.

kinnala:
Here is something:
```
modified   docs/examples/ex36.py
@@ -144,7 +144,12 @@ def volume(w):
     return J
 
 
-mesh = MeshTet().refined(2)
+mesh = MeshTet().refined(2).with_boundaries({
+    'left': lambda x: x[0] < 1.e-6,
+    'bottom': lambda x: x[1] < 1.e-6,
+    'back': lambda x: x[2] < 1.e-6,
+    'front': lambda x: np.abs(x[2] - 1.) < 1e-6,
+})
 uelem = ElementVectorH1(ElementTetP2())
 pelem = ElementTetP1()
 elems = {
@@ -160,36 +165,17 @@ du = basis["u"].zeros()
 dp = basis["p"].zeros()
 stretch_ = 1.
 
-ddofs = [
-    basis["u"].find_dofs(
-        {"left": mesh.facets_satisfying(lambda x: x[0] < 1.e-6)},
-        skip=["u^2", "u^3"]
-    ),
-    basis["u"].find_dofs(
-        {"bottom": mesh.facets_satisfying(lambda x: x[1] < 1.e-6)},
-        skip=["u^1", "u^3"]
-    ),
-    basis["u"].find_dofs(
-        {"back": mesh.facets_satisfying(lambda x: x[2] < 1.e-6)},
-        skip=["u^1", "u^2"]
-    ),
-    basis["u"].find_dofs(
-        {"front": mesh.facets_satisfying(lambda x: np.abs(x[2] - 1.) < 1e-6)},
-        skip=["u^1", "u^2"]
-    )
+D = [
+    basis["u"].get_dofs('front').all('u^3'),
+    basis["u"].get_dofs('left').all('u^1'),
+    basis["u"].get_dofs('bottom').all('u^2'),
+    basis["u"].get_dofs('back').all('u^3'),
 ]
 
-dofs = {}
-for dof in ddofs:
-    dofs.update(dof)
-
-du[dofs["left"].all()] = 0.
-du[dofs["bottom"].all()] = 0.
-du[dofs["back"].all()] = 0.
-du[dofs["front"].all()] = stretch_
+du[D[0]] = stretch_
 
 I = np.hstack((
-    basis["u"].complement_dofs(dofs),
+    basis["u"].complement_dofs(np.hstack(D)),
     basis["u"].N + np.arange(basis["p"].N)
 ))
 ```

gdmcbain:
I'm pretty sure we can get rid of the `x[0] < 1.e-6` stuff, that's just an idiom common in FEniCS but not at all a good idea for reasons argued above.  (Meanwhile, I've launched #791 for predefining boundaries in the built-in mesh-generators.)

PR #789, opened 2021-11-15
Add InitVar 'validate' parameter to Mesh

gatling-nrl:
useful to suppress validation even when loglevel<=DEBUG

### DIFF
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -166,6 +166,7 @@ class Mesh:
                 **{name: self.facets_satisfying(test, boundaries_only)
                    for name, test in boundaries.items()}
             },
+            validate=False,
         )
 
     def with_subdomains(self,
@@ -187,7 +188,8 @@ class Mesh:
                 **({} if self._subdomains is None else self._subdomains),
                 **{name: self.elements_satisfying(test)
                    for name, test in subdomains.items()},
-            }
+            },
+            validate=False,
         )
 
     def _encode_point_data(self) -> Dict[str, List[ndarray]]:
@@ -646,6 +648,7 @@ class Mesh:
             self,
             doflocs=np.array([self.doflocs[itr] * factors[itr]
                               for itr in range(len(factors))]),
+            validate=False,
         )
 
     def translated(self, diffs):
@@ -662,6 +665,7 @@ class Mesh:
             self,
             doflocs=np.array([self.doflocs[itr] + diffs[itr]
                               for itr in range(len(diffs))]),
+            validate=False,
         )
 
     def mirrored(self,
@@ -702,6 +706,7 @@ class Mesh:
         return replace(
             self,
             doflocs=p,
+            validate=False,
         )
 
     def smoothed(self, fixed_nodes=None):
@@ -735,6 +740,7 @@ class Mesh:
         return replace(
             self,
             doflocs=p,
+            validate=False,
         )
 
     def _uniform(self):
@@ -848,6 +854,7 @@ class Mesh:
             self,
             doflocs=p,
             t=t,
+            validate=False,
         )
 
     def element_finder(self, mapping=None):
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -1,6 +1,6 @@
 import logging
 
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, replace, InitVar
 from typing import Callable, Dict, List, Optional, Tuple, Type, Union
 
 import numpy as np
@@ -34,6 +34,7 @@ class Mesh:
     # However, some algorithms (e.g., adaptive refinement) require switching
     # off this behaviour and, hence, this flag exists.
     sort_t: bool = False
+    validate: InitVar[bool] = True # run validation check if log_level<=DEBUG
 
     @property
     def p(self):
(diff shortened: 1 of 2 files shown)


skfem/mesh/mesh.py
```diff
@@ -418,7 +419,8 @@ def __post_init__(self):
             self.t = np.ascontiguousarray(self.t)
 
         # run validation
-        self.is_valid(debug=False)
+        if validate:
+            self.is_valid()
```
kinnala, line 423: By omitting debug=False, I think you are causing the time consuming checks to run always. Maybe get rid of debug kwarg and move log level check here?

skfem/mesh/mesh.py
```diff
@@ -34,6 +34,7 @@ class Mesh:
     # However, some algorithms (e.g., adaptive refinement) require switching
     # off this behaviour and, hence, this flag exists.
     sort_t: bool = False
+    validate: InitVar[bool] = True # run validation check if log_level<=DEBUG
```
kinnala, line 37: Flake fails here because there should be two spaces before the comment.

skfem/mesh/mesh.py
gatling-nrl: I added a logger.DEBUG msg any time is_valid gets called (this should probably stay to alert you that something expensive happened if you forgot debug was on). It seems like is_valid is getting called at least twice per mesh creation.

The traceback is just temporary while I try to figure out where these calls to `is_valid` are coming from.

gatling-nrl: I also added a `raise_` keyworkd to `is_valid`, which is occasionally useful to halt and get the stack trace when troubleshooting

gatling-nrl: Ah, there's a temp mesh in `from_meshio`

https://github.com/kinnala/scikit-fem/blob/926c165a21cacad65855a0c4e15409bf84763af7/skfem/io/meshio.py#L111-L112

Probably only should validate the actual mesh returned.

gatling-nrl:
I don't know what it wants for formatting on line 441... that's not my line. To be honest, that line is _really_ hard to read.

gdmcbain:
> I don't know what it wants for formatting on line 441... that's not my line. To be honest, that line is _really_ hard to read.

What I do in such cases is run `black` on the file, copy the reformatted line, undo the general blackening, replace the offending line with the blackened version.

gatling-nrl:
> What I do in such cases is run black on the file, copy the reformatted line, undo the general blackening, replace the offending line with the blackened version.

I split it into two lines. The extra variable helps communicate intent as well as shorten the line. I've never tried `black`... its looks aggressive :)

gdmcbain:
Yes, it is a bit. If you have worked out the line yourself then that's better; black can be handy when you can't be bothered with that. I had to learn it as it's mandated in other packages; e.g. meshio.

gatling-nrl:
Looks like `InitVar` has changed behavior since `3.7`...

gatling-nrl:
https://bugs.python.org/issue33805

gatling-nrl:
I think that calls to `replace` should have `validate=False` to fix this. Presumably one check is enough, on the original object. Thoughts?

kinnala:
What confuses me is that why is this only visible in the doctests?

gatling-nrl:
Ok, I updated all the calls to `replace` in `mesh.py`.

kinnala:
I think there are plenty of `replace` calls in other places as well. Wasn't the issue in the test coming from `refined`?

skfem/mesh/mesh.py
```diff
@@ -34,6 +34,7 @@ class Mesh:
     # However, some algorithms (e.g., adaptive refinement) require switching
     # off this behaviour and, hence, this flag exists.
     sort_t: bool = False
+    validate: InitVar[bool] = True  # run validation check if log_level<=DEBUG
```
kinnala, line 37: How about removing `InitVar` and just using `self.validate` in `__post_init__`?

gatling-nrl: Oh, just let `validate` be a normal property of the object? Sure, that's even simpler!

gatling-nrl: But.... isn't this going to cause excessive validating? I'm pretty sure you want `validate=False` when using `replace`. I don't know all the reasons why to call `replace` on a mesh, but it seems like you'd only want `is_valid` run on whatever the original thing was. What do you want to do?

kinnala: I think it's fine to validate in DEBUG log level just in case. Not all the algorithms we have are perfect.

gatling-nrl:
Ok, took `validate=False` out of the calls to `replace` and made it an normal property of a mesh. So now you'd just use `validate=False` when you knew you were constructing an invalid mesh.

skfem/mesh/mesh.py
```diff
@@ -1,6 +1,6 @@
 import logging
 
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, replace, InitVar
```
kinnala, line 3: Unused import.

kinnala, line 3:
```suggestion
from dataclasses import dataclass, replace
```

gatling-nrl: done

PR #803, opened 2021-11-21
Make meshio optional

kinnala:
Many use cases work without `meshio` so I thought about making it optional and suggesting in readme to install via `pip install scikit-fem[all]` to get also `meshio` and `matplotlib`. What do you think @gdmcbain? The minimal installation could be useful, e.g., in some libraries using `scikit-fem` as a dependency. I'm not including `vedo` in `[all]` because it pulls quite many dependencies and I've noticed the installation to be a bit unpredictable in different environments, e.g., in JupyterHub/Google Colab.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,9 +163,9 @@ The minimal dependencies for installing `scikit-fem` are
 many
 [examples](https://scikit-fem.readthedocs.io/en/latest/listofexamples.html) use
 [matplotlib](https://matplotlib.org/) for visualization and
-[meshio](https://github.com/nschloe/meshio) for loading external formats.  Some
-examples demonstrate the use of other external packages; see `requirements.txt`
-for a list of test dependencies.
+[meshio](https://github.com/nschloe/meshio) for loading/saving different mesh
+file formats.  Some examples demonstrate the use of other external packages;
+see `requirements.txt` for a list of test dependencies.
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -27,8 +27,12 @@ Features:
 
 The most recent release can be installed simply by
 ```
-pip install scikit-fem
+pip install scikit-fem[all]
 ```
+Specifying `[all]` includes `meshio` for loading
+and saving to external formats,
+and `matplotlib` for visualization.
+The minimal dependencies are `numpy` and `scipy`.
 You can also try the library in browser through [Google Colab](https://colab.research.google.com/github/kinnala/scikit-fem-notebooks/blob/master/ex1.ipynb).
 
 ## Examples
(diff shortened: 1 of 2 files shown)


gdmcbain:
Yes, definitely.  Myself I do almost always need meshio for both input and output but I can imagine applications where the mesh is simple to generate and the output is just the evaluation of a functional and yes it should be possible to import the core of scikit-fem for that without having to pull in unneeded dependencies.  (I will add though that meshio does seem to be doing a reasonable job of not growing in its requirements for too many additional dependencies.)

README.md
```diff
@@ -155,12 +159,13 @@ python -c "import pkg_resources; print(pkg_resources.get_distribution('scikit-fe
...
-[numpy](https://numpy.org/), [scipy](https://www.scipy.org/) and
-[meshio](https://github.com/nschloe/meshio).  In addition, many
+[numpy](https://numpy.org/) and [scipy](https://www.scipy.org/).  In addition,
+many
 [examples](https://scikit-fem.readthedocs.io/en/latest/listofexamples.html) use
-[matplotlib](https://matplotlib.org/) for visualization.  Some examples
-demonstrate the use of other external packages; see `requirements.txt` for a
-list of test dependencies.
+[matplotlib](https://matplotlib.org/) for visualization and
+[meshio](https://github.com/nschloe/meshio) for loading external formats.  Some
```
gdmcbain, line 166: I find the output perhaps even more important than the input; e.g.

https://github.com/kinnala/scikit-fem/blob/4f12dcad2822a3705e029dfa3f0544ff4f01858b/docs/examples/ex32.py#L161-L163

gdmcbain: ...so 'loading and saving.

Also, where it says 'external formats', does that imply an internal format? I thought `skfem.io.json` was deprecated.

gatling-nrl:
I'm not versed enough in pip to know all the intricacies of `pip install thing[*]` so it might not be possible to do what I suggest here.  But if there was a way for `pip install scikit-fem` to pull in everything it could possibly need and `pip install scikit-fem[bare]` or whatever name, I would argue for that. One of the most appealing parts of skfem is its lack of dependencies, so you guys are doing that exactly right.

(For comparison, I spend a month trying to get dolfinx (aka fenics 2.0) compiled on the system I'm using (HP Cray) and in the end it just wouldn't stop seg-faulting from some interaction with petsc.  skfem was up and running in 10 minutes.)

That being said, if a user actually types at the prompt `pip install scikit-fem` they should get what they need. It will be (I think) incredibly frustrating to new users to have to dig into the docs and realize they needed `[all]` to make skfem work as advertised. On the other hand, people that are trying to optimize or include skfem as a dependency won't mind having to read the docs to know they can use `[bare]` or whatever to get an extremely lean skfem.

kinnala:
I'll take it into consideration. Personally I think ImportError is rather obvious indicator that something is missing. Examples in the docs mostly work without meshio. I use Mesh.load almost never in my work because I'm calculating mostly in squares or cubes.

PR #825, opened 2021-12-06
Implement AbstractBasis.project & improvements to docs

kinnala:
L2 projection is a fairly simple concept but it gets tedious to write the same forms over and over again. Because projections are so common it makes sense to have it available as an utility inside the library to save some typing.

`L2_projection` turned into `project` turned into `projection` and tried to be the utility but it became complex and still didn't do everything we need. Moreover, judging by the examples it really didn't even do its existing job well.

After thinking about #816, I wondered what if it really was just the inverse of `basis.interpolate()`?
```python
>>> from skfem import *
>>> m = MeshTri()
>>> basis = Basis(m, ElementTriP1())
>>> y = m.p[1]
>>> (basis.project(basis.interpolate(y)) - y).sum()
-1.3877787807814457e-16
```
So here `basis.project()` takes in `y` as `DiscreteField` and solves the variational problem `(u, v) = (y, v)`. Super simple, only few lines of code.

Then I added a convenient shortcut: if the parameter is callable, then `basis.project()` will pass in the global quadrature points to create `y`:
```python
>>> basis.project(lambda x: x[0] ** 2)
array([-0.11111111,  0.80555556, -0.19444444,  0.88888889])
```
Gradient-averaged derivative is short and sweet:
```python
>>> basis.project(basis.interpolate(m.p[0]).grad[0])
array([1., 1., 1., 1.])
```
Real derivative is not much longer, you just need to initialize the P0 basis:
```python
>>> basis0 = basis.with_element(ElementTriP0())
>>> basis0.project(basis.interpolate(m.p[0]).grad[0])
array([1., 1.])
```
Using `FacetBasis` it will do `solve` only on the boundary DOFs and extend by zero in the interior (which is what you usually want to do, see changes to `howto.rst`.)

I feel this is finally going somewhere, and judging by the respective changes to the examples it really simplifies things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ The most recent release can be installed simply by
 pip install scikit-fem[all]
 ```
 Specifying `[all]` includes `meshio` for mesh input/output,
-and `matplotlib` for simple visualization.
+and `matplotlib` for simple visualizations.
 The minimal dependencies are `numpy` and `scipy`.
 You can also try the library in browser through [Google Colab](https://colab.research.google.com/github/kinnala/scikit-fem-notebooks/blob/master/ex1.ipynb).
 
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@
 </p>
 
 
-`scikit-fem` is a lightweight Python 3.7+ library for performing [finite element
-assembly](https://en.wikipedia.org/wiki/Finite_element_method). Its main purpose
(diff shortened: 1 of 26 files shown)


kinnala:
I will add extend `howto.rst` and add documentation/doctests for this so that I'm sure that all use cases are covered.

PR #832, opened 2021-12-20
ElementQuadRT0

gdmcbain:
As discussed in #831.

Currently this is failing as `DF` isn't implemented.


gdmcbain:
The `ex43` is just ex37 minimally rewritten to use quadrilaterals instead of tetrahedra; I wouldn't think it worth including in `docs/examples`.  It's failing without the `DF`, as mentioned above

```
  File "/home/gdmcbain/src/scikit-fem/docs/examples/ex43.py", line 12, in <module>
    basis = Basis(m, e)
  File "/home/gdmcbain/src/scikit-fem/skfem/assembly/basis/cell_basis.py", line 80, in __init__
    self.basis = [self.elem.gbasis(self.mapping, self.X, j, tind=elements)
  File "/home/gdmcbain/src/scikit-fem/skfem/assembly/basis/cell_basis.py", line 80, in <listcomp>
    self.basis = [self.elem.gbasis(self.mapping, self.X, j, tind=elements)
  File "/home/gdmcbain/src/scikit-fem/skfem/element/element_composite.py", line 101, in gbasis
    output.append(e.gbasis(mapping, X, ind, tind)[0])
  File "/home/gdmcbain/src/scikit-fem/skfem/element/element_hdiv.py", line 21, in gbasis
    DF = mapping.DF(X, tind)
  File "/home/gdmcbain/src/scikit-fem/skfem/mapping/mapping.py", line 101, in DF
    raise NotImplementedError
NotImplementedError
```

as is `pytest -k ConvergenceQuadRT0 tests/test_convergence_hdiv.py`.

This `ex43` does work with `ElementTriRT0`.  The implementation of `ElementQuadRT0` is very like `ElementTriRT0`.

kinnala:
I think `MappingIsoparametric.DF` is essentially `MappingIsoparametric.J` without `i` and `j` arguments. Maybe we can get rid of `J` altogether: isn't it always called for all values of `i` and `j`?

kinnala:
However, there are some optimizations in `MappingIsoparametric.J` so maybe it's safer to make another method for now.

gdmcbain:
Thank you. Yes I took a few more steps with this here this afternoon and was on the track of seeing which indices of DF a.k.a. J were wanted. I figured "J" is for Jacobian, of F, and that DF is something pretty similar. Ta!

gdmcbain:
With this simple `MappingIsoparametric.DF`, based on the `J` in `detDF`, 

https://github.com/kinnala/scikit-fem/blob/ef493c0735fcc5d3dc09ad0608a83117b4114065/skfem/mapping/mapping_isoparametric.py#L159-L164

the simple example here called "ex43" works:

Figure_1

gdmcbain:
> FAILED tests/test_convergence_hdiv.py::ConvergenceQuadRT0::runTest - AssertionError: 0.06297384149299402 not less than 0.04

Not sure whether this is significant.  The tolerance is inherited in the new test from `ConvergenceRaviartThomas`

https://github.com/kinnala/scikit-fem/blob/cbdc6504e0eccdba49757e80636fffd5d205363a/tests/test_convergence_hdiv.py#L16

gdmcbain:
I see that other tests have loosened this, e.g. `ConvergenceBDM1`,

https://github.com/kinnala/scikit-fem/blob/cbdc6504e0eccdba49757e80636fffd5d205363a/tests/test_convergence_hdiv.py#L143

gdmcbain:
As I say, I don't think the mwe used during development is interesting enough to include in `docs/examples`, but I'll link to it here.

https://github.com/kinnala/scikit-fem/blob/881a7edc88b1323a07b399c6adda6323aa5747c9/docs/examples/ex43.py#L1-L35

kinnala:
Can you add the element to this table? https://github.com/kinnala/scikit-fem/blob/master/docs/api.rst#module-skfemelement

Unfortunately, I was not able to generate such a table automatically in Sphinx.

kinnala:
Otherwise it looks great! If you want, you can also add a changelog entry to README.md, there are plenty of lines adding new elements so you can just copy paste one of those.

PR #834, opened 2021-12-23
avoid DeprecationWarning from union of Dofs

gdmcbain:
The deprecation doesn't say what to do instead of union, but stacking and sorting passes the test.

### DIFF
--- a/docs/examples/ex24.py
+++ b/docs/examples/ex24.py
@@ -29,7 +29,7 @@ element = {'u': ElementVector(ElementTriP2()),
 basis = {variable: Basis(mesh, e, intorder=3)
          for variable, e in element.items()}
 
-D = basis['u'].get_dofs(mesh.boundaries)
+D = basis['u'].get_dofs(['inlet', 'ceiling', 'floor'])
 
 A = asm(vector_laplace, basis['u'])
 B = -asm(divergence, basis['u'], basis['p'])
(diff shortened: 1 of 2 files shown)


gdmcbain:
To reproduce the problem without this patch:
```shell
pytest -k TestDofsFromBoundaries tests
```
which raises
```
tests/test_dofs.py::TestDofsFromBoundaries::runTest
  /home/gdmcbain/src/scikit-fem/skfem/assembly/dofs.py:201: DeprecationWarning: Use numpy.hstack to combine sets of DOFs
    warn("Use numpy.hstack to combine sets of DOFs", DeprecationWarning)
```

gdmcbain:
There is another `DeprecationWarning` raised in the same module.
```shell
pytest -k TestDofsNodalSubsets tests/test_dofs.py 
```
raises
```
tests/test_dofs.py::TestDofsNodalSubsets::runTest
  /home/gdmcbain/miniconda3/envs/skfem/lib/python3.9/site-packages/numpy/core/fromnumeric.py:43: DeprecationWarning: Calling nonzero on 0d arrays is deprecated, as it behaves surprisingly. Use `atleast_1d(arr).nonzero()` if the old behavior was intended.
    result = getattr(asarray(obj), method)(*args, **kwds)
```
but I don't think it's related and don't know exactly where it's coming from.

PR #841, opened 2021-12-30
Add standard hyperelasticity example

kinnala:
Example 36 includes two field formulation of incompressible hyperelasticity. This is an example of the classical approach to hyperelasticity by @adtzlr from #839.


kinnala:
Also new pic for Example 21:
scren

Issue #847, opened 2022-01-03
'DiscreteField' object has no attribute 'reshape'


gatling-nrl:
This leads to confusing behavior in forms.

```python
def some_functional():
    def _form(w):
        # arrays are shaped to group by facet, but we don't care for that here so reshape(2,-1)
        ii = w['ind_p0'].reshape(2,-1).T
        gg = grad(w['ind_p1']).reshape(2,-1).T
        pp = w.x.reshape(2,-1).T  
        nn = w.n.reshape(2,-1).T
        ...
        return w.x[0] # must return *something* of the right shape
    return skfem.Functional(_form)

skfem.asm(
    some_functional(),
    skfem.InteriorFacetBasis(mesh, skfem.ElementTriP1(), side=0),
    ind_p1=ind_p1,
    ind_p0=basis_p1.project(basis_p0.interpolate(ind_p0)),
)
```

`gg`, `pp`, `nn` all work, but `ii` returns 'DiscreteField' object has no attribute 'reshape'.

The "work around" or maybe the intended usage appears to be 

```python
ii = w['ind_p0'].value.reshape(2,-1).T
```

and this might be okay, except that

```python
gg = grad(w['ind_p1']).reshape(2,-1).T
```

strongly implies (imo) that `w['ind_p1']` _is_ the "value". Moreover `w[ind_p0]**2` works, and returns `(w[ind_p0].value)**2` making me think maybe  `w[ind_p0]` supports any `ufunc`. 

I think `reshape` should work.

kinnala:
You have it correct, the idea has been to have DiscreteField behave like DiscreteField.value, mainly for convenience. Would it be reasonable to simply add DiscreteField.reshape which returns reshape applied to the value?

gatling-nrl:
That works, but there might be other `ndarray` methods that come up. Is duck typing DiscreteField so it acts like an ndarray the best implementation? Or should it also subclass ndarray. What is the difference between the return value form `solve` and a DiscreteField?

What's the difference between `skfem.helpers.grad(w['thing'])` and `w['thing'].grad`? Is `w['thing'].grad` a lazy evaluation (i.e. not created until it is used somewhere?

[...]

PR #848, opened 2022-01-03
DiscreteField delegates to its value

gdmcbain:
Demonstrating a suggested alternative to deriving from `numpy.ndarray`, as suggested in #847.

### DIFF
--- a/docs/examples/ex24.py
+++ b/docs/examples/ex24.py
@@ -29,7 +29,7 @@ element = {'u': ElementVector(ElementTriP2()),
 basis = {variable: Basis(mesh, e, intorder=3)
          for variable, e in element.items()}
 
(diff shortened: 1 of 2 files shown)


gatling-nrl (commented):
This doesn't change the behavior when iterating over DiscreteField which can have surprising results, see https://github.com/kinnala/scikit-fem/issues/847#issuecomment-1004392959

kinnala:
> This doesn't change the behavior when iterating over DiscreteField which can have surprising results, see [#847 (comment)](https://github.com/kinnala/scikit-fem/issues/847#issuecomment-1004392959)

I don't think this is really an issue. I don't see when you would want to iterate like that. It just never makes any sense.

PR #849, opened 2022-01-04
Make DiscreteField subclass ndarray

### LINKED ISSUE #847: 'DiscreteField' object has no attribute 'reshape'

kinnala:
This PR turns `DiscreteField` into a subclass of `ndarray` instead of `NamedTuple`.
The benefit is that now `DiscreteField` objects automatically behave just like `DiscreteField.value` previously did and there is hopefully no more need to write `u.value` inside the forms.

As a side effect you can do stuff like this:
```python
In [1]: from skfem import *                                                                              
In [2]: m = MeshTri()                                                                                    
In [3]: basis = Basis(m, ElementTriP1())                                                                 
In [4]: fun = basis.project(lambda x: x[0]) 
In [7]: @LinearForm 
   ...: def linf(v, w): 
   ...:     return w.fun.grad[0] * v 
   ...:                                                                                                  
In [8]: linf.assemble(basis, fun=fun)                                                                      
Out[8]: array([0.16666667, 0.33333333, 0.33333333, 0.16666667])
```
Traditionally such form would fail because `w.fun` would be just `ndarray` and not `DiscreteField`.

Fixes #847.

### DIFF
--- a/skfem/element/discrete_field.py
+++ b/skfem/element/discrete_field.py
@@ -79,7 +79,8 @@ class DiscreteField(ndarray):
         rep = ""
         rep += "<skfem DiscreteField object>"
         if not self.is_zero():
-            rep += "\n  Quadrature points per element: {}".format(self.shape[-1])
+            rep += ("\n  Quadrature points per element: {}"
+                    .format(self.shape[-1]))
             rep += "\n  Number of elements: {}".format(self.shape[-2])
             rep += "\n  Order: {}".format(len(self.shape) - 2)
         rep += "\n  Attributes: {}".format(
--- a/skfem/element/discrete_field.py
+++ b/skfem/element/discrete_field.py
@@ -4,86 +4,73 @@ import numpy as np
 from numpy import ndarray
 
 
-class DiscreteField(NamedTuple):
-    """A function defined at the global quadrature points.
+class DiscreteField(ndarray):
 
-    Created using :meth:`~skfem.element.Element.gbasis` or
-    :meth:`~skfem.assembly.CellBasis.interpolate`.
+    def __new__(cls,
+                value=np.array([0]),
+                grad=None,
+                div=None,
+                curl=None,
+                hess=None,
+                grad3=None,
+                grad4=None,
+                grad5=None,
+                grad6=None):
+        obj = np.asarray(value).view(cls)
+        obj.grad = grad
+        obj.div = div
+        obj.curl = curl
+        obj.hess = hess
+        obj.grad3 = grad3
+        obj.grad4 = grad4
+        obj.grad5 = grad5
+        obj.grad6 = grad6
+        return obj
 
-    """
+    def __array_finalize__(self, obj):
+        if obj is None: return
+        self.grad = getattr(obj, 'grad', None)
+        self.div = getattr(obj, 'div', None)
+        self.curl = getattr(obj, 'curl', None)
+        self.hess = getattr(obj, 'hess', None)
+        self.grad3 = getattr(obj, 'grad3', None)
+        self.grad4 = getattr(obj, 'grad4', None)
+        self.grad5 = getattr(obj, 'grad5', None)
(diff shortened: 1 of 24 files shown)


skfem/element/discrete_field.py
```diff
@@ -52,28 +52,31 @@ def __array_wrap__(self, out_arr, context=None):
...
+        return getattr(self, self._extra_attrs[n - 1])
+
     @property
     def astuple(self):
         return tuple(self.get(i) for i in range(len(self._extra_attrs) + 1))
 
     @property
     def value(self):
+        # increase backwards-compatibility
         return self
```
gatling-nrl, line 67: Is this line right? It allows `x.value.grad`. Previously `.value` was a naked ndarray. This is subtle and the line might fine. I wanted to point out the change from previous behavior since this is here for backward comapt.

kinnala: Alright, good point. This is the reason why w.fun.grad now also works. Test suite passes, no changes related to this required. Since we are already breaking the backwards compatibility to some extent we could do it here also. I need to think about it.

kinnala: One option is to cast to ndarray here and then in FormParams not use DiscreteField.value.

gatling-nrl: Can you go so far as to say that `.value` is deprecated? I'm not sure there is a use case for it now, other than supporting previously written code?

skfem/element/discrete_field.py
```diff
@@ -4,86 +4,73 @@
...
+                value=np.array([0]),
+                grad=None,
+                div=None,
+                curl=None,
+                hess=None,
+                grad3=None,
+                grad4=None,
+                grad5=None,
+                grad6=None):
+        obj = np.asarray(value).view(cls)
```
gatling-nrl, line 28: This line is wrong. I tried to fix and push, but duh, this failed obviously. 

Anyway, the correct code should use None in the function def and then

```python
        if value is None:
            obj = np.asarray([0]).view(cls)
        else:
            obj = np.asarray(value).view(cls)
```

See [here](https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html) for why, if it matters.

kinnala: Thanks!

skfem/element/discrete_field.py
kinnala: Fair enough, should get rid of `.value` from all code under `skfem`. I have the tendency of leaving around deprecated stuff so that I remember the state of affairs and support those as long as it's reasonable.

PR #851, opened 2022-01-04
Add SubdomainFacetBasis

gatling-nrl:
Start by refactoring `BoundaryFacetBasis` so we have

```python
class AbstractBasis: ...
class CellBasis(AbstractBasis): ...
class FacetBasis(AbstractBasis): ...
class BoundaryFacetBasis(FacetBasis): ...  # perhaps this will just become an alias?
class InteriorFacetBasis(FacetBasis): ...
class MortarFacetBasis(FacetBasis): ...
```
Then add
```python
class SubdomainFacetBasis(AbstractBasis): ...
```
Not planning to do in this PR, but seems like a natural evolution...
```python
class SubdomainCellBasis(AbstractBasis): ...
```

`SubdomainFacetBasis` should accept a set of mesh elements that define the subdomain. Elements that have a boundary facet are not allowed (although this is probably up for debate). The FacetBasis will be the external boundary of the subdomain and will define outward pointing normals at each quadrature point along this boundary.


gatling-nrl:
@kinnala Not actually trying to merge this yet. Just started PR to get the checks.

By the way, black and whatever is checking formatting in skfem disagree about how to do spaces around colons in slices.

gatling-nrl:
@gdmcbain If you thought I abused `asm` and `Functional` before... fasten your seatbelt mate.

@kinnala I'm _quite sure_ there's a better way, but hey, it does kinda work.

kinnala:
Why the refactoring and so many unrelated changes though? Hard for me to judge what has actually changed. Shouldn't this be just adding one file under skfem.assembly.basis and then some imports? I don't use black so no idea why it fails.

skfem/assembly/basis/subdomain_facet_basis.py
```diff
@@ -0,0 +1,80 @@
...
+from .cell_basis import CellBasis
+from .facet_basis import FacetBasis
+from .interior_facet_basis import InteriorFacetBasis
+from ..dofs import Dofs
+from ..form import Functional
+
+logger = logging.getLogger(__name__)
+
+
+class SubdomainFacetBasis(FacetBasis):
```
kinnala, line 20: This is the relevant part I presume.

gatling-nrl: > Why the refactoring and so many unrelated changes though? 

Yeah, I just dropped a grenade in, partly by accident. The problems started with me being irked that things things weren't really `BoundaryFacetBasis` and so I thought I could maybe refactor to `FacetBasis` easily enough, and it would all just be okay. So I tried it.

Last time I tried a pull request I was overrun with formatting bugs and someone suggested black. I hadn't used it prior, but now admit its a nice tool... anyway I thought it was the secret sauce to fixing format/style issues. I was wrong. That caused a ton of line changes that I didn't mean, but I didn't realize until the tests started failing that black was perhaps not the answer. By that point it was too late, and you see the end result. :(

> This is the relevant part I presume.

Yes, definitely. And I already feel like it is a train wreck of an implementation. But I hope you see some interesting ideas you hadn't thought of before.

gatling-nrl:
Ok, I unraveled the mess. `SubdomanFacetBasis` can't be added to the `__init__.py` files yet because it can't use `Functional` without making a circular import. I'll work on getting rid of the `Functional` and `asm` today.

gatling-nrl:
@kinnala The main things happening in this PR now are

1. Add SubdomainFacetBasis
2. Fixes for _tind_normals (I merged your V2 into this one)
3. Move normalize_* to Mesh class
4. Clean up some of the `__init__.py` stuff

kinnala:
Let me suggest some tests:
- Take some _continuous_ function and integrate over the subdomain boundary with SubdomainFacetBasis and InteriorFacetBasis. Should get the same result? With discontinuous function the mean value should be equal?
- Take some vector-valued function like f(x,y) = (x, 1) for which divergence is 1. Then by Gauss divergence theorem, the area of the subdomain should be equal to the integral of w.x[0] * w.n[0] + 1. * w.n[1] over the subdomain boundary?

kinnala:
To keep on track with the imports, can you show me what is the output of this:
```python
In [3]: g1 = globals().copy().keys()
In [4]: from skfem import *
In [6]: set(globals().keys()) - set(g1)
Out[6]: 
{'Basis',
 'BilinearForm',
 'BoundaryFacetBasis',
 'CellBasis',
 'DiscreteField',
 'Dofs',
 'DofsView',
 'Element',
 'ElementComposite',
 'ElementDG',
 'ElementGlobal',
 'ElementH1',
 'ElementHcurl',
 'ElementHdiv',
 'ElementHex0',
 'ElementHex1',
 'ElementHex1DG',
 'ElementHex2',
 'ElementHexDG',
 'ElementHexS2',
 'ElementLineHermite',
 'ElementLineMini',
 'ElementLineP0',
 'ElementLineP1',
 'ElementLineP1DG',
 'ElementLineP2',
 'ElementLinePp',
 'ElementQuad0',
 'ElementQuad1',
 'ElementQuad1DG',
 'ElementQuad2',
 'ElementQuadBFS',
 'ElementQuadDG',
 'ElementQuadP',
 'ElementQuadRT0',
 'ElementQuadS2',
 'ElementTetCCR',
 'ElementTetCR',
 'ElementTetDG',
 'ElementTetMini',
 'ElementTetN0',
 'ElementTetP0',
 'ElementTetP1',
 'ElementTetP2',
 'ElementTetRT0',
 'ElementTri15ParamPlate',
 'ElementTriArgyris',
 'ElementTriBDM1',
 'ElementTriCCR',
 'ElementTriCR',
 'ElementTriDG',
 'ElementTriHermite',
 'ElementTriMini',
 'ElementTriMorley',
 'ElementTriP0',
 'ElementTriP1',
 'ElementTriP1DG',
 'ElementTriP2',
 'ElementTriP3',
 'ElementTriP4',
 'ElementTriRT0',
 'ElementTriSkeletonP0',
 'ElementTriSkeletonP1',
 'ElementVector',
 'ElementVectorH1',
 'ElementWedge1',
 'ExteriorFacetBasis',
 'FacetBasis',
 'Functional',
 'InteriorBasis',
 'InteriorFacetBasis',
 'LinearForm',
 'MappingAffine',
 'MappingIsoparametric',
 'MappingMortar',
 'Mesh',
 'Mesh2D',
 'Mesh3D',
 'MeshHex',
 'MeshHex1',
 'MeshHex1DG',
 'MeshHex2',
 'MeshLine',
 'MeshLine1',
 'MeshLine1DG',
 'MeshQuad',
 'MeshQuad1',
 'MeshQuad1DG',
 'MeshQuad2',
 'MeshTet',
 'MeshTet1',
 'MeshTet2',
 'MeshTri',
 'MeshTri1',
 'MeshTri1DG',
 'MeshTri2',
 'MeshWedge1',
 'MortarFacetBasis',
 'TrilinearForm',
 '__version__',
 '_i4',
 '_i5',
 '_i6',
 'adaptive_theta',
 'asm',
 'build_pc_diag',
 'build_pc_ilu',
 'condense',
 'enforce',
 'penalize',
 'project',
 'projection',
 'solve',
 'solver_direct_scipy',
 'solver_eigen_scipy',
 'solver_eigen_scipy_sym',
 'solver_iter_krylov',
 'solver_iter_pcg'}
```

gatling-nrl:
Looks like some extraneous stuff has leaked through, so you were indeed restricting the imports with `__all__`. But I think I can get the import list to match without putting the `__all__` back in.
```
{'AbstractBasis',
 'Any',
 'BOUNDARY_ELEMENT_MAP',
 'Basis',
 'BilinearForm',
 'BoundaryFacetBasis',
 'Callable',
 'CellBasis',
 'CondensedSystem',
 'Dict',
 'DiscreteField',
 'Dofs',
 'DofsCollection',
 'DofsView',
 'EigenSolver',
 'Element',
 'ElementComposite',
 'ElementDG',
 'ElementGlobal',
 'ElementH1',
 'ElementHcurl',
 'ElementHdiv',
 'ElementHex0',
 'ElementHex1',
 'ElementHex1DG',
 'ElementHex2',
 'ElementHexDG',
 'ElementHexS2',
 'ElementLineHermite',
 'ElementLineMini',
 'ElementLineP0',
 'ElementLineP1',
 'ElementLineP1DG',
 'ElementLineP2',
 'ElementLinePp',
 'ElementQuad0',
 'ElementQuad1',
 'ElementQuad1DG',
 'ElementQuad2',
 'ElementQuadBFS',
 'ElementQuadDG',
 'ElementQuadP',
 'ElementQuadRT0',
 'ElementQuadS2',
 'ElementTetCCR',
 'ElementTetCR',
 'ElementTetDG',
 'ElementTetMini',
 'ElementTetN0',
 'ElementTetP0',
 'ElementTetP1',
 'ElementTetP2',
 'ElementTetRT0',
 'ElementTri15ParamPlate',
 'ElementTriArgyris',
 'ElementTriBDM1',
 'ElementTriCCR',
 'ElementTriCR',
 'ElementTriDG',
 'ElementTriHermite',
 'ElementTriMini',
 'ElementTriMorley',
 'ElementTriP0',
 'ElementTriP1',
 'ElementTriP1DG',
 'ElementTriP2',
 'ElementTriP3',
 'ElementTriP4',
 'ElementTriRT0',
 'ElementTriSkeletonP0',
 'ElementTriSkeletonP1',
 'ElementVector',
 'ElementVectorH1',
 'ElementWedge1',
 'ExteriorFacetBasis',
 'FacetBasis',
 'Form',
 'Functional',
 'InteriorBasis',
 'InteriorFacetBasis',
 'LinearForm',
 'LinearSolver',
 'LinearSystem',
 'Mapping',
 'MappingAffine',
 'MappingIsoparametric',
 'MappingMortar',
 'Mesh',
 'Mesh2D',
 'Mesh3D',
 'MeshHex',
 'MeshHex1',
 'MeshHex1DG',
 'MeshHex2',
 'MeshLine',
 'MeshLine1',
 'MeshLine1DG',
 'MeshLineConstructor',
 'MeshQuad',
 'MeshQuad1',
 'MeshQuad1DG',
 'MeshQuad2',
 'MeshTet',
 'MeshTet1',
 'MeshTet2',
 'MeshTri',
 'MeshTri1',
 'MeshTri1DG',
 'MeshTri2',
 'MeshWedge1',
 'MortarFacetBasis',
 'Optional',
 'Solution',
 'SubdomainFacetBasis',
 'TrilinearForm',
 'Tuple',
 'Type',
 'Union',
 'abstract_basis',
 'adaptive_theta',
 'asm',
 'assembly',
 'basis',
 'boundary_facet_basis',
 'build_pc_diag',
 'build_pc_ilu',
 'cell_basis',
 'condense',
 'discrete_field',
 'dofs',
 'element',
 'element_composite',
 'element_dg',
 'element_global',
 'element_h1',
 'element_hcurl',
 'element_hdiv',
 'element_hex',
 'element_line',
 'element_quad',
 'element_tet',
 'element_tri',
 'element_vector',
 'element_wedge_1',
 'enforce',
 'form',
 'g1',
 'generic_utils',
 'interior_facet_basis',
 'logger',
 'logging',
 'mapping',
 'mapping_affine',
 'mapping_isoparametric',
 'mapping_mortar',
 'mesh',
 'mesh_2d',
 'mesh_2d_2',
 'mesh_3d',
 'mesh_dg',
 'mesh_hex_1',
 'mesh_hex_1_dg',
 'mesh_hex_2',
 'mesh_line_1',
 'mesh_line_1_dg',
 'mesh_quad_1',
 'mesh_quad_1_dg',
 'mesh_quad_2',
 'mesh_simplex',
 'mesh_tet_1',
 'mesh_tet_2',
 'mesh_tri_1',
 'mesh_tri_1_dg',
 'mesh_tri_2',
 'mesh_wedge_1',
 'mortar_facet_basis',
 'ndarray',
 'np',
 'penalize',
 'product',
 'project',
 'projection',
 'quadrature',
 'rcm',
 'refdom',
 'solve',
 'solve_eigen',
 'solve_linear',
 'solver_direct_scipy',
 'solver_eigen_scipy',
 'solver_eigen_scipy_sym',
 'solver_iter_krylov',
 'solver_iter_pcg',
 'sp',
 'spg',
 'spl',
 'spmatrix',
 'subdomain_facet_basis',
 'utils',
 'warn'}
```

gatling-nrl:
@kinnala  is the the list you posted exactly what you want? Does it have extra or missing things?

kinnala:
> @kinnala is the the list you posted exactly what you want? Does it have extra or missing things?

Yes. Nothing important is missing AFAIK.

kinnala:
What was the reason for getting rid of `__all__` again? Less LOC? I think we need to have a good control of the stuff exported via `from skfem import *`.

gatling-nrl:
Less places to update when you add something, was my motivation. Currently there are 3 or 4 places in several files that have to get updated if you want to add a symbol. 

If I can't easily get it to match your exported symbols, I'll revert it.

kinnala:
> Less places to update when you add something, was my motivation.

However, it's quite rarely that we need to add new symbols to the wildcard import. Most often it is `Element*` that I've added to the list. I personally don't prefer one way or another.

gatling-nrl:
@kinnala there's a lot going on in some of the `__init__.py` files some I'm just going to revert it to `all`. It's too nuisancy.

gatling-nrl:
@kinnala All the init stuff is reverted, but compared to your exported symbol table, this is what I have now
```
extra: SubdomainFacetBasis
extra: g1
missing: _i4
missing: _i5
missing: _i6
```
Aside from the first thing, obviously, I don't know what that other stuff is. Anyway the exported symbols from master match the ones from this PR.

@kinnala and @gdmcbain I'm happy with `SubdomainFacetBasis` now. What are your thoughts what else needs to be done before merging to master?

gatling-nrl:
@kinnala ok, that was easier than I thought. There was just some lingering confusion about how tind worked and a missing minus sign. I'm happy now.

kinnala:
Fantastic. I'll take a quick look again and then merge if it looks OK.

skfem/assembly/basis/subdomain_facet_basis.py
gdmcbain: Black, which has been discussed a couple of times in scikit-fem, is O. K. to run if it has been adopted across the project, but it hasn't.  I do still use it here, but in a restricted way, for new code, like so:
* Write my additions without any regard for format.
* Run `black` on the modified file.
* Copy my addition from the blackened file.
* Undo all the edits; e.g. `git checkout $THE_FILE`
* Paste my blackened addition into the otherwise original file.

kinnala:
I'm sorry about slowing this down. I'm really busy this and next week because of a job application and a paper revision that are both due by the end of next week. I expect to have this PR sorted out by 23rd of January.

gatling-nrl:
Thanks for the update, and no worries. I'm synced to #851 in my project and its working well, so I'm not in a rush. And I appreciate any feedback you have to make `SubdomainFacetBasis` more useful/correct. Also, you might want especially to think about how I moved `normalize_elements/facets` to `Mesh`. I added some comments here and there, and please make sure they're not wrong. I may add a few more of them to the PR since I've got some time before you look it over.

skfem/assembly/basis/subdomain_facet_basis.py
```diff
@@ -0,0 +1,80 @@
...
+        # Used by .with_element()
+        self._elements = elements
+        self._side = side
+
+        elements = mesh.normalize_elements(elements)
+        # Get indices of all facets in the mesh
+        all_facets, all_counts = np.unique(mesh.t2f,
+                                           return_counts=True)
+        # Find the exterior ones
+        exterior_facets = all_facets[all_counts == 1]
```
kinnala, line 33: Isn't this same as `mesh.boundary_facets()`?

kinnala: If yes, you might want to change because something similar to the `np.unique` call is done already there.

skfem/assembly/basis/subdomain_facet_basis.py
```diff
@@ -0,0 +1,80 @@
...
+        # Get indices of all facets in the subdomain
+        subdomain_facets, sub_counts = np.unique(mesh.t2f[:, elements],
+                                                 return_counts=True)
+        # Don't want facets on the exterior boundary, even if they are on the
+        # subdomain boundary
```
kinnala, line 38: What is the rationale behind silently dropping exterior facets vs. raising an error? Can you see an use case where this behavior is useful?

skfem/assembly/basis/subdomain_facet_basis.py
```diff
@@ -0,0 +1,80 @@
...
+        # Find the indices of the triangles for these facets (note some of
+        # these triangles won't be in the subdomain.)
+        tind = mesh.f2t[:, facets].flatten('F')
+        # Boolean indicator array for which triangles are in the subdomain
+        ix = np.isin(tind, elements)
+        # Get triangles to use for this basis based on `side` parameter
+        if self._side == 0:
+            # outward normals orginate from inside triangles
+            # but use the outside value for the integral
+            _tind_normals = tind[ix]
```
kinnala, line 54: Is this now correct? By default, the normal should point outwards from the subdomain to be consistent with `BoundaryFacetBasis`.

kinnala:
I'm now thinking about the comment in #854 by @gdmcbain. There are also other cases where you'd want to specify the orientation of the normal and I wonder if we should actually generalize this PR so that you can orient normals on arbitrary curves inside the domain. I'll make a version of this PR which tries to address this.

gdmcbain:
Were facet-bases optionally orientable on construction in various convenient and flexible ways, would that obviate the various subclasses? How else do they differ? I confess I've become lost in their multiplicity. Is orientation something that might reasonably be specified alongside the subset of facets, either explicitly (array of signs) or by intention (like `intorder` instead of`quadrature`). Perhaps the current subclasses could be replaced by constructor classmethods.

kinnala:
Yes, there are two things: orientation of normals and the direction of the trace. These are both related to Mesh.f2t but could be freely changed facet by facet. One could even flip values on a specific column of Mesh.f2t to change both of these at the same time. InteriorFacetBasis has traditionally used unique (but random) normal direction for both values of 'side'. This is convenient (but not necessary) for implementing jumps. Parameter 'side' has changed onlt the direction of the trace. Additional constructors for InteriorFacetBasis are an interesting idea worth looking at.

skfem/assembly/basis/subdomain_facet_basis.py
gatling-nrl: The rational was that you already have boundary facet basis, so it is possible to use 2 bases in a list to `asm` and include the exterior basis. The inverse of this is not as easy... if subdomain basis automatically _includes_ the exterior boundary, I didn't see an easy way to remove it.

In my E&M work, I do not expect to ever want the external boundary mixed with an interior boundary.

skfem/assembly/basis/subdomain_facet_basis.py
gatling-nrl: Yes, default (`side==0`) normals point out and use the gradient outside. with `side==1` normals point in and use gradient inside. 

There is not a provision for the other two combinations right now.

gdmcbain:
> there are two things: orientation of normals and the direction of the trace

I see, yes, of course. The normals will only differ in sign but the traces can differ arbitrarily.
   In the 2012 work, i was not using DG but was using two P1 (or similar) fields and was using one for each side. The current ex26 includes one such field; it has values throughout the domain but only those in the core are meaningful.  There would be another such field for the annulus, thus enabling two values at each point of the interface, i.e. different traces depending on the side.

PR #863, opened 2022-01-13
restrict ex29 to MINI

gdmcbain:
This follows a weird-looking irrelevant change to the example in the review of #851.

When I first developed the example, I wasn't sure what element might be best to use for velocity but with experience `ElementLineMini` is a good choice.  This is also the obvious example in which one might want to use this element.

### DIFF
--- a/docs/examples/ex24.py
+++ b/docs/examples/ex24.py
@@ -29,7 +29,7 @@ element = {'u': ElementVector(ElementTriP2()),
 basis = {variable: Basis(mesh, e, intorder=3)
          for variable, e in element.items()}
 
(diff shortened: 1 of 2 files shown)


gatling-nrl:
I forgot about that change. When I was tampering with the `__all__` statements in the imports, that import line in 29 failed. There doesn't seem to be any difference between the changed line and the old line...

And I didn't fully understand why the old line was failing. But it was the only example file that had this kind of import, and since just re-writing it to be more normal looking fixed it, I didn't dig any deeper.

Then when I reverted the `__all__` stuff, EX29 was still passing with the new line, which is why I forgot about it.

PR #874, opened 2022-02-02
serialize oriented boundaries

gdmcbain:
Preliminary push to enable linking to code from discussion #873.


gdmcbain:
As of 5e17b75, this passes the test `test_oriented_saveload` that I found commented out in `test_basis`.

It's a fairly minor (and backwardly compatible) extension of the old #681, extending the powers of two used to encode the facets of a cell with those facets listed again when they're flipped; i.e. for a `MeshTri`, it's like 0, 1, 2, as is, then 0, 1, 2 flipped.

I think the remaining errors after 69fcaa4 are nasty typing things due to having a `Union` as the type of value of a boundary (`ndarray` and `OrientedBoundary`).

gdmcbain:
Perhaps `OrientedBoundary` counts as a subclass of `ndarray` so that the type doesn't have to be specified as a `Union`, just as an `ndarray`.

kinnala:
Great! I'll take a look.

kinnala:
I didn't know you could just put two rows into a single cell data entry. It's magical, I'll try it out tomorrow and merge.

gdmcbain:
> I didn't know you could just put two rows into a single cell data entry.

And not just two; already in #681 we effectively put `.reform.nfacets` and here we put twice that, but they're only Booleans, encoded as the bits of the binary representation of integers.

I think one can put three or nine rows in point or cell data, as is used to store vectors and second-order tensors, depending on the output format that meshio writes to, but no that's not used here, meshio only sees one int per cell.

gdmcbain:
I wonder whether that suggests an alternative encoding: a three-component vector for each cell where the first entry is one if the cell is the trace-side of a facet in the subset and if so then the second is the index of the cell on the other side and otherwise both get a sentinel (zero?) as does the third in all cases. I don't know whether that has any advantages but it should be possible.

gdmcbain:
I didn't know whether the present approach would work for `MeshLine1` but it does seem to, passing `test_oriented_saveload`. I think this is a case in which Gmsh's MSH encoding of facets as (*n* &minus; 1)-dimensional cells would fail; the order of the sole supporting point is unique and can't imply its orientation.

gdmcbain:
Perhaps a simpler way, assuming we only want facets that have traces from inside cells, which I think is correct. Then every oriented facet is just a specific facet of a specific cell and we just need to put `True` in the corresponding slot of a boolean array shaped like `.t2f` and bit-pack the column of facets for each cell, which is really what we have been doing already since #681. Let's try that before merging this.

gdmcbain:
The present PR works for an arbitrary `.ori` but what I'm wondering now is whether that's only meaningful for internal facets which separate two cells rather than boundary facets which only pertain to one cell and so only have one meaningful orientation (hence our not bothering for years to orient `.boundaries` since they were mostly strictly boundaries).

gdmcbain:
Closing pending investigating this promising alternative back in #873.

PR #875, opened 2022-02-07
serialize oriented boundaries

gdmcbain:
This succeeds #874 for #873 on the understanding that the `t2f`-based (rather than subset of indices to `.facets`) indexing of facets is intrinsically ordered.


gdmcbain:
This passes the previous test but is failing on the examples of `test_meshio_cycle_boundaries` (which aren't oriented, but the current scheme more or less thinks all facets are oriented).  The examples are based on 

https://github.com/kinnala/scikit-fem/blob/64da34c4445ec49eda163f369b54fbdf895ff673/tests/test_mesh.py#L462-L465

and

https://github.com/kinnala/scikit-fem/blob/64da34c4445ec49eda163f369b54fbdf895ff673/tests/test_mesh.py#L490

gdmcbain:
The four 'left' facets of the `MeshTri()` example, `[0, 1, 2, 4]` first get turned into five `[0, 2, 2, 4, 1]` because the diagonal facet, 2, is accidentally counted twice because as well as it correctly appearing in `.t2f[:, 1]` it also appears in `.t2f[:, 0]` and this column (i.e. cell) is included because it is backs facets 0 and 1.  The offending line is

https://github.com/kinnala/scikit-fem/blob/408c07b33f3c28eb8183f432d7bd3500b427de7a/skfem/mesh/mesh.py#L254

gdmcbain:
I hadn't considered that the cell on the outside of one facet might be on the inside of another facet; that can't happen for a consistently `OrientedBoundary`.  Perhaps unoriented boundaries do have to be treated more carefully, not just as boundaries with `.ori` all zero.

gdmcbain:
O. K., this is passing all tests now, particularly:
* `test_mesh.test_meshio_cycle_boundaries` and
* `test_basis.test_oriented_saveload`

It doesn't much distinguish between oriented and unoriented &ldquo;boundaries&rdquo;, assuming `ori` is all zero when encoding unoriented boundaries and then returning an unoriented boundary should all `ori` be inferred as zero on decoding.

gdmcbain:
Oops, I've blackened the whole file `mesh.py` somewhere along the way.

This is the encoding:
```python
     def encode_boundary(boundary: Union[ndarray, OrientedBoundary]) -> ndarray:
            """return an array with an int per cell encoding a 'boundary'
            i.e. a subset of (optionally oriented) facets.
            Although these are stored internally as arrays of indices of
            facets, with an optional array of 0/1 orientations, an alternative
            representation is as a boolean array of shape (self.nelements,
            self.refdom.nfacets) with entries `True` if that facet belongs to
            the boundary.  This array can be bit-packed to give a one-
            dimensional array of ints, one per cell.  This is convenient as it
            can be treated as `cell_data` by `skfem.io.meshio` and stored in
            many external formats (VTK, XDMF, ...).
            That is, the binary expansion of the int for a cell gives the bit-
            flags for whether the corresponding facets are in `boundary`.  This
            naturally implies the orientation, as the cell is on the inside of
            the facet.
            """
            b = (
                boundary
                if isinstance(boundary, OrientedBoundary)
                else OrientedBoundary(boundary, np.zeros_like(boundary))
            )
            t2f_mask = np.zeros_like(self.t2f)
            columns = self.f2t[(b.ori, b)]
            r, c = np.nonzero(self.t2f[:, columns] == b)
            t2f_mask[(r, columns[c])] = 1
            return (1 << np.arange(self.refdom.nfacets)) @ t2f_mask
```
and this is the decoding:
```python
               mask = (
                    (1 << np.arange(self.refdom.nfacets))[:, None] 
                    & data[0].astype(int)
                ).astype(bool)
                facets = np.sort(self.t2f[mask])
                cells = mask.nonzero()[1]
                ori = np.arange(2) @ (self.f2t[:, facets] == cells)
                boundaries[subnames[2]] = (
                    OrientedBoundary(facets, ori) if ori.any() else facets
                )
```

PR #890, opened 2022-04-11
Experimental autodiff support

kinnala:
Obstacle problem with penalty?
```python
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import *
import autograd.numpy as np


m = MeshTri().refined(5)
basis = Basis(m, ElementTriP1())

eps = 1e-3
f0 = -1.

def g(x):
    return np.sin(np.pi * x[0]) - 1.05


@NonlinearForm
def F(u, v, w):
    return (dot(grad(u), grad(v))
            + 1. / w.h ** 2 * (u[0] - g(w.x)) * (u[0] - g(w.x) <= 0) * v[0]
            - f0 * v[0])


x = basis.zeros()
D = basis.get_dofs()

for itr in range(50):

    J, f = F.assemble(x, basis)

    xprev = x.copy()
    x += 0.99 * solve(*condense(J, -f, D=D))

    res = np.linalg.norm(x - xprev)
    print(res)
    if res < 1e-8:
        break


basis.plot(x, shading='gouraud', colorbar=True).show()
```
Minimal surface problem?
```python
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import grad, dot
import numpy as np
import autograd.numpy as anp

m = MeshTri().refined(5)


@NonlinearForm(hessian=True)
def energy(u, _):
    return anp.sqrt(1. + dot(grad(u), grad(u)))


basis = Basis(m, ElementTriP1())

D = m.boundary_nodes()
x = np.sin(np.pi * m.p[0])

for itr in range(100):
    J, F = energy.assemble(x, basis)
    x_prev = x.copy()
    x += 0.7 * solve(*condense(J, -F, D=D))
    res = np.linalg.norm(x - x_prev)
    if res < 1e-8:
        break
    if __name__ == "__main__":
        print(res)

if __name__ == "__main__":
    from skfem.visuals.matplotlib import plot3, show
    plot3(m, x)
    show()
```
Lid driven cavity with Navier-Stokes?
```python
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import *
import numpy as np


m = MeshTri().refined(3)
basis = Basis(m, ElementVector(ElementTriP2()) * ElementTriP1())


@NonlinearForm
def F(u, p, v, q, w):
    return (ddot(grad(u), grad(v))
            - dot(mul(grad(u), u[0]), v[0])
            - 1e-2 * p[0] * q[0]
            - div(u) * q[0]
            - div(v) * p[0])


x = basis.zeros()
D = basis.get_dofs().all(['u^1^1', 'u^2^1'])
x[basis.get_dofs('top').all(['u^1^1'])] = 100.

for itr in range(50):

    J, f = F.assemble(x, basis)

    xprev = x.copy()
    x += solve(*condense(J, -f, D=D))

    res = np.linalg.norm(x - xprev)
    print(res)
    if res < 1e-8:
        break


(u, ubasis), (p, pbasis) = basis.split(x)

pbasis.plot(p).show()
ubasis.plot(u).show()
```


kinnala:
I tried adding the Hessian, here is a really stupid way of calculating the Poisson problem:
```python
from skfem import *
from skfem.autodiff import NonlinearFunctional
from skfem.autodiff.helpers import *
import jax.numpy as np


m = MeshTri().refined(5)
basis = Basis(m, ElementTriP1())

f0 = 1.


@NonlinearFunctional
def G(u, w):
    return .5 * dot(grad(u), grad(u)) - f0 * u[0]


x = basis.zeros()
D = basis.get_dofs()

for itr in range(50):

    J, f = G.assemble(x, basis)

    xprev = x.copy()
    x += solve(*condense(J, f, D=D))

    res = np.linalg.norm(x - xprev)
    print(res)
    if res < 1e-8:
        break


basis.plot(x, shading='gouraud', colorbar=True).show()
```

kinnala:
I believe this is some sort of prototype nonlinear elasticity:
```python
from skfem import *
from skfem.autodiff import NonlinearFunctional
from skfem.autodiff.helpers import *
import jax.numpy as np


m = MeshTri().init_tensor(
    np.linspace(0, 4, 30),
    np.linspace(0, 0.1, 5),
)
basis = Basis(m, ElementVector(ElementTriP1()))


@NonlinearFunctional
def G(u, w):
    du = grad(u)
    duT = transpose(du)
    eps = .5 * (du + duT) + mul(duT, du)
    sig = eps + eye(trace(eps), 2)
    return .5 * ddot(sig, eps)


x = basis.project(lambda x: np.array([-x[0] / 25., x[1] * 0.]))
D = basis.get_dofs(lambda x: np.isclose(x[0], 0) + np.isclose(x[0], 4))

for itr in range(50):

    J, f = G.assemble(x, basis)

    xprev = x.copy()
    x += 0.95 * solve(*condense(J, f, D=D))

    res = np.linalg.norm(x - xprev)
    print(res)
    if res < 1e-8:
        break


mdefo = m.translated(x[basis.nodal_dofs])
m.draw().show()
mdefo.draw().show()
```

skfem/autodiff/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+                                      "an argument.")
+
+        @BilinearForm
+        def DF(u, v, w):
+            F1 = lambda U: self.form(U, v, w)
+            return jvp(F1, (u0.astuple,), (u.astuple,))[1]
+
+        @LinearForm
+        def y(v, w):
+            return -self.form(u0.astuple, v, w)
```
bhaveshshrimali, line 27: Although for nonlinear problems one always ends up solving J * du = -f, it may be a good idea to keep this at the level of solving? And return `self.form()` instead of `-self.form()`

skfem/autodiff/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+        @BilinearForm
+        def HF(u, v, w):
+            F1 = lambda U: self.form(U, w)
+            F2 = lambda U: jvp(F1, (U,), (v.astuple,))[1]
+            return jvp(F2, (u0.astuple,), (u.astuple,))[1]
+
+        @LinearForm
+        def DF(v, w):
+            F1 = lambda U: self.form(U, w)
+            return -jvp(F1, (u0.astuple,), (v.astuple,))[1]
```
bhaveshshrimali, line 57: Same as above. Might be better to keep this at the level of solving and return just `jvp(...)` here?

bhaveshshrimali:
> Figure_1 Figure_1 I believe this is some sort of prototype nonlinear elasticity:
> 
> ```python
> from skfem import *
> from skfem.autodiff import NonlinearFunctional
> from skfem.autodiff.helpers import *
> import jax.numpy as np
> 
> 
> m = MeshTri().init_tensor(
>     np.linspace(0, 4, 30),
>     np.linspace(0, 0.1, 5),
> )
> basis = Basis(m, ElementVector(ElementTriP1()))
> 
> 
> @NonlinearFunctional
> def G(u, w):
>     du = grad(u)
>     duT = transpose(du)
>     eps = .5 * (du + duT) + mul(duT, du)
>     sig = eps + eye(trace(eps), 2)
>     return .5 * ddot(sig, eps)
> 
> 
> x = basis.project(lambda x: np.array([-x[0] / 25., x[1] * 0.]))
> D = basis.get_dofs(lambda x: np.isclose(x[0], 0) + np.isclose(x[0], 4))
> 
> for itr in range(50):
> 
>     J, f = G.assemble(x, basis)
> 
>     xprev = x.copy()
>     x += 0.95 * solve(*condense(J, f, D=D))
> 
>     res = np.linalg.norm(x - xprev)
>     print(res)
>     if res < 1e-8:
>         break
> 
> 
> mdefo = m.translated(x[basis.nodal_dofs])
> m.draw().show()
> mdefo.draw().show()
> ```

Awesome! This works perfectly, but if I try to invoke `jit` explicitly it ends up taking humongous time for me, i.e.

```python
@BilinearForm
        def DF(u, v, w):
            F1 = lambda U: self.form(U, v, w)
            return jvp(F1, (u0.astuple,), (u.astuple,))[1]
"""
instead of
@BilinearForm
        def DF(u, v, w):
            F1 = lambda U: self.form(U, v, w)
            return return jvp(jit(F1), (u0.astuple,), (u.astuple,))[1]
"""
and so on... (will have to check up on `jax.jvp` in more detail)
```

skfem/autodiff/__init__.py
kinnala: Sure, I think you have a better overview of this area and I trust your judgment.

kinnala:
I didn't yet try any optimizations, e.g., using `jit`. I'm not even sure how well this combines with JAX automatic differentiation functionality. The background of JAX seems to be that the autodiff functionality comes from Autograd (another older package) and then on top of it is the functionality to `jit` the code, but I think just-in-time compilation might only be related to running the code on GPU/TPU:

> With its updated version of [Autograd](https://github.com/hips/autograd), JAX can automatically differentiate native Python and NumPy code. ... What’s new is that JAX uses [XLA](https://www.tensorflow.org/xla) to compile and run your NumPy code on accelerators, like GPUs and TPUs. Compilation happens under the hood by default, with library calls getting just-in-time compiled and executed.

So I'm not sure if this JIT actually is relevant here as we are not attempting to do any GPU/TPU stuff here.

kinnala:
It seems that one should be able to install JAX with only autodiff/CPU functionality via `pip install jax[cpu]` and it should be pure Python. I'm trying out if that works in the test suite.

kinnala:
I suppose there is now everything that I'd like this kind of functionality to have. It's marked experimental and probably it can be improved performance wise a lot if someone actually understands how this stuff works (I don't). However, for the basic cases that I need it for it seems to be working pretty nicely.

I tried using this idea of doing forward AD over reverse AD by combining `jax.jvp` and `jax.vjp` but I couldn't make it to work. Apparently it should make it more effective. There should also be some sort of caching happening because these derivatives now get evaluated multiple times inside the forms for the same values.

kinnala:
@adtzlr You seem to be well versed in AD. Do you have any comments on this PR? Am I doing something which doesn't make sense?

kinnala:
I guess casADi is about differentiating through expressions defined using it's own syntax and not really arbitrary Python code. There could definitely be different ways of defining the forms supported by scikit-fem, e.g., symbolic stuff using sympy.

bhaveshshrimali:
> So I'm not sure if this JIT part is actually relevant here as we are not attempting to do any GPU/TPU stuff. I think the GPU/TPU stuff might be more relevant for applications with dense N-arrays such as deep learning.

Cool, I just wanted to test something like `ex43` or the one from #439 using this functionality -- hence thought of naively jitting the function, but that doesn't work. In its current state this takes a lot of time, maybe the problem is somewhere else since even without jitting it just stalls... 

```python
@NonlinearFunctional
def G(u, w):
    du = grad(u)
    F = identity(du) + du
    J = det(F)
    logJ = np.log(J)
    I1 = trace(F) 
    W = mu/2. * (I1 - 3.) - mu * logJ + lmbda/2. * (J - 1.)**2 
    return W#.5 * ddot(sig, eps)
```

where the `identity` helper is the same as before except now with `jax.numpy`

```
def identity(w, N=None):
    """Create identity matrix."""
    if N is None:
        if len(w.shape) > 2:
            N = w.shape[-3]
        else:
            raise ValueError("Cannot deduce the size of the identity matrix. "
                             "Give an explicit keyword argument N.")
    return eye(jnp.ones(w.shape[-2:]), N)
```

adtzlr:
> @adtzlr You seem to be well versed in AD. Do you have any comments on this PR? Am I doing something which doesn't make sense? Would it be better to use, e.g., casADi?

Hi @kinnala!

I think basically it doesn't matter which package you use. JAX is probably the best option if you're able to install ist but as I'm 99% on Windows I haven't worked with it yet. Personally I really like the older package Autograd because it is a minimal-dependancy package and can be installed everywhere - just like scikit-fem.

I only tried to implement the AD stuff for the constitutive part (material formulation in mechanics: strain energy function -> stress tensor -> elasticity tensor) and not on the full form like you did if I understand it correctly. But I like the idea, if AD, then consequently on the full form.

Recently I documented how to use Autograd on input data with trailing axes, if that is interesting for you (as a concept, without using my package matadi at all): https://github.com/adtzlr/matadi/discussions/11#discussioncomment-2554834

The two ML packages PyTorch and Tensorflow also have AD capabilities built in. A few comments above of the provided link there is also an example with PyTorch. A code using Tensorflow is also very similar. But for your idea; extending an existing code-base with NumPy it is easier to use Autograd or JAX. casADi is also fine but it behaves more like the sparse-package from scipy: only matrices with two dimensions are allowed. To sum up: I'd keep JAX if you've got it to work.

adtzlr:
> I guess casADi is about differentiating through expressions defined using it's own syntax and not really arbitrary Python code. There could definitely be different ways of defining the forms supported by scikit-fem, e.g., symbolic stuff using sympy.

Symbolic derivatives are super slow. I tried to create some simple derivatives, see e.g. https://adtzlr.github.io/blog/markdown/2021/10/12/SympyTensor.html - that worked really well but everything more complicated is a mess. Please correct me if I'm wrong. Also you have to find an efficient way to somehow broadcast the derivatives over the trailing axes of the input data.

skfem/autodiff/__init__.py
adtzlr: I agree on that.

skfem/experimental/autodiff/helpers.py
```diff
@@ -0,0 +1,42 @@
...
+    return jnp.einsum('ij...->ji...', T)
+
+
+def mul(A, B):
+    if len(A.shape) == len(B.shape):
+        return jnp.einsum('ij...,jk...->ik...', A, B)
+    return jnp.einsum('ij...,j...->i...', A, B)
+
+
+def trace(T):
```
adtzlr, line 36: Just a sidenote: the default NumPy function for the trace is fine here as the default axes are `0` and `1`, see https://numpy.org/doc/stable/reference/generated/numpy.trace.html.

```python
def trace(T):
    return jnp.trace(T, axis1=0, axis2=1)
```

bhaveshshrimali:
> Can you try installing jax[cpu] on Windows? They claim it's pure Python.

I tried it but it doesn't work, at least with the latest/stable release (which I think is `jax-0.3.5`).. here is a truncated output

```terminal
Collecting scipy>=1.2.1
  Downloading scipy-1.8.0-cp39-cp39-win_amd64.whl (36.9 MB)
     |████████████████████████████████| 36.9 MB 6.4 MB/s
Collecting typing_extensions
  Downloading typing_extensions-4.1.1-py3-none-any.whl (26 kB)
Collecting jax[cpu]
  Using cached jax-0.3.5-py3-none-any.whl
  Downloading jax-0.3.4.tar.gz (924 kB)
     |████████████████████████████████| 924 kB 6.4 MB/s
  Downloading jax-0.3.3.tar.gz (924 kB)
     |████████████████████████████████| 924 kB ...
  Downloading jax-0.3.2.tar.gz (926 kB)
     |████████████████████████████████| 926 kB ...
  Downloading jax-0.3.1.tar.gz (912 kB)
     |████████████████████████████████| 912 kB 6.8 MB/s
  Downloading jax-0.3.0.tar.gz (896 kB)
     |████████████████████████████████| 896 kB 6.4 MB/s
  Downloading jax-0.2.28.tar.gz (887 kB)
     |████████████████████████████████| 887 kB ...
  Downloading jax-0.2.27.tar.gz (873 kB)
     |████████████████████████████████| 873 kB ...
  Using cached jax-0.2.26-py3-none-any.whl
  Downloading jax-0.2.25.tar.gz (786 kB)
     |████████████████████████████████| 786 kB ...
  Downloading jax-0.2.24.tar.gz (786 kB)
     |████████████████████████████████| 786 kB 6.8 MB/s
  Downloading jax-0.2.22.tar.gz (776 kB)
     |████████████████████████████████| 776 kB 6.4 MB/s
  Downloading jax-0.2.21.tar.gz (756 kB)
     |████████████████████████████████| 756 kB 6.8 MB/s
  Downloading jax-0.2.20.tar.gz (745 kB)
     |████████████████████████████████| 745 kB 6.8 MB/s
  Downloading jax-0.2.19.tar.gz (728 kB)
     |████████████████████████████████| 728 kB ...
  Downloading jax-0.2.18.tar.gz (708 kB)
     |████████████████████████████████| 708 kB ...
  Downloading jax-0.2.17.tar.gz (693 kB)
     |████████████████████████████████| 693 kB ...
  Downloading jax-0.2.16.tar.gz (680 kB)
     |████████████████████████████████| 680 kB ...
  Downloading jax-0.2.15.tar.gz (681 kB)
     |████████████████████████████████| 681 kB 6.4 MB/s
  Downloading jax-0.2.14.tar.gz (669 kB)
     |████████████████████████████████| 669 kB 6.4 MB/s
  Downloading jax-0.2.13.tar.gz (630 kB)
     |████████████████████████████████| 630 kB ...
  Using cached jax-0.2.12-py3-none-any.whl
  Downloading jax-0.2.11.tar.gz (596 kB)
     |████████████████████████████████| 596 kB ...
  Downloading jax-0.2.10.tar.gz (589 kB)
     |████████████████████████████████| 589 kB 6.4 MB/s
WARNING: jax 0.2.10 does not provide the extra 'cpu'
Requirement already satisfied: six in c:\miniconda3\lib\site-packages (from absl-py->jax[cpu]) (1.16.0)
Building wheels for collected packages: jax
  Building wheel for jax (setup.py) ... done
  Created wheel for jax: filename=jax-0.2.10-py3-none-any.whl size=679775 sha256=8cb6dc8392f124e771bf25e6dafff38cefed5fe9bffbbd179be36518f1af8555
  Stored in directory: c:\users\bshri\appdata\local\pip\cache\wheels\0b\36\5e\c025acb1b9e65160c25f33c9370675402c8e23ed6f613401ea
Successfully built jax
Installing collected packages: numpy, opt-einsum, absl-py, jax
Successfully installed absl-py-1.0.0 jax-0.2.10 numpy-1.22.3 opt-einsum-3.3.0
```

The alternative that worked for me on windows was to use the pre-built wheels (https://github.com/kinnala/scikit-fem/discussions/714#discussioncomment-2548254)

adtzlr (commented):
Looks good! However I can't test it right now. I like the distinction between a `Functional` and a `Form`.

skfem/experimental/autodiff/helpers.py
adtzlr: The only thing I wonder: In my tests like the one from my posted link above (https://github.com/adtzlr/matadi/discussions/11#discussioncomment-2554834) I had to sum the result of a functional over the trailing axes before taking the jacobian. Otherwise I got a full jacobian of every integration point of every element w.r.t. each other element. And I only want to have it quadrature-point-element wise. As you're using the `jvp` that *problem* seems to be solved?

kinnala: I need to look into this more carefully. I'm not exactly sure why the current approach works and I just managed to get the correct result by trying out jvp. @bhaveshshrimali  reports some slowness so there might be optimizations.

adtzlr:
> It seems that one should be able to install JAX with only autodiff/CPU functionality via `pip install jax[cpu]` and it should be pure Python. I'm trying out if that works in the test suite. I also tried looking a way to combine Numba and JAX but couldn't find anything so far.

No, there are no official binary wheels for Windows on PyPI:

> We support installing or building jaxlib on Linux (Ubuntu 16.04 or later) and macOS (10.12 or later) platforms. Windows users can use JAX on CPU and GPU via the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about). There is some initial native Windows support, but since it is still somewhat immature, there are no binary releases and it must be [built from source](https://jax.readthedocs.io/en/latest/developer.html#additional-notes-for-building-jaxlib-from-source-on-windows).

I think JAX is not meant to be used with Numba, they use XLA for compilation instead.

adtzlr:
I adopted [Example 43](https://github.com/kinnala/scikit-fem/blob/master/docs/examples/ex43.py) and added a template code with the [autograd-package](https://github.com/HIPS/autograd). However, I didn't figure out how to use jacobian-vector or vector-jacobian producs directly. Hence, I had to create the full jacobians and do the matrix-vector-products by hand inside the forms.

<details>
<summary>Well, it is slow but it works. 🚀 </summary>

```python
r"""Hyperelasticity.

The strain energy density function per unit undeformed volume of the
isotropic hyperelastic Neo-Hookean material formulation is given as

..  math::

    \psi(\boldsymbol{F}) = \frac{\mu}{2} ( I_1 - 3) - \mu \ln(J) +
    \frac{\lambda}{2} \ln^2(J)


by the first invariant of the right Cauchy-Green deformation tensor
and the determinant of the deformation gradient (volume change).

..  math::

    I_1 &= \text{tr}(\boldsymbol{F}^T \boldsymbol{F}) =
    \boldsymbol{F} : \boldsymbol{F}

    J &= \det(\boldsymbol{F})

"""
import numpy as np

from skfem import *
from skfem.helpers import grad
from autograd import numpy as np, jacobian as jac

# helpers
transpose = lambda T: np.einsum("ij...->ji...", T)
ddot = lambda u, v: np.einsum("ij...,ij...->...", u, v)
mul = lambda u, v: np.einsum("ij...,jk...->ik...", u, v)
det = lambda T: np.linalg.det(T.T).T

def eye(w, size):
    return np.array([[w if i == j else 0. * w for i in range(size)]
                      for j in range(size)])

def identity(w, N=None):
    """Create identity matrix."""
    if N is None:
        if len(w.shape) > 2:
            N = w.shape[-3]
        else:
            raise ValueError("Cannot deduce the size of the identity matrix. "
                             "Give an explicit keyword argument N.")
    return eye(np.ones(w.shape[-2:]), N)

# jacobian function
def reduce(fun, argnum=0):
    "Function decorator for jacobians on inputs with trailing axes."
    
    def inner(*args, **kwargs):
        "Sum the result over the trailing axes."
        axes = tuple(-np.arange(1, len(args[argnum].shape) - 1))
        
        return np.sum(fun(*args, **kwargs), axes)
    
    return inner

def jacobian(fun, argnum=0):
    "Apply `jacobian` from autograd on a `reduce`-decorated function."
    return jac(reduce(fun, argnum=argnum), argnum=argnum)

# note: rough mesh to make tests fast
mesh = MeshHex.init_tensor(
    np.linspace(0, 1, 10),
    np.linspace(-.1, .1, 3),
    np.linspace(-.1, .1, 3),
).with_boundaries({
    'left': lambda x: x[0] == 0.,
    'right': lambda x: x[0] == 1.,
})
element = ElementVector(ElementHex1())
basis = Basis(mesh, element, intorder=1)

def deformation_gradient(w):
    dudX = grad(w["displacement"])
    F = dudX + identity(dudX)
    return F

def W(F, mu=1.0, lmbda=2.0):
    r"""The strain energy density function per unit undeformed volume of the
    isotropic hyperelastic Neo-Hookean material formulation is given as

    ..  math::

        \psi(\boldsymbol{F}) = \frac{\mu}{2} ( I_1 - 3) - \mu \ln(J) +
        \frac{\lambda}{2} \ln^2(J)
        
    """
    
    I1 = ddot(F, F)
    J = det(F)
    lnJ = np.log(J)
    
    return mu / 2 * (I1 - 3) - mu * lnJ + lmbda * lnJ ** 2 / 2

@LinearForm
def L(v, w):
    F = deformation_gradient(w)
    dF = grad(v)
    return ddot(jacobian(W)(F), dF)

@BilinearForm
def a(u, v, w):
    F = deformation_gradient(w)
    DF = grad(u)
    dF = grad(v)
    return ddot(dF, ddot(jacobian(jacobian(W))(F), DF))

u  = basis.zeros()
du = basis.zeros()
uv = basis.interpolate(u)

dofs = basis.get_dofs('right')
dirichlet = basis.get_dofs({'right', 'left'})

f = L.assemble(basis, displacement=uv)
K = a.assemble(basis, displacement=uv)

right = -0.1
tol = 1e-10
nsteps = 7

for step in range(nsteps):
    for iteration in range(10):
        c = (step + 1.) / nsteps

        du_D = u.copy()
        x, y, z = basis.doflocs[:, dofs.nodal['u^1']]
        du_D[dofs.nodal['u^1']] = c * right - du_D[dofs.nodal['u^1']]
        du_D[dofs.nodal['u^2']] = (
            y * np.cos(c * np.pi) - z * np.sin(c * np.pi) - y
            - du_D[dofs.nodal['u^2']]
        )
        du_D[dofs.nodal['u^3']] = (
            y * np.sin(c * np.pi) + z * np.cos(c * np.pi) - z
            - du_D[dofs.nodal['u^3']]
        )

        du = solve(*condense(K, -f, x=du_D, D=dirichlet))
        norm_du = np.linalg.norm(du)
        u += du

        uv = basis.interpolate(u)

        f = L.assemble(basis, displacement=uv)
        K = a.assemble(basis, displacement=uv)

        print(1 + iteration, norm_du)

        if norm_du < tol:
            break


if __name__ == '__main__':
    (mesh.translated(u[basis.nodal_dofs])
         .draw('vedo', point_data={'uy': u[basis.nodal_dofs[1]]})
         .show())
```

</details>

kinnala:
I made some benchmarks and it seems that neither the number of quadrature points or the number of elements has a large effect on the assembly time, but the number of basis functions has. And for `ElementVector(ElementHex1())` this is already 24 and the linearization is performed within the loop 24 x 24 times and that's where the slowness comes from. Two dimensional examples run quite quick because they have much less linearizations.

kinnala:
I was able to make it approximately one magnitude faster by caching some of those derivatives within the basis function loop. Some generality is lost but let's see if I can recover that later.

kinnala:
Here is one example which takes around 4 secs per assembly on my laptop:
```python
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import *
import jax.numpy as np
import time


m = MeshHex().refined(3)
basis = Basis(m, ElementVector(ElementHex1()), intorder=1)
mu, lmbda = 1., 2.


@NonlinearForm
def G(u, w):
    du = grad(u)
    F = eye(np.ones(du.shape[-2:]), 3) + du
    J = det(F)
    lnJ = np.log(J)
    I1 = ddot(F, F)
    return mu / 2. * (I1 - 3.) - mu * lnJ + lmbda / 2. * lnJ ** 2


x = basis.project(lambda x: np.array([x[0] / 10., x[1] * 0., x[2] * 0.]))
D = basis.get_dofs(lambda x: np.isclose(x[0], 0) + np.isclose(x[0], 1))

for itr in range(50):

    start = time.time()
    J, f = G.assemble(x, basis)
    stop = time.time()
    print('Assembly time: {} seconds'.format(stop - start))

    xprev = x.copy()
    x += solve(*condense(J, -f, D=D))

    res = np.linalg.norm(x - xprev)
    print('Residual: {}'.format(res))
    if res < 1e-8:
        break


mdefo = m.translated(x[basis.nodal_dofs])
m.draw().show()
mdefo.draw().show()
```

adtzlr:
> Here is one example which takes around 4 secs per assembly on my laptop


<details>
<summary>Out of interest, I re-wrote your example using autograd. Assembly time ~2.7s on my laptop.</summary>

```python
from skfem import *
from skfem.helpers import grad, eye
import time

import numpy as np
from autograd import numpy as anp, make_hvp


m = MeshHex().refined(3)
basis = Basis(m, ElementVector(ElementHex1()), intorder=1)
mu, lmbda = 1., 2.

# autograd helpers
transpose = lambda T: anp.einsum("ij...->ji...", T)
ddot = lambda u, v: anp.einsum("ij...,ij...->...", u, v)
mul = lambda u, v: anp.einsum("ij...,jk...->ik...", u, v)
det = lambda T: anp.linalg.det(T.T).T


def functional(F):
    J = det(F)
    lnJ = anp.log(J)
    I1 = ddot(F, F)
    return anp.sum(mu / 2. * (I1 - 3.) - mu * lnJ + lmbda / 2. * lnJ ** 2)


@BilinearForm
def a(u, v, w):
    dF = grad(w['prev'])
    dv = grad(v)
    du = grad(u)
    F = eye(np.ones(dv.shape[-2:]), 3) + dF
    return ddot(make_hvp(functional)(F)[0](du), dv)


@LinearForm
def L(v, w):
    dF = grad(w['prev'])
    dv = grad(v)
    F = eye(np.ones(dv.shape[-2:]), 3) + dF
    return ddot(make_hvp(functional)(F)[1], dv)


x = basis.project(lambda x: np.array([x[0] / 10., x[1] * 0., x[2] * 0.]))
D = basis.get_dofs(lambda x: np.isclose(x[0], 0) + np.isclose(x[0], 1))

for itr in range(50):

    w = basis.interpolate(x)
    start = time.time()
    J = asm(a, basis, prev=w)
    f = asm(L, basis, prev=w)
    stop = time.time()
    print('Assembly time: {} seconds'.format(stop - start))

    xprev = x.copy()
    x += solve(*condense(J, -f, D=D))

    res = np.linalg.norm(x - xprev)
    print('Residual: {}'.format(res))
    if res < 1e-8:
        break


mdefo = m.translated(x[basis.nodal_dofs])
m.draw().show()
mdefo.draw().show()
```

</details>

bhaveshshrimali:
> this is already 24 and the linearization is performed within the loop 24 x 24 times and that's where the slowness comes from.

I think I may have asked this before, but could you explain why we need to loop for a full 24 x 24 matrix?  Its been a while since I looked at the vectorized assembly procedure  but my hunch was that you could only loop for the relevant indices for each component. (for `u_x` -- 8, for `u_y` --8 and for `u_z` -- 8) or is this what it is doing? For `u`-x you wouldn't need to loop over the indices for `u_y` and `u_z` and so on.. where `u_i`'s are essentially the local basis functions for the `i-th` component of the field..

kinnala:
Those different directions affect each other, right? There might be other, more solid mechanics specific ways of assembling. However, here the thinking is that if you have ElementHex1() with phi_1, phi_2, ..., phi_8 and you make it ElementVector(ElementHex1()) then you have the local basis (phi_1, 0, 0), (phi_2, 0, 0), ..., (phi_8, 0, 0), (0, phi_1, 0), ..., (0, phi_8, 0), (0, 0, phi_1), ..., (0, 0, phi_8) which is 24 basis functions in total and, in general, all have nontrivial energy inner products. This means the local matrices are 24 x 24 matrices with nontrivial entries.

kinnala:
I've thought about allowing the specification of some structure, e.g., symmetry which would get rid of about half of the form evaluations for larger bases. Haven't gotten to it yet since I don't do vectorial 3D problems too often.

bhaveshshrimali:
> Those different directions affect each other, right? There might be other, more solid mechanics specific ways of assembling. However, here the thinking is that if you have ElementHex1() with phi_1, phi_2, ..., phi_8 and you make it ElementVector(ElementHex1()) then you have the local basis (phi_1, 0, 0), (phi_2, 0, 0), ..., (phi_8, 0, 0), (0, phi_1, 0), ..., (0, phi_8, 0), (0, 0, phi_1), ..., (0, 0, phi_8) which is 24 basis functions in total and, in general, all have nontrivial energy inner products. This means the local matrices are 24 x 24 matrices with nontrivial entries.

Hmm.. I think you are right. I was thinkining only about the gradients of the local-basis function `d(phi_1)/dx, ... ` and so on (that would have a banded structure , say if you have `cell_dofs = [cell_dofs_x, cell_dofs_y, cell_dofs_z]` ) but not the energy inner product with the hessian..

bhaveshshrimali:
> Seems that Autograd and JAX perform nearly equal if no jit-compilation is used.

@adtzlr I did a quick check with both (Autograd version 1.4 and jax) on my windows laptop and it turns out that `autograd` is slightly faster. But with the changes in `skfem.experimental.autodiff` especially now with the `NonlinearForm` I am at least able to run ex43 without any problems

<details>
<summary>Autograd</summary>

```python
import numpy as np

from skfem import *
from skfem.helpers import grad, identity
from autograd import numpy as anp, jacobian as jac, make_hvp

# helpers
ddot = lambda u, v: anp.einsum("ij...,ij...->...", u, v)
det = lambda T: anp.linalg.det(T.T).T

# note: rough mesh to make tests fast
mesh = MeshHex.init_tensor(
    np.linspace(0, 1, 10),
    np.linspace(-.1, .1, 3),
    np.linspace(-.1, .1, 3),
).refined().with_boundaries({
    'left': lambda x: x[0] == 0.,
    'right': lambda x: x[0] == 1.,
})
element = ElementVector(ElementHex1())
basis = Basis(mesh, element, intorder=1)

def deformation_gradient(w):
    dudX = grad(w["displacement"])
    F = dudX + identity(dudX)
    return F

def W(F, mu=1.0, lmbda=2.0):
    r"""The strain energy density function per unit undeformed volume of the
    isotropic hyperelastic Neo-Hookean material formulation is given as

    ..  math::

        \psi(\boldsymbol{F}) = \frac{\mu}{2} ( I_1 - 3) - \mu \ln(J) +
        \frac{\lambda}{2} \ln^2(J)
        
    """
    
    I1 = ddot(F, F)
    J = det(F)
    lnJ = anp.log(J)
    
    return anp.sum(mu / 2 * (I1 - 3) - mu * lnJ + lmbda * lnJ ** 2 / 2)

@LinearForm
def L(v, w):
    F = deformation_gradient(w)
    dF = grad(v)
    return ddot(make_hvp(W)(F)[1], dF)

@BilinearForm
def a(u, v, w):
    F = deformation_gradient(w)
    DF = grad(u)
    dF = grad(v)
    return ddot(make_hvp(W)(F)[0](DF), dF)

u  = basis.zeros()
du = basis.zeros()
uv = basis.interpolate(u)

dofs = basis.get_dofs('right')
dirichlet = basis.get_dofs({'right', 'left'})

f = L.assemble(basis, displacement=uv)
K = a.assemble(basis, displacement=uv)

right = -0.1
tol = 1e-10
nsteps = 7

from time import time

times = []

for step in range(nsteps):
    print(f"Step: {step+1:d}")
    for iteration in range(10):
        c = (step + 1.) / nsteps

        du_D = u.copy()
        x, y, z = basis.doflocs[:, dofs.nodal['u^1']]
        du_D[dofs.nodal['u^1']] = c * right - du_D[dofs.nodal['u^1']]
        du_D[dofs.nodal['u^2']] = (
            y * np.cos(c * np.pi) - z * np.sin(c * np.pi) - y
            - du_D[dofs.nodal['u^2']]
        )
        du_D[dofs.nodal['u^3']] = (
            y * np.sin(c * np.pi) + z * np.cos(c * np.pi) - z
            - du_D[dofs.nodal['u^3']]
        )

        du = solve(*condense(K, -f, x=du_D, D=dirichlet))
        norm_du = np.linalg.norm(du)
        u += du

        uv = basis.interpolate(u)
        t1 = time()
        f = L.assemble(basis, displacement=uv)
        K = a.assemble(basis, displacement=uv)
        times.append(time() - t1)
        norm_df = np.linalg.norm(f[basis.complement_dofs(dirichlet)])
        
        # print(1 + iteration, norm_du)
        print(f"Newton iteration: {iteration+1:d} | du: {norm_du:.3e} | df: {norm_df:.3e}")

        if norm_df < tol:
            break

mesh.save('disps_auto.xdmf', point_data={'disps': u[basis.nodal_dofs].T})
print(f"Assemble time average: {np.mean(times):.3e}")

```
</details>


<details>
<summary>Jax</summary>

```python

from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import *
import jax.numpy as np
import time
import jax
import numpy as onp
jax.config.update("jax_enable_x64", True)
jax.config.update("jax_platform_name", "cpu")


# note: rough mesh to make tests fast
m = MeshHex.init_tensor(
    np.linspace(0, 1, 10),
    np.linspace(-.1, .1, 3),
    np.linspace(-.1, .1, 3),
).refined().with_boundaries({
    'left': lambda x: x[0] == 0.,
    'right': lambda x: x[0] == 1.,
})
# element = ElementVector(ElementHex1())
# basis = Basis(mesh, element, intorder=1)
basis = Basis(m, ElementVector(ElementHex1()), intorder=1)
mu, lmbda = 1., 2.


@NonlinearForm(nthreads=2)
def G(u, w):
    du = grad(u)
    F = eye(np.ones(du.shape[-2:]), 3) + du
    J = det(F)
    I1 = ddot(F, F)
    return mu / 2. * (I1 - 3.) - mu * (J-1) + lmbda / 2. * (J-1) ** 2


u  = basis.zeros()
du = basis.zeros()
uv = basis.interpolate(u)

dofs = basis.get_dofs('right')
dirichlet = basis.get_dofs({'right', 'left'})

J, f = G.assemble(u, basis)

right = -0.1
tol = 1e-10
nsteps = 7
from time import time
times = []

for step in range(nsteps):
    print(f"Step: {step+1:d}")
    for iteration in range(10):
        c = (step + 1.) / nsteps

        du_D = u.copy()
        x, y, z = basis.doflocs[:, dofs.nodal['u^1']]
        du_D[dofs.nodal['u^1']] = c * right - du_D[dofs.nodal['u^1']]
        du_D[dofs.nodal['u^2']] = (
            y * np.cos(c * np.pi) - z * np.sin(c * np.pi) - y
            - du_D[dofs.nodal['u^2']]
        )
        du_D[dofs.nodal['u^3']] = (
            y * np.sin(c * np.pi) + z * np.cos(c * np.pi) - z
            - du_D[dofs.nodal['u^3']]
        )

        du = solve(*condense(J, -f, x=du_D, D=dirichlet))
        norm_du = np.linalg.norm(du)
        u += du

        uv = basis.interpolate(u)
        t1 = time()
        J, f = G.assemble(u, basis)
        times.append(time() - t1)
        norm_df = np.linalg.norm(f[basis.complement_dofs(dirichlet)])
        # print(1 + iteration, norm_df)
        print(f"Newton iteration: {iteration+1:d} | du: {norm_du:.3e} | df: {norm_df:.3e}")
        if norm_df < tol:
            break

m.save('disps.xdmf', point_data={'disps': u[basis.nodal_dofs].T})
print(f"Assemble time average: {onp.mean(times):.3e}")

```
</details>

<details>
<summary>Autograd/Jax average assembly time (& residuals)</summary>

```log
Step: 1
Newton iteration: 1 | du: 5.974e-01 | df: 9.542e-03
Newton iteration: 2 | du: 9.244e-02 | df: 7.066e-04
Newton iteration: 3 | du: 4.438e-02 | df: 7.479e-05
Newton iteration: 4 | du: 1.391e-02 | df: 7.373e-06
Newton iteration: 5 | du: 7.690e-04 | df: 2.213e-08
Newton iteration: 6 | du: 4.728e-06 | df: 8.417e-13
Step: 2
Newton iteration: 1 | du: 5.975e-01 | df: 9.471e-03
Newton iteration: 2 | du: 9.207e-02 | df: 6.938e-04
Newton iteration: 3 | du: 4.641e-02 | df: 8.112e-05
Newton iteration: 4 | du: 1.418e-02 | df: 7.578e-06
Newton iteration: 5 | du: 8.527e-04 | df: 2.695e-08
Newton iteration: 6 | du: 5.380e-06 | df: 1.078e-12
Step: 3
Newton iteration: 1 | du: 5.975e-01 | df: 9.401e-03
Newton iteration: 2 | du: 9.171e-02 | df: 6.811e-04
Newton iteration: 3 | du: 4.835e-02 | df: 8.733e-05
Newton iteration: 4 | du: 1.439e-02 | df: 7.723e-06
Newton iteration: 5 | du: 9.331e-04 | df: 3.195e-08
Newton iteration: 6 | du: 5.984e-06 | df: 1.320e-12
Step: 4
Newton iteration: 1 | du: 5.974e-01 | df: 9.332e-03
Newton iteration: 2 | du: 9.135e-02 | df: 6.686e-04
Newton iteration: 3 | du: 5.023e-02 | df: 9.341e-05
Newton iteration: 4 | du: 1.454e-02 | df: 7.806e-06
Newton iteration: 5 | du: 1.008e-03 | df: 3.695e-08
Newton iteration: 6 | du: 6.511e-06 | df: 1.546e-12
Step: 5
Newton iteration: 1 | du: 5.973e-01 | df: 9.264e-03
Newton iteration: 2 | du: 9.098e-02 | df: 6.560e-04
Newton iteration: 3 | du: 5.202e-02 | df: 9.933e-05
Newton iteration: 4 | du: 1.464e-02 | df: 7.826e-06
Newton iteration: 5 | du: 1.077e-03 | df: 4.177e-08
Newton iteration: 6 | du: 6.940e-06 | df: 1.738e-12
Step: 6
Newton iteration: 1 | du: 5.971e-01 | df: 9.195e-03
Newton iteration: 2 | du: 9.059e-02 | df: 6.434e-04
Newton iteration: 3 | du: 5.376e-02 | df: 1.051e-04
Newton iteration: 4 | du: 1.468e-02 | df: 7.790e-06
Newton iteration: 5 | du: 1.140e-03 | df: 4.630e-08
Newton iteration: 6 | du: 7.263e-06 | df: 1.884e-12
Step: 7
Newton iteration: 1 | du: 5.967e-01 | df: 9.126e-03
Newton iteration: 2 | du: 9.018e-02 | df: 6.306e-04
Newton iteration: 3 | du: 5.545e-02 | df: 1.109e-04
Newton iteration: 4 | du: 1.468e-02 | df: 7.705e-06
Newton iteration: 5 | du: 1.196e-03 | df: 5.050e-08
Newton iteration: 6 | du: 7.486e-06 | df: 1.981e-12

Assemble time average: 3.394e+00

```
vs Jax

```log
Step: 1
Newton iteration: 1 | du: 5.931e-01 | df: 7.294e-03
Newton iteration: 2 | du: 7.685e-02 | df: 5.711e-04
Newton iteration: 3 | du: 3.202e-02 | df: 2.458e-05
Newton iteration: 4 | du: 5.443e-03 | df: 7.505e-07
Newton iteration: 5 | du: 6.135e-05 | df: 1.018e-10
Newton iteration: 6 | du: 3.742e-08 | df: 3.585e-17
Step: 2
Newton iteration: 1 | du: 5.919e-01 | df: 7.085e-03
Newton iteration: 2 | du: 7.653e-02 | df: 5.552e-04
Newton iteration: 3 | du: 3.244e-02 | df: 2.529e-05
Newton iteration: 4 | du: 5.556e-03 | df: 7.630e-07
Newton iteration: 5 | du: 6.690e-05 | df: 1.170e-10
Newton iteration: 6 | du: 4.153e-08 | df: 4.385e-17
Step: 3
Newton iteration: 1 | du: 5.906e-01 | df: 6.875e-03
Newton iteration: 2 | du: 7.620e-02 | df: 5.396e-04
Newton iteration: 3 | du: 3.291e-02 | df: 2.602e-05
Newton iteration: 4 | du: 5.647e-03 | df: 7.692e-07
Newton iteration: 5 | du: 7.238e-05 | df: 1.325e-10
Newton iteration: 6 | du: 4.531e-08 | df: 5.283e-17
Step: 4
Newton iteration: 1 | du: 5.891e-01 | df: 6.662e-03
Newton iteration: 2 | du: 7.587e-02 | df: 5.243e-04
Newton iteration: 3 | du: 3.342e-02 | df: 2.677e-05
Newton iteration: 4 | du: 5.717e-03 | df: 7.694e-07
Newton iteration: 5 | du: 7.769e-05 | df: 1.478e-10
Newton iteration: 6 | du: 4.861e-08 | df: 5.882e-17
Step: 5
Newton iteration: 1 | du: 5.874e-01 | df: 6.446e-03
Newton iteration: 2 | du: 7.553e-02 | df: 5.093e-04
Newton iteration: 3 | du: 3.398e-02 | df: 2.755e-05
Newton iteration: 4 | du: 5.767e-03 | df: 7.640e-07
Newton iteration: 5 | du: 8.278e-05 | df: 1.625e-10
Newton iteration: 6 | du: 5.136e-08 | df: 6.202e-17
Step: 6
Newton iteration: 1 | du: 5.855e-01 | df: 6.230e-03
Newton iteration: 2 | du: 7.520e-02 | df: 4.948e-04
Newton iteration: 3 | du: 3.458e-02 | df: 2.835e-05
Newton iteration: 4 | du: 5.798e-03 | df: 7.540e-07
Newton iteration: 5 | du: 8.761e-05 | df: 1.763e-10
Newton iteration: 6 | du: 5.352e-08 | df: 6.848e-17
Step: 7
Newton iteration: 1 | du: 5.832e-01 | df: 6.012e-03
Newton iteration: 2 | du: 7.488e-02 | df: 4.807e-04
Newton iteration: 3 | du: 3.522e-02 | df: 2.918e-05
Newton iteration: 4 | du: 5.812e-03 | df: 7.404e-07
Newton iteration: 5 | du: 9.216e-05 | df: 1.892e-10
Newton iteration: 6 | du: 5.513e-08 | df: 7.022e-17
Assemble time average: 5.820e+00


```
</details>


Slight differences in the residual, but overall seems that Autograd is slightly faster on my laptop at least.

kinnala:
I appreciate these timing. However, as you might be aware, these are not really comparing apples to apples because the version in this branch evaluates the Jacobian with respect to all parameters of `DiscreteField` (values, gradients, second derivatives, etc.) and thus supporting arbitrary forms. It is however nice to something to compare against when looking for possible optimizations.

Here I'm solving the Navier-Stokes equations:
```python
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import *
import numpy as np


m = MeshTri().refined(3)
basis = Basis(m, ElementVector(ElementTriP2()) * ElementTriP1())


@NonlinearForm
def F(u, p, v, q, w):
    return (ddot(grad(u), grad(v))
            - dot(mul(grad(u), u[0]), v[0])
            - 1e-2 * p[0] * q[0]
            - div(u) * q[0]
            - div(v) * p[0])


x = basis.zeros()
D = basis.get_dofs().all(['u^1^1', 'u^2^1'])
x[basis.get_dofs('top').all(['u^1^1'])] = 100.

for itr in range(50):

    J, f = F.assemble(x, basis)

    xprev = x.copy()
    x += solve(*condense(J, -f, D=D))

    res = np.linalg.norm(x - xprev)
    print(res)
    if res < 1e-8:
        break


(u, ubasis), (p, pbasis) = basis.split(x)

pbasis.plot(p).show()
ubasis.plot(u).show()
```

kinnala:
Ugh, I made a variant using `autograd.make_jvp` and it is about 3 times faster than JAX.

bhaveshshrimali:
> Ugh, I made a variant using autograd.make_jvp and it is about 3 times faster than JAX.

Wait, is that so? I was following the updates this morning and to me example 10 with Jax instead of Autograd was faster (and quite fast honestly). I may have to revert to the commit before the last one to check things systematically later today, but I was testing the average assembly time per Newton iteration: 

<details>
<summary>Autograd</summary>

```python
r"""Automatic differentiation.

Warning: This example uses experimental features.

This example solves the minimal surface problem from example 10 using automatic
differentation in order to derive the tangent system for Newton's method.

"""
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import grad, dot
import numpy as np
import autograd.numpy as anp

m = MeshTri().refined(7)


@NonlinearForm(hessian=True)
def energy(u, _):
    return anp.sqrt(1. + dot(grad(u), grad(u)))


basis = Basis(m, ElementTriP1())

D = m.boundary_nodes()
x = np.sin(np.pi * m.p[0])

from time import time
times = []
for itr in range(100):
    t1 = time()
    J, F = energy.assemble(x, basis)
    t2 = time()
    times.append(t2 - t1)
    
    x_prev = x.copy()
    x += 0.7*solve(*condense(J, -F, D=D))
    res = np.linalg.norm(x - x_prev)
    if res < 1e-8:
        break
    if __name__ == "__main__":
        print(res)


print(np.mean(times))
```
</details>

<details>
<summary>Manual</summary>

```python
from skfem import *
from skfem.helpers import grad, dot
import numpy as np

m = MeshTri().refined(7)


@BilinearForm
def jacobian(u, v, p):
    w = p['prev']
    return (1 / np.sqrt(1 + dot(grad(w), grad(w))) * dot(grad(u), grad(v))
            -2 * dot(grad(u), grad(w)) * dot(grad(w), grad(v))
            / 2 / (1 + dot(grad(w), grad(w))) ** (3 / 2))


@LinearForm
def rhs(v, p):
    w = p['prev']
    return dot(grad(w), grad(v)) / np.sqrt(1 + dot(grad(w), grad(w)))


basis = Basis(m, ElementTriP1())

x = basis.zeros()

D = m.boundary_nodes()
x[D] = np.sin(np.pi * m.p[0, D])
from time import time
times = []
for itr in range(100):
    w = basis.interpolate(x)
    t1 = time()
    J = asm(jacobian, basis, prev=w)
    F = asm(rhs, basis, prev=w)
    t2 = time()
    times.append(t2 - t1)
    x_prev = x.copy()
    x += 0.7*solve(*condense(J, -F, D=D))
    if np.linalg.norm(x - x_prev) < 1e-8:
        break
    if __name__ == "__main__":
        print(np.linalg.norm(x - x_prev))

print(np.mean(times))
```
</details>


<details>
<summary>Jax</summary>

```python
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import grad, dot
import numpy as np
import jax.numpy as jnp

m = MeshTri().refined(7)


@NonlinearForm
def energy(u, _):
    return jnp.sqrt(1. + dot(grad(u), grad(u)))


basis = Basis(m, ElementTriP1())

D = m.boundary_nodes()
x = np.sin(np.pi * m.p[0])

from time import time
times = []
for itr in range(100):
    t1 = time()
    J, F = energy.assemble(x, basis)
    times.append(time() - t1)
    x_prev = x.copy()
    x += 0.7*solve(*condense(J, -F, D=D))
    res = np.linalg.norm(x - x_prev)
    if res < 1e-8:
        break
    if __name__ == "__main__":
        print(res)

print(np.mean(times))

```
</details>

And the times were something like `0.11` for manual, `0.14` for Jax and now `0.23` for autograd. But TBD...

kinnala:
I tested the nonlinear elasticity example and there assembly was down to about 1 seconds (autograd) from about 4 seconds (jax).

kinnala:
Code:
```python
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import *
import autograd.numpy as np
import time


m = MeshHex().refined(3)
basis = Basis(m, ElementVector(ElementHex1()), intorder=1)
mu, lmbda = 1., 2.


@NonlinearForm(hessian=True)
def G(u, w):
    du = grad(u)
    F = eye(np.ones(du.shape[-2:]), 3) + du
    J = det(F)
    lnJ = np.log(J)
    I1 = ddot(F, F)
    return mu / 2. * (I1 - 3.) - mu * lnJ + lmbda / 2. * lnJ ** 2


x = basis.project(lambda x: np.array([x[0] / 10., x[1] * 0., x[2] * 0.]))
D = basis.get_dofs(lambda x: np.isclose(x[0], 0) + np.isclose(x[0], 1))

for itr in range(50):

    start = time.time()
    J, f = G.assemble(x, basis)
    stop = time.time()
    print('Assembly time: {} seconds'.format(stop - start))

    xprev = x.copy()
    x += solve(*condense(J, -f, D=D))

    res = np.linalg.norm(x - xprev)
    print('Residual: {}'.format(res))
    if res < 1e-8:
        break


mdefo = m.translated(x[basis.nodal_dofs])
m.draw().show()
mdefo.draw().show()
```

kinnala:
Here is what I have running the above:
```python
Assembly time: 1.43265962600708 seconds
Residual: 0.3220785730519255
Assembly time: 1.42574143409729 seconds
Residual: 0.008043863608335308
Assembly time: 1.4221115112304688 seconds
Residual: 1.238790544973782e-05
Assembly time: 1.431492567062378 seconds
Residual: 4.522337656560638e-11
```
With JAX it was ~4.3 seconds.

kinnala:
Example 45 takes less than 0.01 seconds per assembly.

bhaveshshrimali:
> kes less than 0.01 seconds per asse

I'll test example 45. Just looking at your numbers and mine I don't think things remain the same in 2D and 3D. I just reverted to 8a365f5f2d48e1bdef3ee1484bf9cd39b7c2106d and tested jax and the timings are `0.113` for manual and `0.14` for Jax and `0.242` for Autograd.

bhaveshshrimali:
> Here is what I have running the above:
> 
> ```python
> Assembly time: 1.43265962600708 seconds
> Residual: 0.3220785730519255
> Assembly time: 1.42574143409729 seconds
> Residual: 0.008043863608335308
> Assembly time: 1.4221115112304688 seconds
> Residual: 1.238790544973782e-05
> Assembly time: 1.431492567062378 seconds
> Residual: 4.522337656560638e-11
> ```
> 
> With JAX it was ~4.3 seconds.

Cool similar numbers.. With Jax and the following code
```python
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import grad
from skfem.element import DiscreteField
from skfem.assembly.form.form import FormExtraParams
from typing import Union
from numpy import ndarray
FieldOrArray = Union[DiscreteField, ndarray]
import numpy as np
import jax.numpy as jnp
import time


m = MeshHex().refined(3)
basis = Basis(m, ElementVector(ElementHex1()), intorder=1)
mu, lmbda = 1., 2.

def dot(u: FieldOrArray, v: FieldOrArray):
    """Dot product."""
    if isinstance(u, DiscreteField) and u.is_zero():
        return u
    if isinstance(v, DiscreteField) and v.is_zero():
        return v
    return jnp.einsum('i...,i...', u, v)


def ddot(u: FieldOrArray, v: FieldOrArray):
    """Double dot product."""
    if isinstance(u, DiscreteField) and u.is_zero():
        return u
    if isinstance(v, DiscreteField) and v.is_zero():
        return v
    return jnp.einsum('ij...,ij...', u, v)

def eye(w, n):
    """Create diagonal matrix with w on diagonal."""
    return jnp.array([[w if i == j else 0. * w for i in range(n)]
                     for j in range(n)])

def det(A):
    """Determinant of an array `A` over trailing axis (if any)."""
    detA = jnp.zeros_like(A[0, 0])
    if A.shape[0] == 3:
        detA = A[0, 0] * (A[1, 1] * A[2, 2] -
                          A[1, 2] * A[2, 1]) -\
               A[0, 1] * (A[1, 0] * A[2, 2] -
                          A[1, 2] * A[2, 0]) +\
               A[0, 2] * (A[1, 0] * A[2, 1] -
                          A[1, 1] * A[2, 0])
    elif A.shape[0] == 2:
        detA = A[0, 0] * A[1, 1] - A[1, 0] * A[0, 1]
    return detA

@NonlinearForm(hessian=True)
def G(u, w):
    du = grad(u)
    F = eye(jnp.ones(du.shape[-2:]), 3) + du
    J = det(F)
    lnJ = jnp.log(J)
    I1 = ddot(F, F)
    return mu / 2. * (I1 - 3.) - mu * lnJ + lmbda / 2. * lnJ ** 2


x = basis.project(lambda x: np.array([x[0] / 10., x[1] * 0., x[2] * 0.]))
D = basis.get_dofs(lambda x: np.isclose(x[0], 0) + np.isclose(x[0], 1))

for itr in range(50):

    start = time.time()
    J, f = G.assemble(x, basis)
    stop = time.time()
    print('Assembly time: {} seconds'.format(stop - start))

    xprev = x.copy()
    x += solve(*condense(J, -f, D=D))

    res = np.linalg.norm(x - xprev)
    print('Residual: {}'.format(res))
    if res < 1e-8:
        break


mdefo = m.translated(x[basis.nodal_dofs])
m.draw().show()
mdefo.draw().show()
```

I get

```python
Assembly time: 13.871878147125244 seconds
Residual: 0.3220785730519255
Assembly time: 10.271016120910645 seconds
Residual: 0.008043863608335282
Assembly time: 10.850549459457397 seconds
Residual: 1.2387905449715476e-05
Assembly time: 10.949408531188965 seconds
Residual: 4.522337223516465e-11
```

and Autograd and the following code
```python
from skfem import *
from skfem.experimental.autodiff import NonlinearForm
from skfem.experimental.autodiff.helpers import *
import autograd.numpy as np
import time


m = MeshHex().refined(3)
basis = Basis(m, ElementVector(ElementHex1()), intorder=1)
mu, lmbda = 1., 2.


@NonlinearForm(hessian=True)
def G(u, w):
    du = grad(u)
    F = eye(np.ones(du.shape[-2:]), 3) + du
    J = det(F)
    lnJ = np.log(J)
    I1 = ddot(F, F)
    return mu / 2. * (I1 - 3.) - mu * lnJ + lmbda / 2. * lnJ ** 2


x = basis.project(lambda x: np.array([x[0] / 10., x[1] * 0., x[2] * 0.]))
D = basis.get_dofs(lambda x: np.isclose(x[0], 0) + np.isclose(x[0], 1))

for itr in range(50):

    start = time.time()
    J, f = G.assemble(x, basis)
    stop = time.time()
    print('Assembly time: {} seconds'.format(stop - start))

    xprev = x.copy()
    x += solve(*condense(J, -f, D=D))

    res = np.linalg.norm(x - xprev)
    print('Residual: {}'.format(res))
    if res < 1e-8:
        break


mdefo = m.translated(x[basis.nodal_dofs])
m.draw().show()
mdefo.draw().show()
```

I get 

```python
Assembly time: 1.7155475616455078 seconds
Residual: 0.32207857305192544
Assembly time: 1.6871576309204102 seconds
Residual: 0.008043863608335288
Assembly time: 1.701439380645752 seconds
Residual: 1.2387905449711992e-05
Assembly time: 1.6698722839355469 seconds
Residual: 4.52233849072165e-11
```


which for me is about ~6x faster than Jax..

adtzlr:
Personally I'd ask the developers of Autograd/JAX (same team) if we have two examples. I **think** there is a relatively simple explanation for the timings. Plus, with JAX there is the (currently unused) jit option. While I personally really like Autograd, JAX is the succesor of it and it is the package which is actively maintained...

kinnala:
Change from autograd to JAX is rather simple, affects just a few lines of code, so we can go back to JAX if it gets faster in the future.

skfem/element/element_tri/element_tri_p2b.py
```diff
@@ -0,0 +1,55 @@
+import numpy as np
+
+from ..element_h1 import ElementH1
+from ...refdom import RefTri
+
+
+class ElementTriP2B(ElementH1):
+    """Quadratic triangular element with a cubic bubble."""
```
bhaveshshrimali, line 8: I'm wondering if this is the same as [`element_tri_ccr`](https://github.com/kinnala/scikit-fem/blob/master/skfem/element/element_tri/element_tri_ccr.py), is it ?

bhaveshshrimali (commented):
The performance is quite good with autograd and `NonlinearForm` usage is clear too. I just happen to take a look at `element_tri_p2b` which looks identical to `element_tri_ccr` (the last I checked in Brezzi its the same though the former is a more common name).

skfem/element/element_tri/element_tri_p2b.py
kinnala: Thanks, yes. I deleted element_tri_ccr and also renamed element_tri_mini to element_tri_p1b. Those old names are still exported in __init__.

PR #909, opened 2022-06-20
RT facet asm

### LINKED ISSUE #869: FacetBasis and ElementTriRT0 fails
I think there is some issues with broadcasting because of the following failure:
```python
In [1]: from skfem import *
In [2]: FacetBasis(MeshTri(), ElementTriRT0())
...
ValueError: operand has more dimensions than subscripts given in einstein sum, but no '...' ellipsis provided to broadcast the extra dimensions.
```

kinnala:
Fixes #869


kinnala:
Kuvakaappaus - 2022-08-02 11-49-19

New pic for RT example.

PR #937, opened 2022-10-19
Add ElementTriN0 and extend ElementHcurl to 2D

kinnala:
Related to #895
- [ ] Add tests


kinnala:
There is something wrong with `ElementHcurl` 2D implementation. I'll need to look into it more carefully.

kinnala:
Alright, now I realized I had to set div E = 0 using a Lagrange multiplier to get rest of the modes. I don't think the implementation in the reference is entirely correct.

PR #942, opened 2022-10-20
Add example for mode solving of rectangular waveguide

HelgeGehring:
I've started working on a simple example for the ElementTriN0, but it needs some fixes for the boundary conditions/eigenvalue finding to match the example in the link.

### DIFF
--- a/docs/examples/ex46.py
+++ b/docs/examples/ex46.py
@@ -1,0 +1,57 @@
+"""
+Like http://diyhpl.us/~nmz787/pdf/Using_the_FEniCS_Package_for_FEM_Solutions_in_Electromagnetics.pdf (4.1)
+"""
+
+import numpy as np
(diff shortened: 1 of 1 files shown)


docs/examples/ex46.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@BilinearForm
+def t_zz_ij(N_i, L_i, N_j, L_j, w):
+    return epsilon * L_i * L_j
+
+
+s_ij = asm(s_tt_ij, basis) + asm(s_zz_ij, basis)
+t_ij = asm(t_tt_ij, basis) + asm(t_zz_ij, basis)
+
```
kinnala, line 43: I'm quite sure this value of D does not lead to correct boundary conditions. Perhaps D=basis.get_dofs() is correct?

PR #949, opened 2022-11-04
added ones function

gatling-nrl:


gatling-nrl:
@kinnala this is just a simple wrapper to the np.ones() function exactly like the zeros() function. For convenience.

kinnala:
How do you plan using it? Is basis.zeros() + 1 enough for this purpose?

gatling-nrl:
yes, `basis.zeros()+1` is equivalent to this. I added basis.ones() because I use ones much more frequently to initialize coefficients. and e.g. `5*basis.ones()` is much preferred by me over `5*(basis.zeros() + 1)`. There are several other workarounds, including `5*np.ones(basis.N)` but since zeros and ones are both in numpy, it seemed natural to have it here as well.

kinnala:
> I added basis.ones() because I use ones much more frequently to initialize coefficients.

This is quite surprising because many times coefficient = 0 matches some zero essential boundary condition for any finite elements but coefficient = 1 has meaning only for very few elements like P1 and P0.

This is why I'm asking. Should this PR at least provide an example where it is used in a test case or by modifying some of the examples?

gatling-nrl:
> Should this PR at least provide an example where it is used

Yes, it probably should include a test case. 

image

I'm often working with an equation like this, and it is pretty typical lines of code like this:
```python
sigma=np.ones(p0.N)
sigma[p0.get_dofs(elements=something)] = a_number
```

But since in other places I have `p2.zeros()` it seems more natural/explicit to write `p0.ones()`.

I just haven't had time to reduce my codes to something suitable for a testcase or example. But its on the to-do list

kinnala:
It can be something simple, just that it gets called at least once in the test suite. Otherwise it's easily forgotten if there are any refactorings. When I rewrote Mesh class some time ago I forget to add many methods that weren't called in the test suite.

PR #950, opened 2022-11-04
Allow interpolation of complex fields

HelgeGehring:
Would be great to allow interpolating complex fields


gatling-nrl:
@HelgeGehring can you make sure #949 is compatible with this? Looks straight forward enough

HelgeGehring:
@gatling-nrl looks perfectly compatible to me :+1:

PR #952, opened 2022-11-07
add possibility to set the minlength in quiver

HelgeGehring:


kinnala:
Seems mypy has changed its behaviour again...

kinnala:
In #953 you can write `plot(m, plot_kwargs={'minlength': ...})`.

PR #956, opened 2022-11-08
allow projection of vector fields

HelgeGehring:
I'm trying to project on a Basis which has a Vector, like the following:

```python
import numpy as np

from skfem import *

mesh = MeshTri.init_tensor(x=np.linspace(-1, 1, 20), y=np.linspace(-1, 1, 20))

basis = Basis(mesh, ElementVector(ElementTriP1(), 2))

x = basis.project(lambda w: (w[1], -w[0]))

basis.plot(x)
plt.show()
```
Like this I get an error like
```python
  File ".../scikit-fem/skfem/assembly/basis/abstract_basis.py", line 410, in _projection
    assert len(interp) == len(self.basis[0])
AssertionError
```

This pull-request would fix this for this case, but it seems that it should be fixed somehow different.

### DIFF
--- a/skfem/assembly/basis/abstract_basis.py
+++ b/skfem/assembly/basis/abstract_basis.py
@@ -404,7 +404,7 @@ class AbstractBasis:
         from skfem.assembly import BilinearForm, LinearForm
         from skfem.helpers import inner
 
-        interp = self._normalize_interp(interp)
+        interp = np.array(self._normalize_interp(interp))
         if not isinstance(interp, tuple):
             interp = (interp,)
         assert len(interp) == len(self.basis[0])


kinnala:
I think it should work already, the output of the lambda just needs to be a numpy array and not tuple.

HelgeGehring:
Okay, thanks, that fixed it!

HelgeGehring:
The more I think about this the more I have the question:
Is there a reason that project doesn't accept a lambda which returns a tuple/list?
Would there be a reason not to convert a tuple/list in `project` to a numpy array?
Would somehow let it look cleaner 🤔
Just curious, if there's any reason against it

kinnala:
Perhaps there isn't. One needs to carefully think about the distinctiom between ElementComposite and ElementVector. If those work in a consistent fashion, then it'd be fine.

kinnala:
If you look here: https://github.com/kinnala/scikit-fem/blob/master/skfem/assembly/basis/abstract_basis.py#L389

You can see that it is not as simple as you think. If the output of the callable is tuple, those are already interpreted as different components of ElementComposite. It wouldn't be possible to make this change backwards-compatible.

PR #958, opened 2022-11-09
Fix Hcurl in 2D for arbitrary meshes and add tests

kinnala:
TODO:
- [x] Add also simple 3D test
- [x] Changelog entries


HelgeGehring:
Awesome! The calculation of the z-component for my application seems to work fine :)

PR #993, opened 2022-12-14
inmutable defaults for dataclasses for python 3.11

HelgeGehring:
Python 3.11 doesn't allow mutable defaults for dataclasses anymore. This fix solves the problem, but somehow doesn't look so nice. Also, there's no need to have multiple instances of the `doflocs` (?)

Would you have a better way to fix this?


HelgeGehring:
And could you maybe add the autoformat settings you're using to the pyproject.toml? or which command are you using?

kinnala:
Strange for Python to introduce such backwards-incompatibilities. Perhaps changing to using a dataclass for Mesh some time ago was a mistake.

kinnala:
Oops. I use Emacs for autoindent.

kinnala:
I think those reports are from flake8.

HelgeGehring:
Okay, I've updated the formatting. I'd guess for now it would nice to merge it to have compatibility with Python 3.11. If there's an better way to do this we can still update it later :)

kinnala:
Yes, I'll see that the tests are added.

PR #1006, opened 2023-01-12
Multipoint constraints

HelgeGehring:
As discussed in #989

The example in the file compares it to the penalty method. I'm not sure how this would go best together with condense, as condense is a subset of the functionality of this implementation.

### DIFF
--- a/skfem/multipoint.py
+++ b/skfem/multipoint.py
@@ -49,7 +49,8 @@ def restore(y, T, g):
 
 
 if __name__ == "__main__":
-    from skfem import MeshTri, ElementTriP1, Basis, FacetBasis, asm, BilinearForm, solve, condense
+    from skfem import (MeshTri, ElementTriP1, Basis,
+                       FacetBasis, asm, BilinearForm, solve, condense)
     from skfem.models import laplace, unit_load
 
     m = MeshTri().refined(6)
--- a/skfem/multipoint.py
+++ b/skfem/multipoint.py
@@ -17,7 +17,11 @@ def multipoint(D, *Ixs, g=None):
     for Ix in Ixs:
         I, x = Ix
         result += csr_array(
-            (x, (D.flatten(), [I_columns.tolist().index(i) for i in I.flatten()])), shape=shape
+            (x,
+                (D.flatten(),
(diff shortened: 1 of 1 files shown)


HelgeGehring:
It would be great to have it in a form which supports nonlinear problems :thinking:

kinnala:
> It would be great to have it in a form which supports nonlinear problems 🤔

That's not possible I'm afraid. There is no special support for nonlinearities in scikit-fem and all the related iterations are expected to be implemented by the user.

kinnala:
I started looking this PR yesterday and made some refactoring locally. However, I don't have anything ready yet. I want to simplify it since I think we don't need so many different functions.

HelgeGehring:
image
Okay, I should be more precise:
I didn't mean the class of Nonlinear functions (there I see that it's way to hard to generalize and usually as the user I'd like to have control how it's done, i.e. for coupled systems) I was just thinking of quadratic and polynomial problems.

I think also taking care of that the multi-point constraint supports quadratic/polynominal functions should only require very few thoughts (mostly I'm unsure how to transform `g`, the higher order coefficients should be similar to the linear one?)
Iterations for the class of Nonlinear problems should already be possible like it is, as it's treated like a linear system after the assembly.

Also the linearization of quadratic functions is straight forward.
For quadratic it can be done by
image
rewritten as
image
This could be implemented as a simple utility-function.

But I'd also imagine it nicely implemented in the solve-function:
- The solve-function already decides depending on the datatype of `b` between a linear and an eigen problem. 
- The same could be done with `A`: Depending on the datatype (spmatrix or tuple) it's either a linear or a polynominal problem. The standard way of solving would be the linearization of the polynominal problem and then the usage of a solver depending on `b`.
- If the user wants to select the solver, it could be either a direct solver for a polynominal problem, or a linear solver can be wrapped by a linearization.

The perfect compromise I'd see here is
- Keeping the class of Nonlinear problems completely to the user
- Implementing the multi-point constraint compatible to polynominal problems
- Extending `solve` to polynominal problems, while keeping the interface the same (A would in that case just be seen as the left-hand side)

HelgeGehring:
And simplifying sounds good! I mostly concentrated on making it understandable. Looking forward to a cleaner interface :)

PR #1009, opened 2023-01-14
Add skfem.utils.mpc

kinnala:
Here are some examples you can try @HelgeGehring 
```python
import numpy as np
import scipy.sparse
from skfem import *
from skfem.utils import mpc
from skfem.models import laplace, unit_load

m = MeshTri().refined(5)
e = ElementTriP1()

basis = Basis(m, e)

left = basis.get_dofs(facets='left')
right = basis.get_dofs(facets='right')
top = basis.get_dofs(facets='top')
bottom = basis.get_dofs(facets='bottom')

A = asm(laplace, basis)
f = asm(unit_load, basis)

# periodic
M = left
S = np.concatenate((right, top, bottom))
T = None

# dirichlet
# M = None
# S = np.concatenate((right, top, bottom, left))
# T = None

# periodic x2
# M = right
# S = np.concatenate((left, top, bottom))
# T = 2 * scipy.sparse.eye(len(S), len(M))

y = solve(*mpc(A, f, M=M, S=S, T=T))

basis.plot(y[:basis.N], colorbar=True, shading='gouraud', levels=5).show()
```

### DIFF
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -622,7 +622,7 @@ def mpc(A: spmatrix,
         The set of DOFs that are kept in the system.
     T
     g
-        The constraint is of the form `x[M] = T @ x[S] + g`.
+        The constraint is of the form `x[S] = T @ x[M] + g`.
 
     """
     if M is None:
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -211,7 +211,7 @@ def solve_eigen(A: spmatrix,
         L, X = solver(A, M, **kwargs)
         y = np.tile(x.copy()[:, None], (1, X.shape[1]))
         if isinstance(I, tuple):
-            y[I[0]] = np.array([I[1](x) for x in X.T]).T
+            np.add.at(y, I[0], np.array([I[1](x) for x in X.T]).T)
         else:
             y[I] = X
         return L, y
@@ -231,7 +231,7 @@ def solve_linear(A: spmatrix,
     if x is not None and I is not None:
         y = x.copy()
         if isinstance(I, tuple):
-            y[I[0]] = I[1](solver(A, b, **kwargs))
+            np.add.at(y, I[0], I[1](solver(A, b, **kwargs)))
         else:
             y[I] = solver(A, b, **kwargs)
         return y
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -646,7 +646,11 @@ def mpc(A: spmatrix,
         M = np.setdiff1d(M, D)
         S = np.concatenate((S, D))
 
-    U = np.setdiff1d(np.arange(A.shape[0]), np.concatenate((M, S)))
+    # for mypy
+    assert isinstance(S, ndarray) and isinstance(M, ndarray)
+
+    U = np.setdiff1d(np.arange(A.shape[0], dtype=np.int64),
+                     np.concatenate((M, S)))
 
     if T is None:
         T = sp.eye(len(S), len(M))
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -635,6 +635,9 @@ def mpc(A: spmatrix,
         raise NotImplementedError("S must be specified.")
     S = _flatten_dofs(S)
 
+    # for mypy
+    assert isinstance(S, ndarray) and isinstance(M, ndarray)
+
     # remove DOFs in D from (S, M) and add D to S
     if D is not None:
         D = _flatten_dofs(D)
@@ -642,7 +645,6 @@ def mpc(A: spmatrix,
         M = np.setdiff1d(M, D)
         S = np.concatenate((S, D))
 
-    assert isinstance(S, ndarray) and isinstance(M, ndarray)
     U = np.setdiff1d(np.arange(A.shape[0]), np.concatenate((M, S)))
 
     if T is None:
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -611,7 +611,7 @@ def mpc(A: spmatrix,
         g: Optional[ndarray] = None) -> CondensedSystem:
 
     if M is None:
-        M = np.array([])
+        M = np.array([], dtype=np.int64)
     else:
         M = _flatten_dofs(M)
     S = _flatten_dofs(S)
--- a/skfem/utils.py
+++ b/skfem/utils.py
@@ -605,11 +605,12 @@ def condense(A: spmatrix,
 
 def mpc(A: spmatrix,
         b: ndarray,
-        S: Optional[DofsCollection] = None,
-        M: Optional[DofsCollection] = None,
+        S: Optional[ndarray] = None,
+        M: Optional[ndarray] = None,
         T: Optional[spmatrix] = None,
         g: Optional[ndarray] = None,
-        D: Optional[DofsCollection] = None) -> CondensedSystem:
+        D: Optional[DofsCollection] = None,
+        x: Optional[ndarray] = None) -> CondensedSystem:
     """Apply a multipoint constraint on the linear system.
 
     Parameters
@@ -618,13 +619,16 @@ def mpc(A: spmatrix,
     b
         The linear system to constrain.
     S
-        The set of DOFs that are eliminated from the system.
+        Array of DOFs that are eliminated from the system.
     M
-        The set of DOFs that are kept in the system.
+        Array of DOFs that are kept in the system.
     T
     g
         The constraint is of the form `x[S] = T @ x[M] + g`.
     D
+        An array of additional DOFs that are set to a specific value.
+    x
+        An array of values to use together with `D`.
 
     """
     if M is None:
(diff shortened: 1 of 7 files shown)


kinnala:
Took me a while to realize there was a bug in the reference we used for MPC.

HelgeGehring:
Looks good!

Where was a bug in the reference? I also found https://quickfem.com/wp-content/uploads/IFEM.Ch08.pdf , which explaines it in way more detail (Actually, I somehow think that might have been the source for the slides)
(Table of contents see https://quickfem.com/theory/finite-element-analysis/ , seems like a nice reference)

It seems to me, that this implementation only allows two-point constraints, right? I don't see a way to implement something like 
image
That was the reason for me to allow multiple master degrees for one slave degree.

Might it make sense to add a function which combines constraints? While these examples look good, I could think of examples, where such a function could provide:
- a nicer way to define the constraints (like this it needs a moment to understand that i.e. the periodic example also imposes dirichlet boundary constraints)
- checking that DOFs are not constrained twice (does the combination of periodic left<->right and dirichlet top/bottom lead to something completely correct? I applied them one after the other and saw that the order matters) 
- a convenient way to debug, as it'd be simpler to disable some of the constraints

One example which I'll probably want to use is periodic boundaries with top<->bottom and left<->right, I'd guess there we would need to take special care of the edge points of the mesh (?)

Also I'd be happy to have something for quadratic/polynominal equations as described in #1006, of course that should be an extra pull request, but I'm thinking if this could be somehow "prepared" for that. What do you think?

kinnala:
There are several terms missing in "General MPC" slide. The new reference you pointed to has all of the terms.

Yes, typically you'd need to think carefully what happens at the corner points of (doubly) periodic mesh. There is a test case somewhere with doubly periodic implemented using MeshTri1DG and I had to do some serious debugging to get it working. I think the main difference of this new MPC function and MeshTri1DG is that the latter cannot support relative phase.

Anyhow, I'm not sure whether combining these quadruplets (S, M, T, g) is in general as easy as you make it seem in your PR. II'll take a look at it at some point but I don't yet understand it fully.

Now whether solve() starts supporting more general eigenvalue problems, that should be a different PR. I suppose everything would happen inside solve() when, e.g., a tuple is provided as the first argument?

kinnala:
You say "checking that DOFs are not constrained twice". Isn't that sometimes what we want to do?

kinnala:
For instance, you can think the doubly periodic case with the corner points A, B, C, D.

You might constrain A=B and A=C and A=D, or alternatively, A=B and C=D and A=C. But in any case you constrain the same DOF at least twice, unless I misunderstand something.

HelgeGehring:
How would we implement the double periodic mesh with the MPC? Probably it would just be selecting one as the master and making the other three dependent?
With the current implementation, would it work if a DOF is both in slave and in master?
E.g. for the double periodic boundaries, it would result in DOFs being in both lists, would that be okay or should we check for that (or even fix that, as if one is in both, it's eliminated and therefore we can replace it by the master which stays?)

My implementation of the combination was mostly intuitive, I just thought of the reduced system as a new system and how I would apply the constraints to that one. That way the double periodic constraint should work?
But yes, I didn't do the complete math, so there might be some trouble hidden :/

And with double constraint I mean, that each DOF can be only once replaced, so if one DOF is twice in the Slave-List, it would create a problem? For that we could check `len(np.unique(S))==len(S)`?

HelgeGehring:
How would we implement the double periodic mesh with the MPC? Probably it would just be selecting one as the master and making the other three dependent?
With the current implementation, would it work if a DOF is both in slave and in master?
E.g. for the double periodic boundaries, it would result in DOFs being in both lists, would that be okay or should we check for that (or even fix that, as if one is in both, it's eliminated and therefore we can replace it by the master which stays?)

My implementation of the combination was mostly intuitive, I just thought of the reduced system as a new system and how I would apply the constraints to that one. That way the double periodic constraint should work?
But yes, I didn't do the complete math, so there might be some trouble hidden :/

And with double constraint I mean, that each DOF can be only once replaced, so if one DOF is twice in the Slave-List, it would create a problem? For that we could check `len(np.unique(S))==len(S)`?

HelgeGehring:
About the quadratic/polynomial, I'd say that's the perfect way to implement it! A would be generalized as the polynomial left-hand side :)

kinnala:
Doubly periodic:
```python
import numpy as np
import scipy.sparse
from skfem import *
from skfem.utils import mpc
from skfem.models import laplace, unit_load

m = MeshTri.init_sqsymmetric().refined(5)
e = ElementTriP1()

basis = Basis(m, e)

left = basis.get_dofs(facets='left')
right = basis.get_dofs(facets='right')
top = basis.get_dofs(facets='top')
bottom = basis.get_dofs(facets='bottom')

topleft = basis.get_dofs(at=[0, 1])
topright = basis.get_dofs(at=[1, 1])
bottomright = basis.get_dofs(at=[1, 0])
bottomleft = basis.get_dofs(at=[0, 0])

A = asm(laplace, basis)
B = asm(BilinearForm(lambda u, v, w: u * v), basis)
f = asm(LinearForm(lambda v, w: np.sin(2. * np.pi * w.x[0]) * np.sin(2. * np.pi * w.x[1])), basis)

# periodic
# M = left
# S = np.concatenate((right, top, bottom))
# T = None

# doubly periodic
M = np.concatenate((
    np.setdiff1d(left, topleft, bottomleft),
    np.setdiff1d(top, topleft, topright),
    topleft,
    topleft,
    topleft,
))
S = np.concatenate((
    np.setdiff1d(right, topright, bottomright),
    np.setdiff1d(bottom, bottomleft, bottomright),
    topright,
    bottomright,
    bottomleft,
))
T = None

# dirichlet
# M = None
# S = np.concatenate((right, top, bottom, left))
# T = None

# periodic x2
# M = right
# S = np.concatenate((left, top, bottom))
# T = 2 * scipy.sparse.eye(len(S), len(M))

y = solve(*mpc(A + 1e-8 * B, f, M=M, S=S, T=T))

basis.plot(y[:basis.N], colorbar=True, shading='gouraud', levels=5).show()
```

HelgeGehring:
Looks perfect! Thanks!

Would it make sense to introduce checks, that:
- Is in S each DOF only once?
- No DOF which is in M is in S (this would create a problem, right? These could probably be solved in mpc, this would e.g. allow the double periodic constraints without caring ourself about the corner nodes. The constraints would just need to be chained. For example A=2B, B=3C could be corrected to A=6C, B=3C)
- the size of T and g fits to M and S

I'd guess these could save quite some debugging time when using such constraints

kinnala:
Thanks for listing those out, saves me some time. I'll add them together with a docstring in few days.

HelgeGehring:
Awesome, thanks!

skfem/utils.py
```diff
@@ -597,6 +603,68 @@ def condense(A: spmatrix,
...
+    A
+    b
+        The linear system to constrain.
+    S
+        The set of DOFs that are eliminated from the system.
+    M
+        The set of DOFs that are kept in the system.
+    T
+    g
+        The constraint is of the form `x[M] = T @ x[S] + g`.
```
HelgeGehring, line 625: I''d guess it should be 
```suggestion
        The constraint is of the form `x[S] = T @ x[M] + g`.
```
as the slave DOFs are the restricted ones, right?

HelgeGehring:
It would also be great if we could add support for generalized eigenvalue problems, but I'm not sure how to combine `b` and `g` in that case.

skfem/utils.py
```diff
@@ -597,6 +603,68 @@ def condense(A: spmatrix,
...
+            (A[M].T[M].T + T.T @ A[S].T[M].T + A[M].T[S].T @ T
+             + T.T @ A[S].T[S].T @ T),
+        ]], 'csr')
+    y = np.concatenate((b[U] - A[U].T[S].T @ g,
+                        b[M] - A[M].T[S].T @ g))
+
+    return (
+        B,
+        y,
+        np.zeros_like(b),
```
HelgeGehring, line 660:
```suggestion
        np.zeros_like(b, dtype=B.dtype),
```
would be great to also add here support for complex values.

HelgeGehring:
One thing I'm wondering about: 
Why do you use `A[U].T[U].T` instead of `A[U][:,U]`?
I would somehow guess that the latter one is more efficient as it doesn't need to transpose the matrix twice?

skfem/utils.py
kinnala: Ah, yes. Of course.

kinnala:
> One thing I'm wondering about:
> Why do you use `A[U].T[U].T` instead of `A[U][:,U]`?
> I would somehow guess that the latter one is more efficient as it doesn't need to transpose the matrix twice?

These matrices are CSR so rows are basically trivial to slice while cols require some nontrivial calculation. However, I've actually never compared the performance of the two approaches. That should be really easy.

kinnala:
It seems that your suggestion is about 20-30% faster.

HelgeGehring:
> It seems that your suggestion is about 20-30% faster.

Thanks for testing, good to know :) Might also make the code a bit easier to read.

PR #1013, opened 2023-01-26
add multipoint constraint for generalized eigenproblems

HelgeGehring:
I've updated it to the latest version. With the "fixed equations"  in 8bbe102a272ac8d2dfa747054ac3c13001b2ce9f it doesn't work for ne, so I stayed with the previous ones.
I don't know yet how to include the `g`, so it's not yet included in the generalized case

### DIFF
--- a/skfem/mesh/mesh.py
+++ b/skfem/mesh/mesh.py
@@ -390,7 +390,7 @@ class Mesh:
 
         """
         midp = self.p[:, self.t].mean(axis=1)
-        return np.nonzero(test(midp))[0]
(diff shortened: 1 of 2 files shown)


kinnala:
The reason I changed the equations was that the original ones were assuming symmetric problems.

HelgeGehring:
Thanks! Not being restricted to symmetric problems sounds good.
But in the generalized case, the right-hand side should transform exactly like the left-hand side, right? (and `g` would also need to be put in)

PR #1018, opened 2023-02-16
add with-elements to cell_basis

HelgeGehring:
Like `with_element`, but for restricting the basis to a certain subset of elements.
Maybe there's a better name for the method to avoid confusion with `with_element` :thinking:

### DIFF
--- a/skfem/assembly/basis/cell_basis.py
+++ b/skfem/assembly/basis/cell_basis.py
@@ -237,6 +237,16 @@ class CellBasis(AbstractBasis):
             elements=self.tind,
         )
 
+    def with_elements(self, elements: Optional[Any] = None) -> 'CellBasis':
+        """Return a similar basis on a subset of element indices."""
+        return type(self)(
+            self.mesh,
+            self.elem,
+            mapping=self.mapping,
+            quadrature=self.quadrature,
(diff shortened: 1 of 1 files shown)


kinnala:
I wonder if it should be called just `restrict`. There is also `Mesh.restrict` which actually removes the remaining elements.

HelgeGehring:
Sounds good! The only downside I see would be, that there's a difference between 
```Basis(mesh.restrict, ...)```
and
```Basis(mesh, ...).restrict```
as the first one would also restrict the DOFs and the here proposed function leaves the DOFs the same.

I'd guess, if we add `restrict` to `Basis` both expressions above should behave the same?

While I think adding `restrict` to `Basis` has a value, I think the here proposed function should have a different name :thinking:
Maybe `with_elements` is just fine, as it matches the name in the constructor?

kinnala:
I agree with your thinking.

kinnala:
Btw, I now merged this but we usually have a changelog entry for new functionality in README.md. I also like to have some kind of simple test for each method so that it won't be forgotten as time passes and refactorings happen. If I ever rewrite Basis it is easy to forgot things such as this because the test suite runs just fine.

PR #1021, opened 2023-02-28
MeshTri: Keep subdomains when refining

HelgeGehring:
I've made first implementation for keeping the subdomains while refinement.
But I'd guess there's some stuff missing at the moment:

- [ ]  I'd guess it can be made way faster by using numpy better.
- [ ] The facets need to be updated, too (here I don't have an idea yet how)
- [ ] I don't really know how to write tests for this

### DIFF
--- a/skfem/mesh/mesh_tri_1.py
+++ b/skfem/mesh/mesh_tri_1.py
@@ -333,16 +333,40 @@ class MeshTri1(MeshSimplex, Mesh2D):
                   m.p[:, m.facets[1, facets == 1]])
         
         def old_to_new_subdomain(old_elements):
-            sums = [np.count_nonzero(section) for section in (rest, red, blue1, blue2, green)]
+            sums = [np.count_nonzero(section)
+                    for section in (rest, red, blue1, blue2, green)]
             starts = np.cumsum(np.array((1,4,3,3,2)) * sums)
 
-            rest_t = np.cumsum(rest, dtype=np.uint64)[np.intersect1d(np.where(rest)[0].astype(np.uint64),old_elements)]-1
+            rest_t = np.cumsum(rest, dtype=np.uint64)[
+                np.intersect1d(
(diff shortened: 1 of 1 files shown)


kinnala:
There used to be a vectorized implementation of this before Mesh was rewritten using dataclass, I believe.

HelgeGehring:
uijj, vectorized sounds good! was there a reason, that it didn't survive the rewriting?

kinnala:
Me being lazy. It should be relatively easy to do for all Mesh classes at once since I believe the new elements always have a specific order. This order is defined by the hstack call here: https://github.com/kinnala/scikit-fem/blob/507a75008e168adca4848819d401bb697c30bc57/skfem/mesh/mesh_tri_1.py#L219

This defines a mapping between the old and the new elements of shape (4, nelems_old) for MeshTri and MeshQuad, (6, nelems_old) for MeshTet, etc. The refine stacking order should be added to RefDom subclasses and then this could be generalized to all Mesh types.

kinnala:
Now I see that you do it here for the adaptive refinement. This is something else than what I was talking about (uniform refinement). This never existed and I'll check this when I have time.

HelgeGehring:
Yes, I'm hoping do eventually get some adaptive refinement :) Thanks a lot!

kinnala:
Idea for a test: 1. Create subdomains using Mesh.elements_satisfying(lambda x: x[0]<0.5) or similar and refine elements here and there. 2. Refine elements first and then create the subdomains. 3. Check that results match.

kinnala:
I think looking at the order of those hstack calls you can get this new_to_old mapping directly. I'll try it out.

kinnala:
Uniform refinement now preserves subdomains for all meshes. I'll check the adaptive refinement next.

kinnala:
Done in #1024. I chose a slightly different approach which relies on the a priori knowledge of the order of red, blue, green elements in the new mesh.

PR #1034, opened 2023-05-29
Add hexahedral skeleton element

kinnala:


gdmcbain:
Hexagonal? Or hexahedral?

kinnala:
> Hexagonal? Or hexahedral?

Thanks. I was testing some generalisations of tetrahedral elements to hexahedral meshes so I had to implement this variant.

PR #1035, opened 2023-05-31
Fix Mesh.refined()

HelgeGehring:
At the moment Mesh.refined(ix) changes also the current mesh as `t` is directly modified. If we use a copy here only the refined mesh is altered.


kinnala:
Thanks!

kinnala:
@HelgeGehring it is done now, waiting for some pypi caches to refresh

HelgeGehring:
Thanks a lot!

PR #1091, opened 2024-01-10
fix split_bases and split_indices for ElementVector

HelgeGehring:
At those two lines it should be the number of dimensions of the `ElementVector` not the number of dimensions of the mesh, right?
(I'm having a 3D vector in a 2D mesh)

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 with respect to documented and/or tested features.
 
+### Unreleased
+
+- Fixed: `ElementVector` works also for split_bases/split_indices in case `mesh.dim() != elem.dim`
+
 ### [9.0.0] - 2023-12-24
 
 - Removed: Python 3.7 support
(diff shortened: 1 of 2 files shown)


kinnala:
Yes, ElementVector dim kwarg is not used too much in the examples/tests.

HelgeGehring:
Yay, thanks for merging so quickly!
Could you do a minor release? I'd like to use this :)

kinnala:
I'm still considering whether we should also add numpy<2 at the same time because the release is in 2-3 months and I expect there to be issues with "pip install scikit-fem[all]". I already fixed most of the issues within scikit-fem in #1090 but matplotlib and meshio are currently broken so it will affect UX here unless it is done. What do you think?

HelgeGehring:
Pinning now would also lead to quite a bit of trouble - if someone has numpy pre-installed, pinning now would lead to an older version of scikit-fem being installed, which would cause even more trouble :/
Thus, it would probably lead for most users to have more problems (as for example anaconda has numpy pre-installed) :/

I'd guess the best would be to just go on releasing and add a warning at import which lists the current problems and suggests downgrading numpy in case?
Matplotlib will probably be fixed when numpy 2 is released, but meshio I don't know :(

kinnala:
I don't think that should happen. I won't drop support for numpy<2 within scikit-fem any time soon as the changes are very minimal and the users of anaconda or other older numpy version should be able to install whatever version, no matter what numpy version they use.

I'm proposing to add numpy<2 for the short time that the issues with the dependencies are solved. Something along the lines: https://pythonspeed.com/articles/numpy-2/

kinnala:
Perhaps it is not necessary and I'm just overthinking. I've been also considering to fork `meshio` because there are some things I'd like to fix there but I am not sure whether I have the time to maintain such a fork. Perhaps I'd just fork the parts relevant to scikit-fem to make it less popular for other users.

kinnala:
There is now a release 9.0.1 coming in few hours.

PR #1122, opened 2024-04-23
Support composite Basis (e.g., for mortaring)

kinnala:
Right now the use of "composite forms" such as
```python
def form(u1, u2, v1, v2, w):
    return ...
```
requires that we create `ElementComposite`, e.g., multiplying `ElementTriP2() * ElementTriP1()` and then `u1` and `u2` would correspond to different unknowns but be defined over the same mesh.

Now there are situations where you would like to have `u1` and `u2` be defined using different `Basis` objects instead. One simple example is when using `InteriorFacetBasis(..., side=0)` and `InteriorFacetBasis(..., side=1)` to implement DG methods. Another example is when using mortaring (see ex04) so that the `FacetBasis` objects are coming from the boundary of two different meshes.

This PR includes works toward something like `BasisComposite` which allows multiplying two compatible `Basis` objects for the same effect, i.e. using both of them within forms.

### DIFF
--- a/skfem/assembly/basis/abstract_basis.py
+++ b/skfem/assembly/basis/abstract_basis.py
@@ -451,3 +451,23 @@ class AbstractBasis:
             logger.warning("First argument, 'visuals', must be a string.")
         mod = importlib.import_module('skfem.visuals.{}'.format(visuals))
         return mod.draw(self, **kwargs)
+
+    def __mul__(self, other):
+        from copy import deepcopy
+        assert len(self.basis) == len(other.basis)
+        basis = []
+        element_dofs = []
+        for itr in range(len(self.basis)):
+            basis.append((self.basis[itr][0], other.basis[0][0].zeros()))
+            element_dofs.append(self.element_dofs[itr])
+        for itr in range(len(other.basis)):
+            basis.append((self.basis[0][0].zeros(), other.basis[itr][0]))
+            element_dofs.append(other.element_dofs[itr] + self.N)
+        out = deepcopy(self)
+        out.basis = basis
+        out.dofs.N = self.N + other.N
+        out.Nbfun = self.Nbfun + other.Nbfun
+        out._element_dofs = element_dofs
+        out.interpolate = lambda w: (self.interpolate(w[:self.N]),
+                                     other.interpolate(w[self.N:]))
+        return out


kinnala:
Here is what I've been testing:
```python
from skfem import *
from skfem.experimental.autodiff import *
from skfem.experimental.autodiff.helpers import *
import jax.numpy as np
import numpy as onp

m1 = MeshTri.init_tensor(np.linspace(0, 5, 60), np.linspace(0, 0.25, 5))
m2 = MeshTri.init_tensor(np.linspace(0, 5, 51), np.linspace(-0.5, -0.25, 4))

from skfem.supermeshing import intersect, elementwise_quadrature

m1t, orig1 = m1.trace('bottom', mtype=MeshLine, project=lambda p: np.array(p[0]))
m2t, orig2 = m2.trace('top', mtype=MeshLine, project=lambda p: np.array(p[0]))
m12, t1, t2 = intersect(m1t, m2t)
e1 = ElementVector(ElementTriP1())
e2 = ElementVector(ElementTriP1())

basis1 = Basis(m1, e1)
basis2 = Basis(m2, e2)
fbasis1 = FacetBasis(m1, e1, quadrature=elementwise_quadrature(m1t, m12, t1), facets=orig1[t1])
fbasis2 = FacetBasis(m2, e2, quadrature=elementwise_quadrature(m2t, m12, t2), facets=orig2[t2])

fbasis = fbasis1 * fbasis2

@NonlinearForm(hessian=True)
def g(u1, u2, w):
    eps = 1e-2
    return 1 / eps * np.minimum(0.25 + u1[1] - u2[1], 0) ** 2

@NonlinearForm(hessian=True)
def J1(u, w):
    eps = 1/2 * (grad(u) + transpose(grad(u)) + mul(transpose(grad(u)), grad(u)))
    sig = 2 * eps + eye(trace(eps), 2)
    return 1/2 * ddot(sig, eps)


@NonlinearForm(hessian=True)
def J2(u, w):
    eps = 1/2 * (grad(u) + transpose(grad(u)) + mul(transpose(grad(u)), grad(u)))
    sig = 2 * eps + eye(trace(eps), 2)
    return 1/2 * ddot(sig, eps) - 1e-2 * u[1]

x = fbasis.zeros()
dx = onp.array(fbasis.zeros())

for itr in range(40):
    jac1, rhs1 = J1.assemble(basis1, x=x[:basis1.N])
    jac2, rhs2 = J2.assemble(basis2, x=x[basis1.N:])
    jac = bmat([[jac1, None], [None, jac2]])
    rhs = np.concatenate((rhs1, rhs2))
    jacg, rhsg = g.assemble(fbasis, x=x)
    jac = jac + jacg
    rhs = rhs + rhsg
    dx = solve(*enforce(jac, onp.array(rhs), D=onp.array(onp.concatenate((
        basis1.get_dofs({'left', 'right'}).all(),
        basis2.get_dofs({'left', 'right'}).all() + basis1.N,
    )))))
    x += dx
    print(np.linalg.norm(dx))
    if np.linalg.norm(dx) < 1e-8:
        break
    mdefo1 = m1.translated(x[:basis1.N][basis1.nodal_dofs])
    mdefo2 = m2.translated(x[basis1.N:][basis2.nodal_dofs])
    ax = mdefo1.draw()
    mdefo2.draw(ax=ax)
```

kinnala:
Now here is something to consider:
- How should `basis.N` be picked? (Mortaring vs. DG is different?)
- Make sure mesh types can be different (for mortaring)
- How about projection from one mesh to another?

kinnala:
Fancy pics from the code above. One of the earlier iterations:
image
The final iteration:
image

kinnala:
move to #1143

PR #1131, opened 2024-06-11
Fix Mesh.p2e

vohonen:
The `Mesh.p2e` (vertices to edges) method functioned actually as `Mesh.e2t `(edges to elements). I renamed the original method as `e2t` and added a new method `p2e` that produces a map between vertices and edges.

This might have slipped as there were no test cases for `Mesh.p2e `beforehand. I added cases for both `p2e` and `e2t`, and fixed the indentation of the other cases for that particular test.

### DIFF
--- a/tests/test_mesh.py
+++ b/tests/test_mesh.py
@@ -793,17 +793,18 @@ def test_incidence(mesh):
     for itr in range(0, 50, 3):
         assert np.sum((mesh.facets == itr).any(axis=0)) == len(p2f[:, itr].data)
 
-    p2e = mesh.p2e
-    for itr in range(0, 50, 3):
-        assert np.sum((mesh.edges == itr).any(axis=0)) == len(p2e[:, itr].data)
+    if isinstance(mesh, (MeshTet1, MeshHex1)):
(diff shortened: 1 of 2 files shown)


tests/test_mesh.py
```diff
@@ -787,12 +787,23 @@ def test_incidence(mesh):
...
     for itr in range(0, 50, 3):
-       assert np.sum((mesh.t == itr).any(axis=0)) == len(p2t[:, itr].data)
+        assert np.sum((mesh.t == itr).any(axis=0)) == len(p2t[:, itr].data)
 
     p2f = mesh.p2f
     for itr in range(0, 50, 3):
-       assert np.sum((mesh.facets == itr).any(axis=0)) == len(p2f[:, itr].data)
+        assert np.sum((mesh.facets == itr).any(axis=0)) == len(p2f[:, itr].data)
 
+    p2e = mesh.p2e
```
kinnala, line 796: These additional tests should run only for `MeshTet` and `MeshHex`.

vohonen: Added a check for those mesh objects in a new commit

Issue #1137, opened 2024-07-02
Adding advection-diffusion as another usage example


RadostW:
We've managed to solve an advection-diffusion problem

$$0 = \Delta \phi - \mathrm{Pe} (u \cdot \nabla \phi) $$

in cylindrical coordinates with $U$ coming from Stokes flow around a sphere using `scikit-fem` package -- [code here](https://github.com/turczyneq/fem_advection_diffusion/blob/main/stokes_flow_cylindric_metric.py)

I'd like to contribute it as an use example -- is there some clever way this code should be structured to work with `listofexamples.rst`?

![sample_image](https://github.com/kinnala/scikit-fem/assets/51670923/2c1146b9-c984-4b15-b310-40a7ac6b5c30)

kinnala:
This looks great. You can add it as ex50.py, implement a test for it under tests/test_examples.py, and use the following as a template for listofexamples.rst (under fluid mechanics category, perhaps?):
```
Example 36: Nearly incompressible hyperelasticity
-------------------------------------------------

This example demonstrates the implementation of a two field mixed formulation
for nearly incompressible Neo-Hookean solids.

.. figure:: https://user-images.githubusercontent.com/22624037/91212007-4055aa80-e6d5-11ea-8572-f27986887331.png

   The displacement contour of Example 36.
   The figure was created using `ParaView <https://www.paraview.org/>`__.

See the source code of :exlink:`36` for more information.
```
Simply copy the image url from this issue and replace it in the template.

kinnala:
I suggest that you save the mesh to file through mesh.save(”docs/examples/meshes/cylinder.msh”), include the mesh in the commit
 and avoid using pygmsh by loading the mesh in the beginning of the example, via MeshTri.load(…). I think pygmsh is not included in our CI.

kinnala:
Finally, we accept nowadays only examples with a matching license (BSD-3-clause) so hopefully you are able to use such a license instead of GPL.

PR #1138, opened 2024-07-02
Add advection-diffusion as another usage example

RadostW:
Modified documentation as explained in #1137


kinnala:
Thank you, I will simplify it slightly and add a test so that it won't get broken over time.

PR #1199, opened 2025-12-16
Added third order nedelec elements.

Arthaj-Octopus:
Add third-order Nedelec triangle elements.


kinnala:
You can check what is run by the test suite here:

https://github.com/kinnala/scikit-fem/blob/master/.github/workflows/main.yml

So basically flake8 is now reporting various noncritical syntax-related issues that are easy to fix.

PR #1206, opened 2026-04-17
fix(_intersect1D): restrict supermesh to the intersection of both x-r…

luedkej:
Previously, the supermesh was built from the union of all nodes from both input meshes. When the two boundary traces had different x-ranges, the supermesh extended beyond one or both of the original meshes. The midpoints of those out-of-range supermesh elements were then passed to element_finder(), which uses np.digitize without clamping:
- points below the mesh minimum produce a bin index of 0, which matches
  no element's right-node → ValueError("Point is outside of the mesh.")
- points above the mesh maximum produce a bin index of n_nodes, causing
  an IndexError (out of bounds on the sorted index array ix).

Fix: before merging the two node arrays, clip each to the x-range shared by both meshes ([max(min_p1, min_p2), min(max_p1, max_p2)]). The supermesh then only spans the region where both boundaries overlap, so every supermesh midpoint is guaranteed to lie inside both original meshes when element_finder is called.

This also reflects the correct physical interpretation: the contact (or mortar) integral should only be evaluated over the region where the two surfaces actually overlap.

### DIFF
--- a/skfem/supermeshing.py
+++ b/skfem/supermeshing.py
@@ -69,9 +69,17 @@ def _intersect2d(p1, t1, p2, t2):
 
 def _intersect1d(p1, t1, p2, t2):
     """One-dimensional supermesh."""
-    # find unique supermesh facets by combining nodes from both sides
-    p = np.concatenate((p1.flatten().round(decimals=10),
-                        p2.flatten().round(decimals=10)))
+    # find unique supermesh facets by combining nodes from both sides in the intersection of x-ranges 
+    p1f = p1.flatten().round(decimals=10)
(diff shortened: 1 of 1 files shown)


kinnala:
This makes sense. Not quite sure how it works right now in 2D supermesh but at least for 1D this is the way to do it.

luedkej:
Would you like the test script as well in the pr?

kinnala:
This is enough, thanks for the contribution!

luedkej:
Thank you very much for your work and greetings from ETH, Switzerland.

PR #1209, opened 2026-05-22
Correcting lamé parameters for plane stress model

luedkej:
The former calculation of the lamé parameters lambda and mu is to the best of my understanding wrong. The updated fromulae correspond to the derivation of the lamé parameters derived from the generalized Hooks Law for sigma_zz = 0.


kinnala:
Thanks a lot, I wonder what kind of brainrot lead to this. I don’t remember ever using this function. I’ll just once again verify this formula from some source before merging.

PR #1213, opened 2026-06-20
Fix PermissionError in mesh save/load tests on Windows (#882)

### LINKED ISSUE #882: Release tests on Windows
The release tests are breaking on Windows because of temporary files. Could be that `delete=False` must be added such as here: https://github.com/appveyor/ci/issues/2547

bypire:
Fixes #882.

### Problem

The mesh save/load round-trip tests fail on Windows. `test_saveload_cycle_vtk`
and `test_saveload_cycle_tags` use

```python
with NamedTemporaryFile(suffix=...) as f:
    m.save(f.name)
    m2 = Mesh.load(f.name)
```

On Windows the OS keeps an **exclusive lock** on a `NamedTemporaryFile` while it
is open, so when `m.save(f.name)` (via `meshio`) reopens the same path for
writing it raises `PermissionError: [Errno 13] Permission denied`. The CI only
runs `ubuntu-latest`, so this is never exercised.

Reproduced on Windows 11 / Python 3.11: `pytest tests/test_mesh.py -k saveload`
gives 68 failures, all `PermissionError` from `meshio` reopening the locked
temp file. The library save/load code itself is correct — only the test fixture
is Windows-incompatible.

### Fix

Use the idiom suggested in the issue: open the temp file with `delete=False`,
close the handle, then write/read the path, and remove it in a `finally` block.
This is cross-platform (on Linux it simply removes the file in `finally` instead
of on context-manager exit).

### Verification

With the change, `pytest tests/test_mesh.py -k saveload` passes (68 passed), and
the full `pytest tests/test_mesh.py` shows no regressions. `flake8 skfem` is
clean. Verified on Windows 11 / Python 3.11.9.

I kept this PR to the test fix only. If a `windows-latest` entry in the CI
matrix would be welcome to prevent regressions, I am happy to add it in a
separate PR.

### DIFF
--- a/tests/test_mesh.py
+++ b/tests/test_mesh.py
@@ -1,3 +1,4 @@
+import os
 from unittest import TestCase
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


gdmcbain:
I've verified the issue and the remedy on MS-Windows 11, Python 3.13..13. LGTM.

PR #1215, opened 2026-07-29
Add Example 56: Matérn covariance as a sparse precision matrix

psjcodes:
Adds an example that assembles the sparse precision matrix of the Gaussian Markov random field approximating a Matérn field, and plots correlations for a constant range, a varying range and across a barrier. The test pins the error against the analytical Matérn correlation and the marginal variance, and checks that the barrier both blocks and routes.

### DIFF
--- a/docs/examples/ex56.py
+++ b/docs/examples/ex56.py
@@ -1,0 +1,150 @@
+r"""Matérn covariance as a sparse precision matrix.
+
+A Gaussian field with the Matérn covariance is the stationary solution of the
+stochastic partial differential equation
+
+.. math::
+   (\kappa^2 - \Delta) u = \mathcal{W}
+
+driven by Gaussian white noise :math:`\mathcal{W}` [WHITTLE]_.  In two
+dimensions this is the case :math:`\alpha = 2` of the family, so the smoothness
+is :math:`\nu = 1` and the correlation range is :math:`\sqrt 8/\kappa`.
+Discretizing the equation with linear triangles gives a Gaussian Markov random
+field with the sparse precision matrix
+
+.. math::
+   Q = (\kappa^2 C + G)\,C^{-1}(\kappa^2 C + G),
+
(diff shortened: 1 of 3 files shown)


gdmcbain:
I think this may be of interest to some users, certainly including me, but it might be worth expanding the docstring for the rationale.  I have only recently first heard of Matérn, cited by Stein (1999), cited by Kennedy & O'Hagan (2001), cited by Girolami et al (2021), cited by @cottrell here in #666 "Examples providing your own mesh and retriving the basis functions?", back on [2021-07-19](https://github.com/kinnala/scikit-fem/issues/666#issuecomment-882356020).  I've been following that trail slowly (five years have slipped by!), so it's good to see more work on it.
    In the work of cottrell, the motivation was parameter estimation.  Is that the same here?  Or other there other applications of stochastic partial differential equations?

## References

* Girolami, M., Febrianto, E., Yin, G., & Cirak, F. (2021). The statistical finite element method (statFEM) for coherent synthesis of observation data and model predictions. _Computer Methods in Applied Mechanics and Engineering_ **375**:113533. [doi: 10.1016/j.cma.2020.113533](https://doi.org/10.1016/j.cma.2020.113533)
* Kennedy, M. C., & O’Hagan, A. (2001). Bayesian calibration of computer models. _Journal of the Royal Statistical Society, Series B, Statistical Methodology_ **63**(3), 425–464. [doi:10.1111/1467-9868.00294](https://doi.org/10.1111/1467-9868.00294)
* Stein, M. L. (1999). _Interpolation of Spatial Data_. Springer New York. [doi: 10.1007/978-1-4612-1494-6](https://doi.org/10.1007/978-1-4612-1494-6)

kinnala:
I have done some Matérn covariance stuff before when I was doing Gaussian process regression with energy consumption data (for optimizing energy consumption). This is cool, thanks for the contribution.

Edit: I remember reading this book a lot: https://gaussianprocess.org/gpml/chapters/RW.pdf

Issue #1216, opened 2026-08-17
`MeshTri1DG.periodic` drops continuity and prints warning


lorenzocelli:
Method `MeshTri1DG.periodic` internally creates a non-contiguous array and prints a warning even if the input is C contiguous.
Here is some code to replicate:
```python
import numpy as np
from skfem import MeshTri, MeshTri1DG

mesh = MeshTri.init_tensor(np.array([0.0, 1.0]), np.linspace(0.0, 1.0, 1001))
print(f"mesh.doflocs.flags['C_CONTIGUOUS']: {mesh.doflocs.flags['C_CONTIGUOUS']}")

ix = np.array([1000, 2001], dtype=np.int32)
ix0 = np.array([0, 1001], dtype=np.int32)
MeshTri1DG.periodic(mesh, ix, ix0)
```
The output includes a warning `Transforming over 1000 vertices to C_CONTIGUOUS.`:
```
mesh.doflocs.flags['C_CONTIGUOUS']: True
Transforming over 1000 vertices to C_CONTIGUOUS.
```
After debugging I found the issue is [line 69](https://github.com/kinnala/scikit-fem/blob/18e98da7d357c97b2c4d9429fa521ade163924aa/skfem/mesh/mesh_dg.py#L69) in `mesh_dg.py`:
```
doflocs = np.hstack((mesh.doflocs[:, oix], mesh.doflocs[:, ix]))
```
Indexing with `oix`and `ix` drops continuity. This can be fixed by replacing the indexing with `np.take` which instead preserves continuity:
```
doflocs = np.hstack((np.take(mesh.doflocs, oix, axis=1), np.take(mesh.doflocs, ix, axis=1)))
```
I can open a pr if you want. I'm using numpy v. 2.4.4, and scikit-fem v. 12.0.2.

kinnala:
Thanks, I will merge if you create PR. Otherwise I will fix before December maintenance release (doing these twice a year).

PR #1217, opened 2026-08-19
Preserve continuity in `MeshTri1DG.periodic` #1216

lorenzocelli:
Solve issue described in #1216, and keeps `doflocs` array C_CONTIGUOUS in method `MeshTri1DG.periodic`.

I could not contribute a test since the only difference from the caller perspective is the warning being printed. The only option would be to add a test that captures the warning with the `caplog` fixture, but it's not very stable. Let me know if you have other ideas.

Thanks!


kinnala:
I see it as an optimization. Optimizations are not covered by the current test suite. Hence it is enough if the tests pass. I do some rough performance testing every now and then but there is not anything formal in place.

PR #1221, opened 2026-09-17
Fix broken example links in latest documentation

### LINKED ISSUE #1204: DOC exlinks from docs to examples missing tag
The [Gallery of examples](<https://scikit-fem.readthedocs.io/en/latest/listofexamples.html#gallery-of-examples>) links to the source code for each; e.g., [Example 1: Poisson equation with unit load](<https://scikit-fem.readthedocs.io/en/latest/listofexamples.html#example-1-poisson-equation-with-unit-load>) reads

> See the source code of [example 01](https://github.com/kinnala/scikit-fem/blob//docs/examples/ex01.py) for more information.

but that link is 404; it should read

- not <https://github.com/kinnala/scikit-fem/blob//docs/examples/ex01.py>
- but <https://github.com/kinnala/scikit-fem/blob/12.0.0/docs/examples/ex01.py>

i.e., the tag or branch is blank.

The source of that link is https://github.com/kinnala/scikit-fem/blob/365167e295a06ca708a5e066bcbe11051b34f3a3/docs/listofexamples.rst?plain=1#L26 and the Sphinx `:exlink:` there is defined in https://github.com/kinnala/scikit-fem/blob/365167e295a06ca708a5e066bcbe11051b34f3a3/docs/conf.py#L54-L59

So it looks like `previous_release` is empty?

guhou-hvi:
## Problem

On ReadTheDocs the "latest" build of the documentation produces broken
`:exlink:` URLs such as:

```
https://github.com/kinnala/scikit-fem/blob//docs/examples/ex01.py
```

i.e. the version tag in the URL is empty, giving a 404 when a reader clicks
any example link from the latest docs. The "stable" docs and any tag-built
version are unaffected. Reported in #1204.

## Root cause

`docs/conf.py` resolves the version like this:

```python
release = re.sub('^v', '', os.popen('git describe --tags').read().strip())
version = release
previous_release = version.split("-")[0]
```

ReadTheDocs builds the "latest" branch with a **shallow clone** (depth=1,
no tags fetched), so `git describe --tags` fails and the captured stdout
is empty. `previous_release` then becomes `""` and the URL template
`.../blob/{}/docs/examples/...` produces `.../blob//docs/examples/...`.

Locally and on tag-built versions `git describe --tags` succeeds, which is
why nobody could reproduce it outside RTD (see the discussion in #1204).

## Fix

Fall back to the default branch name when the tag is unavailable:

```python
previous_release = version.split("-")[0] or "master"
```

Semantically this is correct: the `latest` docs are built from the
`master` branch, so example links pointing at `master` actually match the
code the docs describe — arguably more useful than pinning to whatever
tag happened to exist when the last stable was cut.

The change is one line plus a short comment. `release` / `version`
are left alone so the HTML header still shows the same value as before.

## Testing

`previous_release` was verified against three scenarios (the same env
that RTD would provide):

| Scenario | `git describe --tags` returns | `previous_release` | URL template |
|---|---|---|---|
| Local full clone | `12.0.2-4-g73e83570` | `12.0.2` | `…/blob/12.0.2/docs/examples/ex%s.py` |
| RTD `latest` (shallow clone) | `` (empty) | `master` | `…/blob/master/docs/examples/ex%s.py` |
| RTD tag / stable | `12.0.2` | `12.0.2` | `…/blob/12.0.2/docs/examples/ex%s.py` |

None of the three produce the broken `…/blob//docs/…` form.
A full Sphinx build (`make html` equivalent: `sphinx -b html -E docs _build/html`)
runs with **0 broken example URLs** in the output HTML.

The existing test suite is unaffected by this change (only `docs/conf.py`
is modified; `flake8 skfem` — the repo's only linter — explicitly excludes
`docs/`).

Closes #1204.


kinnala:
Thanks for the contribution

PR #1223, opened 2026-09-20
Fix periodic Hermite and Argyris geometry and verify BFS

### LINKED ISSUE #902: MeshDG not compatible with ElementGlobal
Here is minimal example:
```python
m = MeshLine(np.linspace(0, 1, 10))
mp = MeshLine1DG.periodic(m, [9], [0])
basis = Basis(mp, ElementLineHermite())
```
leads to
```
LinAlgError: Singular matrix
```

guhou-hvi:
Fixes #902 for one-dimensional Hermite elements and extends the mapped-vertex handling to two dimensions. Boundary normals are evaluated directly from the adjacent cell without constructing a FacetBasis.

Periodic vertex identification can reverse the local edge ordering for an Argyris normal-derivative DOF. Aligning both cells with the first adjacent cell fixes the resulting gradient discontinuity. Existing physical boundary projections are retained.

Tests cover periodic Hermite, Argyris triangles, and BFS on rectangular tensor grids. They compare mass, Poisson, and Hessian matrices with equivalent periodic constraints, check nodal derivatives and seam continuity where applicable, and reproduce local polynomials. They include nonuniform, shifted, and renumbered meshes. Registration monkeypatch and custom-subclass tests have been replaced with built-in element behavior tests.

Local validation on Windows / Python 3.12.2 followed the main CI selection, `pytest -q -k 'not test_mamba'`: 585 passed, 1 skipped, 2 deselected, including existing and newly added tests. The two tests assigned to the separate mamba workflow were not run locally. All 188 repository doctest examples and flake8 pass. Eight mypy errors match the same-environment baseline; Sphinx retains 49 existing warnings. Ordinary and curved mesh comparisons are unchanged. Initialization benchmark medians range from -3.92% to +3.06% on the measured sizes.

This does not add public APIs or general DG interior-facet mapping support. BFS coverage is limited to rectangular tensor grids; existing large-coordinate Vandermonde conditioning remains.

Codex implemented the changes and tests, ran local checks, and drafted this description. A separate Codex agent reviewed the revision. I reviewed this revision before publication.


kinnala:
So the workaround is for 1D only? The code looks like it would run for 2D but is there some obvious reason why it wouldn't work there?

guhou-hvi:
The vertex-mapping step does generalize to 2D. There is a separate obstacle in the current preprocessing: `_eval_dofs` constructs a `FacetBasis` for normals, but `MeshTri1DG` has no boundary element, which leads to a `NoneType.lbasis` error. `ElementTriHermite` does not use those normals.

The update in `9e2cbbe2` uses mapped vertices in 1D and 2D and skips that preprocessing when the element has no facet DOFs and the mesh has no boundary element. Ordinary meshes retain the existing edge data, including for custom elements with no facet DOFs.

The added tests cover x-, y-, and doubly periodic triangular meshes, including nonuniform and shifted coordinates. Mass and Poisson stiffness matrices agree with an ordinary mesh under equivalent periodic DOF constraints; vertex values and both derivatives agree, and local cubic polynomials are reproduced. The existing 1D tests still pass. The full local test selection has 555 passed, 1 skipped, and 2 deselected.

This supports `ElementTriHermite`, not periodic Morley or Argyris. It also does not imply C1 continuity along an entire triangle edge.

kinnala:
The obvious follow up question is: did it not help to add a `bndelem` by modifying `BOUNDARY_ELEMENT_MAP`? `ElementHex1DG` seems to already be there.

kinnala:
Note: here is the line that I am referring to: https://github.com/kinnala/scikit-fem/blob/fe6b326e9919f94615d3c3564d09aeabae669313/skfem/element/__init__.py#L45

guhou-hvi:
Thanks for the pointer. The earlier revision had not tested that registration. Codex-assisted checks now confirm that adding `ElementTriP1DG: ElementLineP1` removes the missing-element error, but singly periodic meshes still fail because boundary facet indices do not identify the DG geometry nodes correctly.

Reviewing the call site revealed a smaller solution: `_eval_dofs` only needs boundary normals, not a complete `FacetBasis`. The update in `da590abe` calls `mapping.normals` directly at the reference-edge quadrature point. This removes both the `FacetBasis` dependency and the `facet_dofs`/`bndelem` guard, while preserving edge data for custom elements.

The revision keeps the original first quadrature point, endpoint ordering, and projection magnitude on curved cells. Registration-invariance and curved-boundary tests pass; 80 ordinary/curved context and matrix comparisons are unchanged. Local validation has 560 passed and 188 doctests passed, with independent review completed.

This avoids expanding the PR into general DG facet mapping or periodic interior-facet coordinate semantics.

kinnala:
It's now genuinely useful! Does it work also for, e.g., Argyris and quadrilateral elements such as BFS?

The tests still cover only ElementTriHermite (and 1D Hermite) as far as I understand. Also, some of the tests are weird, e.g., what is the monkey patching test about? Moreover, I have not expected people to create subclasses of these elements so I think those tests are unnecessary.

guhou-hvi:
Yes—the added tests now cover periodic Argyris triangles and BFS on rectangular tensor grids, including x-, y-, and doubly periodic cases. The Argyris checks exposed an edge-orientation issue when boundary vertices were renumbered: the two cells could use opposite directions for the same normal-derivative DOF. This revision aligns their directions with the first adjacent cell.

The new regression tests in `tests/test_periodic_c1.py` use two independently assembled representations of the same problem: the periodic mesh, and the ordinary mesh with matching boundary DOFs constrained together. They verify `A_periodic ≈ R.T @ A_ordinary @ R` for the mass, Poisson, and Hessian forms, which respectively involve values, first derivatives, and second derivatives. `R` includes the signs required when Argyris normal-derivative DOFs use opposite normal directions.

Those regressions also check all nodal DOFs: `u, u_x, u_y, u_xx, u_xy, u_yy` for Argyris, and `u, u_x, u_y, u_xy` for BFS. With random coefficients, values and both gradient components are compared at five points on each shared edge, including periodic seams. Local polynomial checks reconstruct a representative total-degree-five polynomial for Argyris and a bicubic polynomial for BFS, comparing their values and gradients with the analytical expressions. These use element-local coefficients because the selected polynomials are not globally periodic.

The renumbering regression previously produced a seam-gradient jump of about `2.15`; the revision reduces it to `9.34e-11`, matching the ordinary-mesh control.

The monkeypatch test was introduced by this PR to check whether registering a boundary element changed Hermite assembly. I removed it and the two custom Hermite subclass tests introduced by this PR. The curved-boundary test now uses the built-in Morley element to reconstruct a quadratic polynomial, without a capturing subclass.

For regression checking, Codex ran `pytest -q -k 'not test_mamba'`, following the selection in the repository's main CI workflow: 585 passed, 1 skipped, 2 deselected. This total includes the existing suite and the tests added by this PR; it is not a count of new tests. The two deselected tests belong to the separate mamba workflow and were not run locally. The repository's Sphinx doctests also passed all 188 examples, and flake8 passed. Codex implemented and checked the revision, with a separate agent review.

kinnala:
Great. I double checked that there is not much difference in the run time either. I will once more read through the new tests and then merge if they look fine.

tests/test_periodic_c1.py
```diff
@@ -0,0 +1,171 @@
...
+    tangent = vertices[:, endpoints[0]] - vertices[:, endpoints[1]]
+    normal = np.array([tangent[1], -tangent[0]])
+    normal /= np.linalg.norm(normal)
+    if mesh.f2t[1, facet] == -1:
+        away = vertices[:, endpoints].mean(axis=1) - vertices.mean(axis=1)
+        normal *= np.sign(normal @ away)
+    return normal
+
+
+def restriction(ordinary, periodic):
```
kinnala: Would something like this be an interesting feature, part of the library?

Just leaving a note here, we can think of it later.

